PR #6, opened 2022-12-14
Docker builds

palp:
- Add dockerfile for bot
- Add build actions for docker images (bot + backend)


yk:
do we need to build & publish on every push? Do you think it makes sense to restrict it to tags starting with "v", which would then be our releases?

palp:
> do we need to build & publish on every push? Do you think it makes sense to restrict it to tags starting with "v", which would then be our releases?

Yeah, every push is probably too much. PR builds might be useful for testing, but also probably a lot of noise. I'll set it up to work off tags for now and we can adjust if needed.

palp:
I left in the path filters, which should skip "empty" releases (i.e., only the bot was modified, so no new image is built of the backend). This may or may not make sense to do.

yk (approved):
thank you!

PR #7, opened 2022-12-14
implemented "Ranked Pairs" algorithm for merging rankings from different users

MattAlexMiracle:
This is an implementation of ["Ranked Pairs" (or Tideman's method)](https://en.wikipedia.org/wiki/Ranked_pairs) for merging different preference structures.
There are many impossibility results of how you can merge different preference structures, like Arrow's impossibility theorem that show that mergings can never be unilaterally perfect.
However, for the properties we need, Ranked Pairs seems to be optimal, since the drawbacks it does have don't apply to us.

One downside is the rather high computational overhead of ~O(n^3), but I think we are not going to have too many votes per prompt, making this not that big of an issue for us (we don't need to rank between different prompts).

### DIFF
--- a/backend/postprocessing/rankings.py
+++ b/backend/postprocessing/rankings.py
@@ -1,42 +1,45 @@
-import numpy as np
+# -*- coding: utf-8 -*-
 from typing import List
 
+import numpy as np
 
-def head_to_head_votes(ranks:List[List[int]]):
+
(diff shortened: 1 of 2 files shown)


yk:
thank you! could you add numpy to backend/requirements.txt ?

yk (approved):
lgtm, thank you

PR #8, opened 2022-12-14
Bot to laion update

BirgerMoell:
Updated code to make the bot work on the LAION server


yk (approved):
just to think about: did you consider what happens if these environment variables don't exist?

PR #10, opened 2022-12-15
Integrating Next-Auth to do user authentication on the website

fozziethebeat:
This updates the website to provide two forms of user authentication (Discord & Email) and then proxies data requests to the FastAPI backend while keeping the client api key secure.  This also registered newly signed up users to the labeler backend.

The UI is still pretty ugly and not useful in any way.


yk (approved):
lgtm

suggestion for improvement: use nextjs runtime configuration (https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration)
the reason is: process.env is grabbed at build-time in many places, so we'd have to include secrets into the docker build, which is suboptimal. with runtime configuration, we can supply environment variables at runtime

PR #13, opened 2022-12-15
implemented a simple text-based frontend

yk:


fozziethebeat:
I like the simplicity of the data flow in the text frontend! that'll be helpful in setting up others.

PR #19, opened 2022-12-18
Update the webapp to fetch and upate tasks in sequence

fozziethebeat:
This shows an end to end example of how to handle one task type (ratings) with a task queue in the client side and storing all the task & interaction state with an intermediate database.  This is mostly to get a working example of a full flow.

In a future PR we will:
  - Support the various task types
  - Make it look good
  - Refactor the code to make it easier to work with.


andreaskoepf (approved):
Suggested changes could best be done after switch to Typescript.

yk (approved):

PR #27, opened 2022-12-19
create api_client class to call backend, partial bot rewrite

andreaskoepf:
Basic bot communication works, but we need to decide how the bot-user interaction really works.
Currently I have implemented a dummy time base protocol which sends messages in a bot channel.
Maybe it is better to go with commands to initialize actions.

But then we still need an intuitive way to interact with the bot.


yk (approved):
LGTM, thank you!

Some things to improve (but can be done later):
- use httpx instead of requests for async i/o
- logging instead of prints
- general cleanup, including sharing code between bot and backend (I'll take that)

PR #36, opened 2022-12-19
Link new UI to auth and backend

fozziethebeat:
This integrates the recent new landing page and demo grading page with full user authorization and full task fetching & updating.

included in this is use of Chakra-UI components where they make sense.  The initial uses are:
  - Icon buttons to simplify the code
  - Updating the custom radio buttons for rating a prompt so that the selected value can be stored
  - The pre-built textarea with some default styling

Later we can adopt more Chakra components as desired while still use tailwindcss


fozziethebeat:
Set Yannic as reviewer until Desmond is setup as a reviewer for web.

k-nearest-neighbor (approved):
This looks great.

While I'm still working on getting this fully tested on local. I'm turning in for the night and don't want to hold you up.

I've looked at the code and it all makes sense to me.

Stuff we can work on soon

- instructions for running local
- auth workflows

scripts/frontend-development/docker-compose.yaml
```diff
@@ -10,6 +11,23 @@ services:
...
+    ports:
+      - 5433:5432
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "postgres"]
+      interval: 2s
+      timeout: 2s
+      retries: 10
```
yk, line 28:
```suggestion
    extends: db
```

scripts/frontend-development/docker-compose.yaml
```diff
@@ -26,3 +44,15 @@ services:
         condition: service_healthy
     ports:
       - "8080:8080"
+
+  # This fakes and SMTP email server.  User registration emails can be found by going to
+  #   localhost:1080 and opening the emails listed.
+  maildev:
```
yk, line 50: is the mail server needed just for the custom auth or also for the discord auth workflow?

website/src/components/Header.tsx
```diff
@@ -35,6 +36,31 @@ function MobileNavLink({ children, ...props }) {
...
+    );
+  }
+  return (
+    <Link href="/auth/signin" aria-label="Home" className="flex items-center">
+      <button
+        type="button"
+        className="inline-flex items-center rounded-md border border-gray-300 bg-white px-4 py-2 text-sm font-medium text-gray-700 shadow-sm hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-2"
+      >
+        Log in
+      </button>
```
yk, line 59: button looks exactly the same as the button above, could this be a component?

website/src/pages/grading/grade-output.tsx
```diff
@@ -1,113 +1,150 @@
...
+          <button
+            type="button"
+            className="mr-2 inline-flex items-center rounded-md border border-transparent bg-indigo-100 px-4 py-2 text-sm font-medium text-indigo-700 hover:bg-indigo-200 focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-2"
+          >
+            Skip
+          </button>
+          <button
+            type="button"
+            onClick={() => submitResponse(tasks[0])}
+            className="nline-flex items-center rounded-md border border-transparent bg-indigo-600 px-4 py-2 text-sm font-medium text-white shadow-sm hover:bg-indigo-700 focus:outline-none focus:ring-2 focus:ring-indigo-500 focus:ring-offset-2"
```
yk, line 137: should this be `inline-flex` instead of `nline-flex`?

yk (approved):
clear to merge, but I left a few comments / questions.

Also, from what I remember, using `process.env` directly leads to problems in docker builds, as variables are captured at compile time, not at runtime. Next has a runtime configuration framework here: https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration . I suggest we use that.

scripts/frontend-development/docker-compose.yaml
fozziethebeat: Yes it's just for the custom auth.

I'll enhance the comment in the next PR.

website/src/components/Header.tsx
fozziethebeat: I agree, this was originally from @k-nearest-neighbor.  I'm leaving css and component cleanup for later changes.

website/src/pages/grading/grade-output.tsx
fozziethebeat: I didn't catch this either in my first review.  It probably should be.  Will be fixed later.

fozziethebeat:
Agreed on using an env setup that works well with Docker.  That's what i'll be exploring in the next PR.

PR #39, opened 2022-12-19
Extracted shared python code and made it work with dockerfiles

yk:
Extracts shared python code to `oasst-shared` with package `oasst_shared`.
Caveat is, users must `pip install -e .` inside the `oasst-shared` directory in order to be able to import it. Dockerfiles are adjusted to work out of the box.


andreaskoepf (approved):
Really nice, like the new structure!

PR #56, opened 2022-12-23
added next.js font module, updated fonts, updated login page

rsandb:


yk:
looks cool so far!
for the logos, there is a new one in the repo top-level `assets` folder. could you use that one? it's about 5 times smaller and doesn't include an unnecessary embedded PNG

rsandb:
Yeah, I can change that! My fork wasn't synced when that was added so I'll do that

yk (approved):
lgtm, thank you!

website/src/pages/auth/signup.tsx
```diff
@@ -0,0 +1,92 @@
...
+                leftIcon={<FaEnvelope />}
+                colorScheme="gray"
+                onClick={signinWithEmail}
+                // isDisabled="false"
+              >
+                Continue with Email
+              </Button>
+            </Stack>
+          )}
+          {/* {discord && ( */}
```
fozziethebeat, line 39: Question, why is this part commented out?

fozziethebeat (changes requested):
Can we add in the optional hiding and displaying of the discord button?  we can also do the same with `github` and I'll soon be adding in dev credentials for that method.

website/src/pages/auth/signup.tsx
rsandb: Just to show the Discord button. It wasn't appearing with that in place, so I commented it out while I was formatting everything

fozziethebeat (approved):
nice!

Issue #37, opened 2022-12-19
Scoring-algorithm for crowd-sourced data collection


andreaskoepf:
Interactions with our human-feedback front-ends (discord bot + website) should be incentivized with a user score that is shown on a leader board (gamification) + a news feed that notifies (on discord and website) about recent user activity (potentially anonymized if desired by the user).

It is not sufficient for us to collect a large number of user-interactions. We want to especially incentivize submitting high-quality instruct-fulfillment data points and human written agent-responses (demonstrations). We plan to estimate quality of user provided data in a similar way to how feedback on language-model outputs is generated - by using human feedback (e.g. grading/ranking input provided by other users).

We want to use a mix of distributed community and centralized administrator moderation. Centralized admin moderation (disabling users and deleting their data) is expensive and not easily scalable.

Challenging user behavior may include:

- random ranking and grading
- giving intentionally wrong feedback
- politically motivated fake news (e.g. flooding the dataset)
- duplicate data entry
- DoS attacks
...

yk:
Maybe to add: this goes beyond just the scoring algorithm. The main question is how can we design the system to incentivize as much useful and high-quality feedback as possible.

MattAlexMiracle:
I have a first draft of the scoring and question selection system as a pull request:
The overall goal I'm targetting is:
1. "No trolls": we do not want to reward or amplify malicious characters interacting with the system
2. "No echo chamber": we do not want to greedily reward the consensus opinion as that would just produce an echo chamber of uninteresting and therefore lower quality prompts. It also discourages people from interacting with their system using their honest opinion, since that gives less points than if they ranked based on the presumed "mean opinion".

In this first draft I built a system to rate users in the three domains "voting", "prompting" and "ranking", each of which with their own individual point storages (for a total score we can weight and merge those).

Further, we also tally how many "good questions" a user submitted (with "good" meaning "above average"): This can be used to prevent spam on the leaderboards (and in the system) as we can simply filter out people below a certain quality threshold (e.g. "if you have contributed at least 20 times, and your score is above 80%, we show your contributions on the leaderboard and may include them in our dataset").

To do this we need to (at some point) retire a question from being shown to users, since only then can we use the consensus mechanisms to reward users (since otherwise the consensus might still change).

I envision a workflow like this:

1. use the maximum information gain metric to sample from the questions
$$x\sim Categorical(softmax(infogain(question))$$ to create a pseudo-thompson sampling scheme over all questions
[...]

PR #57, opened 2022-12-25
user scoring and question selection

MattAlexMiracle:
This implements an early version of #37 using the following systems:
- We use "information gain" to model how much a prompt's rating could benefit from new votes
- We score votes based on how close they were to the optimal vote, i.e. if the consensus produces [100,300,200] and you vote for index 2 (counting from 0), you would get 1 point, since the ranking is [0,2,1] as 300 is the biggest, 200 the second, and 100 the smallest value.
- We score prompts using "positive edge": i.e. if your votes are [200,300,100,500], you would multiply it with the (signed) distance to the middle, i.e. [-1,0,1,2] and get $-1\cdot 200+0\cdot 300+1\cdot 100+2\cdot 500$ points
- We then score rankings using the kendall-tau correlation (or "normalized bubble sort distance"): this compares the combined ranking from all rankers to the ranking an individual user produced, and gives points based on how similar they are.

All three point scoring methods (voting, ranking, and prompting) have their own point counts, since the scales are very different. How we weight them between each other is going to have to be experimental, as it also depends on many external factors, such as "how many votes per question do we target".


scripts/postprocessing/infogain_selector.py
```diff
@@ -0,0 +1,98 @@
...
+
+def make_range(*x):
+    """
+    constructs leftover values for the simplex given the first k entries
+    (0,x_k) = 1-(x_1+...+x_(k-1))
+    """
+    return (0, max(0, 1 - sum(x)))
+
+
+def relative_entropy(p, q):
```
andreaskoepf, line 17: A few `np.array` type-annotations would be nice in this file, too .. but we can do this later.

MattAlexMiracle: Sure, I didn't spend too much time prettying up the numerical solver, since I was able to find an analytic solution which is way faster and not numerically horrible.
At the moment I have the numerical solver only here because I needed to do some additional tests, but technically, we can remove all the numerical stuff.

yk (approved):
LGTM, thank you

if I understand correctly, a "vote" is a best-one-of-N judgement. Is there also a variant if each user provides a complete ranking of all N choices? I'm thinking if we already make the user read all of the options, they might as well rank all of them.

scripts/postprocessing/infogain_selector.py
andreaskoepf: OK, fair. I don't want to discourage you from checking in algo-development-code - this work is very important for the project (you can use `postprocessing` as some kind of prototyping area). Dense math heavy code is in general different from the 'verbose' frontend code and other aspects matter for readability and overall solution-quality.

PR #60, opened 2022-12-25
added text labels to the API and the backend

yk:
Adds the ability for frontends to send text labels to the backend to be stored.

The idea is this, for example, when a user stumbles across a text that contains profanity, they can flag that to their frontend, which will send it to the backend. Text labels can be associated with a post ID, but do not have to be. We currently provide a list of labels in `protocol -> TextLabel`, but this is up to change. Labels are submitted with a score from 0.0 to 1.0. Not every label must be present in the sent data.


backend/oasst_backend/prompt_repository.py
```diff
@@ -314,3 +314,17 @@ def insert_reaction(self, post_id: UUID, payload: db_payload.ReactionPayload) ->
...
+
+    def store_text_labels(self, text_labels: protocol_schema.TextLabels) -> TextLabels:
+        model = TextLabels(
+            api_client_id=self.api_client.id,
+            text=text_labels.text,
+            labels=text_labels.labels,
+        )
+        if text_labels.has_post_id:
+            self.fetch_post_by_frontend_post_id(text_labels.post_id, fail_if_missing=True)
+            model.post_id = text_labels.post_id
```
andreaskoepf, line 326: What are we doing with text_labels that have no post_id ... to what do they refer? Maybe we should require either a `post_id` or a `frontend_post_id`?  Maybe you had something else in mind?

oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -204,3 +204,51 @@ class PostRanking(Interaction):
...
+    text: str
+    labels: dict[TextLabel, float]
+    post_id: str | None = None
+
+    @property
+    def has_post_id(self) -> bool:
+        """Whether this TextLabels has a post_id."""
+        return bool(self.post_id)
+
+    # check that each label value is between 0 and 1
```
andreaskoepf, line 248: where did the enum __members__ check go? e.g. `str_value  in TextLabel.__members__`? didn't like it?

andreaskoepf (commented):
test

backend/alembic/versions/2022_12_25_1705-067c4002f2d9_.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""empty message
```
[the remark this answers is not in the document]
andreaskoepf: fn=file name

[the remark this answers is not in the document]
yk: good point, didn't even see that yet

backend/oasst_backend/api/v1/text_labels.py
```diff
@@ -0,0 +1,41 @@
...
+
+router = APIRouter()
+
+
+class LabelTextRequest(pydantic.BaseModel):
+    text_labels: protocol_schema.TextLabels
+    user: protocol_schema.User
+
+
+@router.post("/")  # work with Union once more types are added
```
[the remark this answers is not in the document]
yk: removed. was remnant of copy-paste

backend/oasst_backend/prompt_repository.py
yk: the idea was that this is more lightweight, i.e. the text is always sent, and the table ends up being a text -> labels mapping. If the post id exists, then good, but could also be that the user "reports" just a part of a message, or something else entirely.

oasst-shared/oasst_shared/schemas/protocol.py
yk: it's now handled by the type system of defining `labels: dict[TextLabel, float]`

backend/oasst_backend/models/text_labels.py
```diff
@@ -0,0 +1,25 @@
...
+            pg.UUID(as_uuid=True), primary_key=True, default=uuid4, server_default=sa.text("gen_random_uuid()")
+        ),
+    )
+    created_date: Optional[datetime] = Field(
+        sa_column=sa.Column(sa.DateTime(), nullable=False, server_default=sa.func.current_timestamp()),
+    )
+    api_client_id: UUID = Field(nullable=False, foreign_key="api_client.id")
+    text: str = Field(nullable=False, max_length=2**16)
+    post_id: Optional[UUID] = Field(sa_column=sa.Column(pg.UUID(as_uuid=True), sa.ForeignKey("post.id"), nullable=True))
+    labels: dict[str, float] = Field(default={}, sa_column=sa.Column(pg.JSONB), nullable=False)
```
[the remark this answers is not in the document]
yk: yes I've seen some nulls in the database, but what do you suggest? I'd still leave it `nullable=False` to communicate intent to the programmer

backend/oasst_backend/prompt_repository.py
andreaskoepf: I see. In general self-contained packages are good IMO ... but here we need a use-case how the data that enters the db could be used later (either manually or automatic). For automatic case it could be a sample for a dataset to train these classifications. But manual moderation (e.g. via fulltext search?) without references is a bit tricky.

oasst-shared/oasst_shared/schemas/protocol.py
andreaskoepf: ah, of course. ;-) nice.

backend/oasst_backend/prompt_repository.py
yk: true, my thought was mainly around training the classifiers. for the moderation bit, I think it's easy to filter where post_id is not null. the question is whether we require it or not. I can definitely imagine situation where the labels aren't really matching exactly one post, for example, assume we ask the user to rank 3 assistant answers, so our discord bot makes one discord post with all 3 answers in it. the user just wants to label answer number 1 as profane. how should they do that?

backend/oasst_backend/models/text_labels.py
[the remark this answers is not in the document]
andreaskoepf: ok, we leave it as is it. Just wanted to note that SQLModel will allow submitting `None` and from python there is currently no difference between storing `'null'::jsonb` (which happens when you submit None to non-nullable jsonb column) ... but technically it is of course not the db native null value and it will behave accordingly, e.g. equals to `'null'::jsonb` but not `is null` is false ...

backend/oasst_backend/prompt_repository.py
andreaskoepf: How do you envision users to submit a text labels without a post-id (e.g. where would they enter the classification, where would the text sent with the label-input come from)?

yk: on the website: highlight a piece of text with the mouse (or have the UI assist you). on discord, maybe there's a button (or emoji) for "label this text", after which some sort of interactive dialogue opens. yea I admit, it's probably not the most user friendly solution

andreaskoepf (approved):
Ready for merge.

(Not sure if "free text" labels (without post-reference) are a real use-case but we will see. If it turns out not to be we can add a `not null` constraint on `post_id` later.)

PR #61, opened 2022-12-25
removed ALLOW_ANY_API_KEY and allow skipping API key check in debug

yk:
the new environment variable is called DEBUG_SKIP_API_KEY_CHECK

### DIFF
--- a/backend/oasst_backend/api/deps.py
+++ b/backend/oasst_backend/api/deps.py
@@ -37,8 +37,8 @@ def api_auth(
     db: Session,
 ) -> ApiClient:
 
-    if api_key is not None:
-        if settings.ALLOW_ANY_API_KEY:
+    if api_key is not None or settings.DEBUG_SKIP_API_KEY_CHECK:
+        if settings.DEBUG_SKIP_API_KEY_CHECK:
             # make sure that a dummy api key exits in db (foreign key references)
             ANY_API_KEY_ID = UUID("00000000-1111-2222-3333-444444444444")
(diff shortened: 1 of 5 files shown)


backend/oasst_backend/api/deps.py
```diff
@@ -37,8 +37,8 @@ def api_auth(
     db: Session,
 ) -> ApiClient:
 
-    if api_key is not None:
-        if settings.ALLOW_ANY_API_KEY:
+    if api_key is not None or settings.DEBUG_SKIP_API_KEY_CHECK:
+        if settings.DEBUG_SKIP_API_KEY_CHECK:
```
andreaskoepf, line 41: The `settings.DEBUG_SKIP_API_KEY_CHECK:` was put below the `api_key is not None` check to give the dev a way to see the auth-failure case when no api-key was sent with the request. I personally think it is valuable to able to 'test' this case - even though it requires you to send a random api key. But that is debatable of course.

We might move the whole `if settings.DEBUG_SKIP_API_KEY_CHECK:` block above the if `api_key` check to simplify things.

yk: good point. I've re-introduced it, so there are now both variables to handle both cases. I've put the ALLOW_ANY into the ansible playbook so that against the dev machine, at least the presence of API keys is validated, but in local development, I've activated the complete skip.

andreaskoepf (approved):
ok, we probably need to document the settings at a later point (e.g. with some doc-strings, since with the two options there will be now more questions than before ;-)) but otherwise lgtm.

PR #62, opened 2022-12-25
added Plan and Vision section to readme

yk:


andreaskoepf (approved):
Thanks a lot for adding the high-level plan ... very helpful for first time visitors!

PR #64, opened 2022-12-26
Implement `summarize_story` for web

### LINKED ISSUE #28: Implement `summarize_story` for web
The `summarize_story` task type needs to display these fields:
*  `story`

It takes the following interaction type:
*  `type`: `text_reply_to_post`
*  `text`: The user's written summary

AbdBarho:
Partially closes #28

There is no corresponding endpoint for submission the backend, so the submission is left unimplemented.

Merry Christmas 🎄

### DIFF
--- a/website/src/pages/summarize/story.tsx
+++ b/website/src/pages/summarize/story.tsx
@@ -1,0 +1,118 @@
+// TODO(#65): Unify and simplify the task paths
+import { Textarea } from "@chakra-ui/react";
+import { useRef, useState } from "react";
+import useSWRMutation from "swr/mutation";
+import useSWRImmutable from "swr/immutable";
+
+import fetcher from "src/lib/fetcher";
+import poster from "src/lib/poster";
+
+const SummarizeStory = () => {
+  // Use an array of tasks that record the sequence of steps until a task is
+  // deemed complete.
+  const [tasks, setTasks] = useState([]);
+
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+
+  // Fetch the very fist task.  We can ignore everything except isLoading
+  // because the onSuccess handler will update `tasks` when ready.
+  const { isLoading } = useSWRImmutable("/api/new_task/summarize_story", fetcher, {
+    onSuccess: (data) => {
+      console.log(data);
+      setTasks([data]);
+    },
+  });
+
+  // Every time we submit an answer to the latest task, let the backend handle
+  // all the interactions then add the resulting task to the queue.  This ends
+  // when we hit the done task.
+  const { trigger, isMutating } = useSWRMutation("/api/update_task", poster, {
+    onSuccess: async (data) => {
+      const newTask = await data.json();
+      // This is the more efficient way to update a react state array.
+      setTasks((oldTasks) => [...oldTasks, newTask]);
+    },
+  });
+
+  // Trigger a mutation that updates the current task.  We should probably
+  // signal somewhere that this interaction is being processed.
+  const submitResponse = (task: { id: string }) => {
+    const text = inputRef.current.value.trim();
+    trigger({
+      id: task.id,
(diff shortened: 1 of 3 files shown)


website/src/pages/summarize/story.tsx
```diff
@@ -0,0 +1,107 @@
...
+      console.log(data);
+      setTasks([data]);
+    },
+  });
+
+  // Trigger a mutation that updates the current task.  We should probably
+  // signal somewhere that this interaction is being processed.
+  const submitResponse = (task: { id: string }) => {
+    const text = inputRef.current.value.trim();
+    // TODO: implement after we have the backend endpoint
```
fozziethebeat, line 27: can you uncomment and implement the mutation code here (mostly copied from `website/src/pages/grading/grade-output.tsx`) with a new field in the trigger body (that i'll make work in a following PR): `update_type: "text_reply_to_post"`

I can make this work pretty quick.

website/src/pages/summarize/story.tsx
```diff
@@ -0,0 +1,107 @@
+import { Textarea } from "@chakra-ui/react";
```
fozziethebeat, line 1: Can you add a comment at the top here:

`// TODO(#65): Unify and simplify the task paths.`

fozziethebeat (approved):
If you can, please make the two small requested changes, otherwise I'll make them in a follow up PR.

website/src/pages/summarize/story.tsx
AbdBarho: @fozziethebeat gladly, quick question: `post` in the sense of HTTP POST, or post as in reddit post?

andreaskoepf:
> There is no corresponding endpoint for submission the backend, so the submission is left unimplemented.

Merry X-Mas! :-) The endpoint to submit user generated summaries is the POST `/api/v1/tasks/interaction` endpoint, please see the text-frontend [here](https://github.com/LAION-AI/Open-Assistant/blob/f8c300800326f27180197ce5a7270c7ef3c9b9b2/text-frontend/__main__.py#L54-L63) and the discord-bot [ApiClient](https://github.com/LAION-AI/Open-Assistant/blob/f8c300800326f27180197ce5a7270c7ef3c9b9b2/discord-bot/api_client.py#L72-L74) and [SummarizeStoryHandler](https://github.com/LAION-AI/Open-Assistant/blob/f8c300800326f27180197ce5a7270c7ef3c9b9b2/discord-bot/task_handlers.py#L117-L130).

website/src/pages/summarize/story.tsx
andreaskoepf: post in `text_reply_to_post` refers to a row in the database in the post table. The post table allows to store conversation trees (threads or completions of different users or model sampling runs). The interactions like ratings and rankings are store in a separate table called post_reaction (currently users can only submit a single reaction to a given post).

AbdBarho:
@andreaskoepf thank you, I think we already have a backend that calls the right api:
https://github.com/LAION-AI/Open-Assistant/blob/54c7a942acdc2a438c73f6b1c1eec2555832c8f6/website/src/pages/api/update_task.ts#L42-L59

This was a bit unclear for me at the beginning with 2 backends, it is still unclear to me why it has to be like this, but for an MVP its ok.

yk (approved):
thank you very much :)

just briefly regarding the "two backends". The real backend is the fastapi backend. The nextjs project is technically considered a "frontend" (even though it consists of its own back- and frontend). Other frontends exist, such as the discord bot. Frontends are responsible for making the user interaction happen and then sending all information to the backend.

The architecture is like this so we don't have to re-implement a lot of logic in all the place, like the bot, the website, etc.

fozziethebeat:
There's a small change in the trigger body that I'll make in a follow up PR.

PR #68, opened 2022-12-26
created debug credentials provider

yk:
if DEBUG_FAKE_USERS is set, a credentials provider for next auth is created that lets one log in with an arbitrary username and always get accepted. this simplifies the experience for newcomers who just want to e.g. create a UI component.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -63,6 +63,14 @@ If you're doing active development we suggest the following workflow:
     navigate to `http://localhost:1080`. Check the email listed and click the
     log in link. You're now logged in and authenticated.
 
+### Using debug user credentials
+
+If you want to quickly test the website, you can use the debug user credentials workflow:
+
+1. Run `DEBUG_FAKE_USERS=true npm run dev` to start the website.
+1. Use the `Login` button in the top right to go to the login page.
+1. You should see a section for debug credentials. Enter any username you wish, you will be logged in as that user.
+
 ## Code Layout
 
 ### React Code
--- a/website/README.md
+++ b/website/README.md
@@ -65,9 +65,9 @@ If you're doing active development we suggest the following workflow:
 
 ### Using debug user credentials
 
-If you want to quickly test the website, you can use the debug user credentials workflow:
+Whenever the website runs in development mode, you can use the debug credentials provider to log in without fancy emails or OAuth.
 
-1. Run `DEBUG_FAKE_USERS=true npm run dev` to start the website.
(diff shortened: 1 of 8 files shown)


website/src/components/Avatar.tsx
```diff
@@ -12,7 +12,8 @@ export function Avatar() {
     return <></>;
   }
   if (session && session.user) {
-    const email = session.user.email;
+    console.log(session.user);
```
villelahdenvuo, line 15:
```suggestion
```

yk: thank you!

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -32,6 +34,23 @@ if (process.env.DISCORD_CLIENT_ID) {
   );
 }
 
+if (boolean(process.env.DEBUG_FAKE_USERS)) {
```
Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:37]: Should we just do this whenever we're running in development mode? If so we could cut out a new env variable and make this

```
if (process.env.NODE_ENV === "development") {
```

website/README.md
```diff
@@ -63,6 +63,14 @@ If you're doing active development we suggest the following workflow:
     navigate to `http://localhost:1080`. Check the email listed and click the
     log in link. You're now logged in and authenticated.
 
+### Using debug user credentials
+
+If you want to quickly test the website, you can use the debug user credentials workflow:
```
fozziethebeat, line 68: To simplify development, I think its reasonable to always activate this.  What about putting it into the checked in `.env` file?

fozziethebeat (approved):
I like this, I didn't notice that provider.

I proposed two small alternatives to smooth out the setup process.

website/src/pages/api/auth/[...nextauth].ts
Comment (yk) [reply website/src/pages/api/auth/[...nextauth].ts]: that's a great idea. I've pushed a change for this and adjusted documentation

fozziethebeat (approved):

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -32,6 +34,23 @@ if (process.env.DISCORD_CLIENT_ID) {
   );
 }
 
+if (boolean(process.env.NODE_ENV === "development")) {
```
Comment (villelahdenvuo) [inline website/src/pages/api/auth/[...nextauth].ts:37]:
```suggestion
if (process.env.NODE_ENV === "development") {
```

We can remove the dependency now.

Issue #67, opened 2022-12-26
Sign up form should submit on enter key


yk:
This screen here should submit the form on pressing enter
<img width="678" alt="image" src="https://user-images.githubusercontent.com/858040/209556467-5b043d77-3903-468e-8445-4159b4266bf9.png">

seyfeddinenecib:
I worked on it

PR #69, opened 2022-12-26
fix bug Open-Assistant #67 : submit form on enter

seyfeddinenecib:
Adding submit of the signup form when clicking "Enter"


yk:
hey, thanks so much for contributing :)

is there an argument for / against just making this into a form and then handling the submit? it seems like that could be a bit more streamlined than explicitly checking enter.

rsandb:
> hey, thanks so much for contributing :)
> 
> is there an argument for / against just making this into a form and then handling the submit? it seems like that could be a bit more streamlined than explicitly checking enter.

Eventually, users will need to create a username and password so having this as a form would be the best option. I have a page with that in one form; I just haven't committed it yet. For testing though I think it's easier to keep it the way it is, his enter solution just adds a bit more convenience.

fozziethebeat (changes requested):
I agree with Yannic, Rather than a new input, let's make this a proper form which should get us to the same place.

yk:
for reference, you can probably just spy on #68 where I introduce another login form that uses a proper form

seyfeddinenecib:
I implemented the changes.

fozziethebeat (approved):

Issue #74, opened 2022-12-27
Better UI for Task Options Selection


LucianPetri:
## From This: 
![image](https://user-images.githubusercontent.com/24479867/209597140-f9589aeb-a515-4ade-bb33-c70851103781.png)

## To This: 
![image](https://user-images.githubusercontent.com/24479867/209597011-4fc6ce02-1c76-4b16-a1bb-7c7f178a147b.png)
![image](https://user-images.githubusercontent.com/24479867/209597185-45e0c2b0-0500-4268-a235-dd313018d8eb.png)

### Whats Included:
- new UserChoice Component
- added LAION img ( This needs to be changed)
- added grayscale transition on hover ( disabled on xs size which are phones )
- reused the same text as in the Hero component

### Whats Next:
 - generate contextual images for these choices
 - add images to public folder and use the pre-added img prop to link them

PR #75, opened 2022-12-27
Prettier User Choice on index page

LucianPetri:
This PR is linked to this issue: Better UI Choice #74


LucianPetri:
Finished the Component, in case it looks better without the grayscale text just have undo last commit. IMO looks cooler now.

website/src/components/UserChoice.tsx
```diff
@@ -0,0 +1,51 @@
+import { Card, CardBody, Flex, Heading } from "@chakra-ui/react";
```
fozziethebeat, line 1: The  name `Choice` feels a little odd.  Is there a more intuitive name? Maybe something like `TaskOptions` as the component and then `TaskOption` for each individual thing that can be chosen.  It makes more sense to me at least to tie it to tasks rather than the user.

website/src/components/UserChoice.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Link from "next/link";
+import React from "react";
+import Image from "next/image";
+
+const UserChoice = () => {
+  return (
+    <Flex gap={10} wrap="wrap" justifyContent="center">
```
fozziethebeat, line 8: A plan I have is to have two sections: `create` and `evaluate` type tasks.  What about restructuring (and likely `Option` its own component) this a bit so it's a higher level component where someone can do the following:

```
<TaskOptions title="Create">
  <TaskOption title="Summarize Stories" link="..." icon={....} />
  ...
</TaskOptions>

<TaskOptions title="Evaluate">
  <TaskOption title="Rate Summaries" link="..." icon={....} />
  ...
</TaskOptions>
```

fozziethebeat (changes requested):
Thank you for helping clean up the logged in landing page! I like the use of chakra.

I have some suggestions on how to make this a bit more flexible for the growing set of tasks that'll be added.

website/src/components/UserChoice.tsx
LucianPetri: Great idea! i shall rename the components and restructure them. This was a 4AM idea so my naming convention wont be on point, I'll have them done tomorrow. Also I wanted to suggest to maybe open more tasks for UI improvements? I would like to make more things pretty and also add a dark mode.

fozziethebeat: Sure, I can do a sweep through the current state of things and start listing more opportunities.

LucianPetri:
Ok, I added a TaskSelection, TaskOptions and TaskOption Component. Didn't want clutter the index so I threw everything in the TaskSelection Logic. Also merged the new paths and added them.

LucianPetri:
On the way, I also found that the Footer overlaps when on mobile screen while having multiple tasks. I Can do this after.

LucianPetri:
Also if There's to be added more stuff on index page, I have in mind another style to view the task selection to give more room for other stuff to be added. Hit me up with more details if required.

LucianPetri (commented):
Refactor UserChoice -> TaskSelection

fozziethebeat (approved):

PR #80, opened 2022-12-27
error importing the task handler

ottojoash:
when program tasked to  run it bring in an error "in <module>
    import task_handlers" which i have fixed


yk:
this is probably not something to be "fixed". I.e. I don't see how the additional whitespace would change import behavior in any way. Have you confirmed this is not an error in your setup? What's the error message you got?

PR #76, opened 2022-12-27
Unifying the task structure

fozziethebeat:
This moves the existing tasks to a simpler url hierarchy: `/evaluate` and `/create`.  We rename the pages to match the underlying task names in the backend.

Further, this adds some UI layout to the tasks so users can navigate to the home page without struggle and fixes the UI highlighting for the rating task.

PR #81, opened 2022-12-27
Implement `*_reply` for the web

### LINKED ISSUE #30: Implement `user_reply` for web
The `user_reply` task type needs to display these fields:
*  `conversation.messages`: An array of messages in a conversation.
*  `hint`: Some kind of hint.

It takes the following interaction type:
*  `type`: `text_reply_to_post`
*  `text`: A response to the conversation so far

### LINKED ISSUE #31: Implement `assistant_reply` for web
The `assistant_reply` task type needs to display these fields:
*  `conversation.messages`: An array of messages in request to an assistant.

It takes the following interaction type:
*  `type`: `text_reply_to_post`
*  `text`: A response to the conversation so far

AbdBarho:
Closes #30 
Closes #31

I have started to consolidate some of the re-used components to make the code easier to read, I did not change files modified in #76, this could be done in another pr.

I am unsure how "clean" the code should be for the mvp, but we can always clean it up later.

### DIFF
--- a/website/src/components/Messages.tsx
+++ b/website/src/components/Messages.tsx
@@ -1,0 +1,18 @@
+export interface Message {
+  text: string;
+  is_assistant: boolean;
+}
+
+const getColor = (isAssistant: boolean) => (isAssistant ? "bg-slate-800" : "bg-sky-900");
+
+export const Messages = ({ messages }: { messages: Message[] }) => {
+  const items = messages.map(({ text, is_assistant }: Message, i: number) => {
+    return (
+      <div key={i + text} className={`${getColor(is_assistant)} p-4 my-1 rounded-xl text-white whitespace-pre-wrap`}>
+        {text}
+      </div>
+    );
+  });
+  // Maybe also show a legend of the colors?
+  return <>{items}</>;
+};
--- a/website/src/pages/create/assistant_reply.tsx
+++ b/website/src/pages/create/assistant_reply.tsx
@@ -1,0 +1,83 @@
+import { Textarea } from "@chakra-ui/react";
+import { useRef, useState } from "react";
+import useSWRMutation from "swr/mutation";
+import useSWRImmutable from "swr/immutable";
+
+import fetcher from "src/lib/fetcher";
+import poster from "src/lib/poster";
+import { Messages } from "src/components/Messages";
+import { TwoColumns } from "src/components/TwoColumns";
+import { Button } from "src/components/Button";
+
+const AssistantReply = () => {
+  const [tasks, setTasks] = useState([]);
+
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+
+  const { isLoading } = useSWRImmutable("/api/new_task/assistant_reply ", fetcher, {
+    onSuccess: (data) => {
+      console.log(data);
+      setTasks([data]);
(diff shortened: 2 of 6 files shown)


website/src/pages/create/assistant_reply.tsx
```diff
@@ -0,0 +1,98 @@
+import { Textarea } from "@chakra-ui/react";
+import { useRef, useState } from "react";
+import useSWRMutation from "swr/mutation";
+import useSWRImmutable from "swr/immutable";
+
+import fetcher from "src/lib/fetcher";
+import poster from "src/lib/poster";
+import { Messages } from "@/components/Messages";
```
fozziethebeat, line 8: Can you import the local packages in a consistent way? either `src/` or `@/`?

website/src/pages/create/user_reply.tsx
```diff
@@ -0,0 +1,99 @@
+import { Textarea } from "@chakra-ui/react";
+import { useRef, useState } from "react";
+import useSWRMutation from "swr/mutation";
+import useSWRImmutable from "swr/immutable";
+
+import fetcher from "src/lib/fetcher";
+import poster from "src/lib/poster";
+import { Messages } from "@/components/Messages";
```
fozziethebeat, line 8: Same here regarding consistent imports

website/src/pages/create/user_reply.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </Button>
+        </div>
+      </section>
+    </div>
+  );
+};
+
+export default UserReply;
+
+const Button = (props: React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>) => {
```
fozziethebeat, line 85: This button looks pretty similar to the other task's button.  Can we make it a shared component?

fozziethebeat (approved):

website/src/components/Messages.tsx
```diff
@@ -0,0 +1,18 @@
+export interface Message {
+  text: string;
+  is_assistant: boolean;
+}
+
+const getColor = (isAssistant: boolean) => (isAssistant ? "slate-800" : "sky-900");
+
+export const Messages = ({ messages }: { messages: Message[] }) => {
```
fozziethebeat, line 8: For a later date, this seems like a viable component for Chakra-ui style style variants.  I have an example of this in a test library I made in https://github.com/SurfaceData/surface-react-components/blob/main/src/themes/surface/components/button.ts and https://github.com/SurfaceData/surface-react-components/blob/main/src/Button/Button.tsx.

This would cut down the code and let us drop the `getColor` method.  But we can do that cleanup later.

website/src/pages/create/assistant_reply.tsx
AbdBarho: vscode inserted all the `@`s, probably from `jsconfig.json`.

I will reset to `src/` to consistent with other files

website/src/pages/create/user_reply.tsx
AbdBarho: done

fozziethebeat:
Merging this now as I see changes have been made.

PR #82, opened 2022-12-27
Add per-page layout

AbdBarho:
This PR builds on the discussion in #76 

Some layout fixes are necessary, mainly removing the `h-full mx-auto`.

### DIFF
--- a/website/src/components/Layout.tsx
+++ b/website/src/components/Layout.tsx
@@ -1,0 +1,20 @@
+// https://nextjs.org/docs/basic-features/layouts
+
+import type { NextPage } from "next";
+
+import { Footer } from "./Footer";
+import { Header } from "./Header";
+
+export type NextPageWithLayout<P = {}, IP = P> = NextPage<P, IP> & {
+  getLayout?: (page: React.ReactElement) => React.ReactNode;
+};
+
+export const getDefaultLayout = (page: React.ReactElement) => (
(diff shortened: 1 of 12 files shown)


fozziethebeat:
I'll review this after my PR finally goes in.

website/src/components/Layout.tsx
```diff
@@ -0,0 +1,18 @@
...
+import type { NextPage } from "next";
+
+import { Footer } from "./Footer";
+import { Header } from "./Header";
+
+export type NextPageWithLayout<P = {}, IP = P> = NextPage<P, IP> & {
+  getLayout?: (page: React.ReactElement) => React.ReactNode;
+};
+
+export const getLayout = (page: React.ReactElement) => (
```
fozziethebeat, line 12: This component is pretty magical.  Can you add some high level comments on how it works.  My understanding is that any new page component should set a `getLayout` field and use this method.  That'll use this one unified layout.

Should we name this as something a little more specific like `getDefaultLayout` in the situation where we have another layout?

fozziethebeat (commented):
Looks good.  Will approve after you've merged in my PR changes.

website/src/components/Layout.tsx
AbdBarho: You are totally right, I changed it now so it is the default layout (you don't have to specify anything in the components), with the option of overriding the layout if needed.

fozziethebeat (approved):

Issue #72, opened 2022-12-26
Create new `journal` table and log all human-feedback interactions to it


andreaskoepf:
We want to generate statistics about recent user interactions and calculate scores for a leaderboard (gamification). As a first step we will log all important user interactions into a 'journal' table.

`journal` table (event log table):
```
id: uuid # (time ordered UUIDs as PK)
created_date: datetime # (redundant to pk but simplifies debugging)
person_id: uuid # (nullable) user is called person in db
post_id: uuid # (nullable)
event_type: string
event_payload: jsonb #serialized event object
api_client_id: uuid
```

`journal_integration` table (track state of integration processes):
```
worker_id: string (name of integration worker process)
last_processed_id: uuid
last_run: datetime
last_error: string  # debug
failing_journal_id: uuid
```

andreaskoepf:
First impl of journal is ready.

PR #89, opened 2022-12-27
add journal table and JournalWriter helper class

andreaskoepf:
Log user responses chronologically in a journal table #72. Ratings and Rankings are directly stored in the journal, text-replies currently not.


backend/oasst_backend/journal_writer.py
```diff
@@ -0,0 +1,122 @@
...
+                event_type = type(payload).__name__
+
+        if payload.person_id is None:
+            payload.person_id = self.person_id
+        if payload.post_id is None:
+            payload.post_id = post_id
+        if payload.workpackage_id is None:
+            payload.workpackage_id = workpackage_id
+        if payload.task_type is None:
+            payload.task_type = task_type
```
croumegous, line 107: You can write it in a shorter way : 
```python 
if event_type is None:
    event_type = "null" if payload is None else type(payload).__name__

payload.person_id = payload.person_id or self.person_id
payload.post_id = payload.post_id or post_id
payload.workpackage_id = payload.workpackage_id or workpackage_id
payload.task_type = payload.task_type or task_type
```

yk (approved):

backend/oasst_backend/models/journal.py
```diff
@@ -0,0 +1,56 @@
...
+    (time_low, time_mid, time_hi_version, clock_seq_hi_variant, clock_seq_low, node) = uuid1(node, clock_seq).fields
+    # reconstruct 60 bit timestamp, see version 1 uuid: https://www.rfc-editor.org/rfc/rfc4122
+    timestamp = (time_hi_version & 0xFFF) << 48 | (time_mid << 32) | time_low
+    version = time_hi_version >> 12
+    assert version == 1
+    a = timestamp >> 28  # bits 28-59
+    b = (timestamp >> 12) & 0xFFFF  # bits 12-27
+    c = timestamp & 0xFFF  # bits 0-11 (clear version bits)
+    clock_seq_hi_variant &= 0xF  # (clear variant bits)
+    return UUID(fields=(a, b, c, clock_seq_hi_variant, clock_seq_low, node), version=None)
```
croumegous, line 24: What is the point of using time ordered uuid?  Shouldn't we rely on `created_date` if we ever need to do something related to the time (e.g. weekly leaderboard).

andreaskoepf: You may know the snowflake datatype (as used by twitter or discord). The idea here is the same: Time ordered unique ids that can be client-side generated on different machines. It will be used by integrator processes to track up to which point in time events have been processed.

backend/oasst_backend/journal_writer.py
andreaskoepf: Thank for the tip. Just to keep it short I could even also write 
`event_type = payload and type(payload).__name__ or "null"` 
It is a matter of taste ... I am personally not the biggest fan of the python ternary operator.

PR #96, opened 2022-12-28
fix: upsert user reaction instead of inserting

croumegous:
Hello, 
I played with the end to end demo and saw a bug in the `/grading/grade-output` page.  
If I submit a rating twice the second time I submit it I get the following error in the backend: 
```python
sqlalchemy.exc.IntegrityError: (psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "post_reaction_pkey"
DETAIL:  Key (post_id, person_id)=(702250a7-0f3b-4469-9e86-ab7de669e832, e5602260-05cf-4a4b-8926-7fe53f09a6dc) already exists.
```
Indeed we are trying to do an insert instead of an update.
I understand that in the end a user may not be supposed to submit a rating twice on the website (page updated once submited).

But depending on the frontend it can be nice to have an update instead (e.g. discord)
openassistant_upsert
(In the discord example when I try to "update" the rating it failed)

I don't know if this is also nice to have for `insert_post`, let me know if you want me to also change it.

### DIFF
--- a/backend/oasst_backend/prompt_repository.py
+++ b/backend/oasst_backend/prompt_repository.py
@@ -190,7 +190,7 @@ class PromptRepository:
 
         # store reaction to message
         reaction_payload = db_payload.RatingReactionPayload(rating=rating.rating)
-        reaction = self.insert_reaction(message.id, reaction_payload)
+        reaction = self.upsert_reaction(message.id, reaction_payload)
         if not task.collective:
             task.done = True
             self.db.add(task)
@@ -223,7 +223,7 @@ class PromptRepository:
 
                 # store reaction to message
                 reaction_payload = db_payload.RankingReactionPayload(ranking=ranking.ranking)
-                reaction = self.insert_reaction(task.id, reaction_payload)
+                reaction = self.upsert_reaction(task.id, reaction_payload)
                 # TODO: resolve message_id
                 self.journal.log_ranking(task, message_id=None, ranking=ranking.ranking)
 
@@ -241,7 +241,7 @@ class PromptRepository:
 
                 # store reaction to message
                 reaction_payload = db_payload.RankingReactionPayload(ranking=ranking.ranking)
-                reaction = self.insert_reaction(task.id, reaction_payload)
+                reaction = self.upsert_reaction(task.id, reaction_payload)
                 # TODO: resolve message_id
                 self.journal.log_ranking(task, message_id=None, ranking=ranking.ranking)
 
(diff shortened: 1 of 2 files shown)


fozziethebeat:
Thank you for catching this! Yes, that is a problem i've noticed as well.

backend/oasst_backend/prompt_repository.py
```diff
@@ -308,18 +308,22 @@ def insert_post(
...
         return post
 
-    def insert_reaction(self, post_id: UUID, payload: db_payload.ReactionPayload) -> PostReaction:
+    def upsert_reaction(self, post_id: UUID, payload: db_payload.ReactionPayload) -> PostReaction:
         if self.person_id is None:
             raise ValueError("User required")
 
         container = PayloadContainer(payload=payload)
-        reaction = PostReaction(
+        new_reaction = PostReaction(
```
andreaskoepf, line 316: It would make sense to move the construction of the new `PostReaction` obj into the `if existing_reaction is not None:` block.

andreaskoepf (changes requested):
In general inseart-or-update is a good idea.

We have to take notice that updates will invalidate previous entries entries in the event-journal, it becomes harder to incrementally compute the item scores .. e.g. probably need to query all effective post_reactions and can't rely on the journal data.

backend/oasst_backend/prompt_repository.py
```diff
@@ -308,18 +308,22 @@ def insert_post(
...
             post_id=post_id,
             person_id=self.person_id,
             payload=container,
             api_client_id=self.api_client.id,
             payload_type=type(payload).__name__,
         )
+        existing_reaction = self.db.query(PostReaction).filter_by(post_id=post_id, person_id=self.person_id).one_or_none()
```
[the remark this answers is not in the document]
croumegous: Done :+1:

backend/oasst_backend/prompt_repository.py
croumegous: Done :+1:

croumegous:
I just rebased my branch with latest commits

yk:
Did you see this?
> We have to take notice that updates will invalidate previous entries in the event-journal, e.g. it becomes harder to incrementally compute the item scores .. e.g. we need to query all effective post_reactions and can't rely on the journal data.
the way it looks to me now, even if an `update` happens, the `log_ranking` of the journal is still called, e.g. the old reaction log would still be in the journal. I don't have a good overview over the codebase, do you think I'm overlooking something here?

croumegous:
I added a new commit to remove the entry in the journal table when we update an existing user reaction.  
However, I understand if you have concerns about implementing this change and prefer to stick with the current behaviour of not allowing upserts

yk:
this is a bit above my competency to decide. @andreaskoepf opinions?

53jk1 (approved):

Issue #44, opened 2022-12-19
Make user-submitted data available for new tasks


yk:
We want to be able to run fully on user-submitted data. Thus, when a user is asked to provide a prompt, and does so, we would like to be able to then re-use that prompt for other tasks, for example the "rank prompts" tasks, but also the "act as assistant" task, where a user is asked to answer as if they were the assistant to a given prompt. That prompt should come from the database of user-submitted prompts.

Extensions to this could be that the user-submitted prompts are sampled according to how well they are ranked against other submitted prompts in the ranking task, but it's not necessary for start.

The same requirement exists for when users submit user-answers or assistant-answers, we would also like to be able to re-use those as data for further tasks (so we can build up conversations over time, one message per task).

mjagkow:
Hi, I would like to contribute to this task, but before this I have a couple of questions to ask:

- What do you consider to be a thread? From what I can see right now in the code a thread is only formed by a pair of ("task-started-post", "user-reply-post"). It feels like all messages of one conversation should share the same thread, is it correct?
- Do you consider only linear conversations? With parent post id one could construct a number of conversation variants branching from one or more intermediate posts. 

Also are there any other docs describing the algorithm, data scheme and interaction protocol you would like to implement besides the readmes and high level protocol architecture article on Notion? This would help a lot in understanding :)

yk:
hey thanks so much for being here :)

> What do you consider to be a thread? From what I can see right now in the code a thread is only formed by a pair of ("task-started-post", "user-reply-post"). It feels like all messages of one conversation should share the same thread, is it correct?

the idea (which might not yet be fully reflected in the code) is that the interactions of the system with the users result in tree-structures. At the root is a prompt, which a user has entered. We then send out this prompt as a task to multiple users, each of which is tasked to "play" the assistant and give a response. Given those responses, we can do two things: First, we can have other people rank the responses, which will be useful for training the reward model. Second, we can take each response and form a thread from (initial prompt, assistant response), and give each of these threads as a task to multiple users. Thus, each thread results in multiple replies from users (who are now "playing" the user). Again, we can rank those and also combine each (initial prompt, assistant reply, user reply) as one thread to continue the tasks.

So, in my view, a thread is a chain of messages between user and assistant. We build threads by going down one path of the tree we collect by people fulfilling their tasks. I hope that makes it a bit more clear and answers the second question: The result is always a linear conversation, but during data collection, we collect a tree.

[...]

PR #115, opened 2022-12-28
#44 Sampling tasks from user data

mjagkow:
Tasks are now sampled randomly from user-uploaded data.

- Queries to fetch random post threads and conversations
- Moved "frontend task message" id to WorkPackage
- Added references for thread_id, parent_post_id to WorkPackage
- Added "ack" and "done" flags to WorkPackage
- PostReaction FK changed to WorkPackage
- Added depth and children_count to Post


backend/oasst_backend/models/post.py
```diff
@@ -31,3 +31,5 @@ class Post(SQLModel, table=True):
     )
     payload_type: str = Field(nullable=False, max_length=200)
     payload: PayloadContainer = Field(sa_column=sa.Column(payload_column_type(PayloadContainer), nullable=True))
+    depth: int = Field(sa_column=sa.Column(sa.Integer, default=0, server_default=sa.text("0"), nullable=False))
```
yk, line 34: do we really need all these attributes?
it seems like sqlmodel might do the right thing with `depth: int = 0` without any field definition at all, no?

mjagkow: There was an issue adding default boolean value (ack, done), the value did not propagate properly to existing rows. Solution was to specify server_default, so I did it the same way to other columns.

yk:
LGTM, but I'll leave the review to @andreaskoepf bc more familiar with this code

andreaskoepf (approved):
My thoughts:
1. depth and children_count are incrementally calculated, e.g. we must from now on not directly delete elements from post without going through python code that updates the parent element.
2. regarding sampling & `order by random()` I have read different reports about its performance online (some say it "performs a full table scan and also ordering. Therefore this method is not preferred for tables with large number of rows because of performance reasons." ..) Postgres also supports TABLESAMPLE, which is probably faster.

In general we will need very soon a more sophisticated sampling function that takes a weights into account.

andreaskoepf:
linking #44

PR #116, opened 2022-12-28
Add build website workflow

### LINKED ISSUE #112: Ensure `npm run build` finishes successfully before pull requests are ready for merging
Before we merge in changes to the web, we should ensure `npm run build` finishes successfully to catch any errors that would block building docker images.

AbdBarho:
Closes #112


fozziethebeat (approved):

yk (approved):
thank you! will not prevent bad merges, but at least we get yelled at when builds break.

PR #119, opened 2022-12-28
Unified multiple Docker Compose files into one for easier usage

croumegous:
Remove all the different docker compose files in scripts folder, to use a single docker compose (in root directory).
Also moved documentation to run the end to end demo in the main readme to better onboard new users.


docker-compose.yaml
```diff
@@ -0,0 +1,101 @@
+version: "3.7"
+
```
fozziethebeat, line 2: I assume `docker compose up` would bring up the full stack yes?

fozziethebeat (approved):
I agree that this is quite a bit cleaner and resolves a lot of cross compose referencing.  As long as my question/comment does what I think it does, I like this.

You'll still need review from Yannic or Andreas to finish.

docker-compose.yaml
croumegous: Yes, exactly

yk (approved):
awesome, thank you very much!

PR #120, opened 2022-12-28
Add `OasstError` exception class and corresponding exception filter

### LINKED ISSUE #55: Better error feedback to the frontends
Currently, the backend responds with plain 400 messages to protocol errors, for example when a frontend responds to a ranking task of 3 items with a list of only 2 items. Ideally, the backend would send a helpful message to display to the user along with the 400 response.

andreaskoepf:
This PR introduces the exception class `OasstError` that can always be used to signal a backend error to the API client with message and open-assistant error_code. It should not be used when the message could leak security relevant information. OasstError is 'picklable'.

The ctor of `OasstError` takes parameters `message: str, error_code: int, http_status_code: int = HTTP_400_BAD_REQUEST`. The `error-code` is a numeric reference to the specific error-site in the backend.

An exception handler is registered which generates a json response with `message` and `error_code` (using a http-status code as specified by the OasstError class).

Resolves #55


backend/main.py
```diff
@@ -7,10 +7,20 @@
...
 from oasst_backend.config import settings
+from oasst_backend.exceptions import OasstError
 from starlette.middleware.cors import CORSMiddleware
 
 app = fastapi.FastAPI(title=settings.PROJECT_NAME, openapi_url=f"{settings.API_V1_STR}/openapi.json")
 
+
+@app.exception_handler(OasstError)
+async def oasst_exception_handler(request: fastapi.Request, ex: OasstError):
+    logger.error(f"{request.method} {request.url} failed: {repr(ex)}")
```
yk, line 18: loguru also has `logger.exception` that gives a nice trace of the exception (in the current context) and annotates all parameters with their values, it's really nice to debug, but also very verbose, maybe worth thinking about

backend/oasst_backend/error_codes.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+"""
+Open-Assistant backend API error codes.
```
yk, line 3: suggestion: make these into an IntEnum, so they're still int, but typing could be used

yk (approved):
LGTM, left just small comments

backend/main.py
andreaskoepf: Yes, I had exactly that first but removed it again: For OasstErrors it is clear where they come from, e.g. they are usually post translation of generic internal errors. We could think about referencing 'inner exceptions' in the class that could be forwarded to the client when a server debug setting is True.

I was also looking into general handlers for uncaught Exceptions. IMO it would make sense to remove the general `excet Exception` handlers and do the translation into 500 Internal Server Error centralized .. there `logger.exception()` would definitely make sense.

backend/oasst_backend/error_codes.py
yk: yes I don't think any of that would change with an `IntEnum`. it would just be nicer to instead of having to take an `int` parameter, be able to take and `OAsstErrorCode` parameter (that behaves like an int)

andreaskoepf: (regarding typing .. I normally put such an exception class in a shared lib, it is int to prevent dependencies between the base exception class and each and every error_code that would be defined). It would make sense to derive exception classes from it .. but I am not sure if the exception-filter type check in fast-api is polymorphic. I will try that.

yk: like this: https://docs.python.org/3/library/http.html#http.HTTPStatus

andreaskoepf: I have now used a simple `IntEnum` for the error codes.

Important: There was an error in `api_auth()`, probably introduced when `DEBUG_SKIP_API_KEY_CHECK` was added,
see https://github.com/LAION-AI/Open-Assistant/blob/f208f65fafcac24f6c513dc80cbbcb4fb75178d4/backend/oasst_backend/api/deps.py#L55-L57 ... if the DB query returned None the function returned None (instead of raising an Unauthorized error...)

andreaskoepf: To clarify: the `api_auth()` problem occurred when the `api_key` value sent with the request was not empty and both `settings.DEBUG_SKIP_API_KEY_CHECK == settings.DEBUG_SKIP_API_KEY_CHECK == False` and there was no matching or enabled api_client registered in the DB.

PR #122, opened 2022-12-28
Slightly better README.md

MohamedAliRashad:
My take on the simplicity of experiencing the project


ericsomto (changes requested):
Address readme before merging

yk (commented):
thank you very much. I've just made #123 which might conflict with this one.

while we are improving, do you think you could also remove the period between the first two sentences and add a comma?

MohamedAliRashad:
@yk I fixed what you ask

yk:
thank you :) now just resolve the merge conflicts then we're perfect

yk:
Thank you! We don't need the email link thing anymore, but I can remove that later. Small improvements like this really make the experience better for newcomers :)

yk (approved):

PR #123, opened 2022-12-28
added environment variable DEBUG_LOGIN to display debug credentials provider even in production mode

yk:
background: debug login is always active when `NODE_ENV=development`, but one cannot set this variable manually in a production build, so this new environment variable is needed to run the end-to-end demo, which uses the docker (production) build.


website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -33,7 +34,7 @@ if (process.env.DISCORD_CLIENT_ID) {
   );
 }
 
-if (process.env.NODE_ENV === "development") {
+if (boolean(process.env.DEBUG_LOGIN) || process.env.NODE_ENV === "development") {
```
Comment (AbdBarho) [inline website/src/pages/api/auth/[...nextauth].ts:37]: LGTM

The name `DEBUG_LOGIN` does not really represent what the variables does, maybe `USE_DUMMY_USER`, or `USE_DEFAULT_CREDENTIALS` or something similar?

Unless I misunderstood something...

Comment (yk) [reply website/src/pages/api/auth/[...nextauth].ts:37]: there's not really a dummy user or default credentials, it's the login itself that's in a debug configuration: it just takes any username and gives you a session as that user.

Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts:37]: aha, I see, so I did misunderstand it 😅

fozziethebeat (approved):
Makes sense to me

PR #91, opened 2022-12-27
updated avatar component name, added ability to change username (only works server side for now), updated signin page, updated Footer

rsandb:
A bunch of these commits were already added so I'm not sure why they're showing up here. The only one that I'm requesting is "[de86903](https://github.com/LAION-AI/Open-Assistant/pull/91/commits/de869033bf9edf3c051d01c264637b3d286a65ae)"

PR #130, opened 2022-12-28
Add missing type submit to login page

AbdBarho:


fozziethebeat (approved):
I see there's some merge conflict, I trust you'll fix it without problem

AbdBarho:
Already fixed in #91

Issue #40, opened 2022-12-19
Introduce Text Labels to the protocol


yk:
Text labels are labels that can be assigned to any piece of text the user interacts with.
Examples are:
- contains toxic language
- encourages illegal activity
- good quality
- bad quality
- is spam

These should be introduced into the protocol

PR #135, opened 2022-12-29
Rewrite with `hikari`

AlexanderHOtt:
This is a basic structure for the discord bot. None of the features have been implemented yet.


discord-bot/bot/config.py
```diff
@@ -0,0 +1,37 @@
...
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Config:
```
andreaskoepf, line 16: Could you please use a [pydantic settings](https://docs.pydantic.dev/usage/settings/) class here?

discord-bot/flake8-requirements.txt
```diff
@@ -0,0 +1,26 @@
+flake8==6.0.0
+
+# Plugins
+
+Flake8-pyproject                 # use the pyproject.toml as the config file
+flake8-bandit                    # runs bandit
```
andreaskoepf, line 6: Is this necessary? we have project-wide settings in place via pre-commit...

discord-bot/noxfile.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding=utf-8 -*-
+"""Automated linting, formatting, and typechecking."""
+import nox
+from nox.sessions import Session
+
+
```
andreaskoepf, line 6: we use pre-commit .. please do not add other project-specific style enforcement tools!

discord-bot/pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[project]
+name = "Open-Assistant Discord Bot"
+version = "0.0.1"
+
+[tool.black]
```
andreaskoepf, line 5: same goes here .. please no new toml file!

discord-bot/bot/extensions/example.py
```diff
@@ -0,0 +1,406 @@
+# -*- coding: utf-8 -*-
+"""Example plugins for reference.
+
```
andreaskoepf, line 3: why do we need this 406 lines example file in the project? Could you maybe link this in a readme?

discord-bot/bot/config.py
```diff
@@ -0,0 +1,37 @@
...
+        token = getenv("TOKEN", None)
+
+        if token is None:
+            logger.error("Invalid token, please set the TOKEN environment variable.")
+            exit(1)
+
+        return cls(
+            token=token,
+            declare_global_commands=int(getenv("DECLARE_GLOBAL_COMMANDS", 0)),
+            owner_ids=[int(x) for x in getenv("OWNER_IDS", "").split(",")],
```
andreaskoepf, line 35: see example from pydantic docs...
```
class Settings(BaseSettings):
    numbers: list[int]

    class Config:
        @classmethod
        def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
            if field_name == 'numbers':
                return [int(x) for x in raw_val.split(',')]
            return cls.json_loads(raw_val)
```

andreaskoepf:
We will keep this PR open until we hit feature parity.

discord-bot/bot/extensions/tasks.py
```diff
@@ -0,0 +1,304 @@
...
+        await ctx.respond("Timed out waiting for response")
+        return
+
+    if event.content is None:
+        await ctx.respond("No content in message")
+        return
+    ratings = event.content.replace(" ", "").split(",")
+
+    # Check if the ratings are valid
+    if len(ratings) != 5:
```
andreaskoepf, line 274: this is not always 5, it depends on the number of elements in the task.

discord-bot/bot/extensions/tasks.py
```diff
@@ -0,0 +1,304 @@
...
+        self.presses = []
+        await ctx.respond(f"Received response: {button.label}", flags=hikari.MessageFlag.EPHEMERAL)
+
+
+class SelectRating(miru.View):
+    """View for rating a task with a select menu."""
+
+    @miru.select(
+        options=[
+            hikari.SelectMenuOption(
```
andreaskoepf, line 219: There is `scale: RatingScale` that defines min and max for the rating.

discord-bot/bot/extensions/tasks.py
```diff
@@ -0,0 +1,304 @@
...
+    choices=[hikari.CommandChoice(name=task.split(".")[-1], value=task) for task in TaskRequestType],
+    required=False,
+    default=TaskRequestType.summarize_story,
+    type=str,
+)
+@lightbulb.command("task_dm", "Request a task from the backend.", ephemeral=True)
+@lightbulb.implements(lightbulb.SlashCommand, lightbulb.PrefixCommand)
+async def task_dm(ctx: lightbulb.Context):
+    """Request a task from the backend."""
+    await ctx.respond("Please complete the task in your DMs")
```
andreaskoepf, line 81: these hard-coded message strings should ideally be moved outside the main code files for people to edit.

AlexanderHOtt:
I think the bot now has feature parity, correct me if I'm wrong.

# Tasks

- [x] `/work` to start a new task
(In the server)
image
(In DMs)
image
- [x] User can select which type of task (`rate_summary` and `summarize_story` are not implemented yet)
image
- [x] Correct API interaction though the api client.
- [x] Timeout/cancel handling (Accepting the task is a 1 min timeout, responding to the task is a 1 hour timeout)
image

- [x] Embed templates (these are not "quickly editable" by people who unfamiliar with code, but they look a lot better)
image


# Text Labels

- [ ] text labels through a message context menu
I'm not sure how to report these to the backend yet, but the frontend is working.
I chose to let the user edit what text they are labeling, but I do keep track if they edit the message. Should I change this?
image
image
image
image

# Log Channel Configuration
- [x] `/settings log_channel` to set the log channel the bot sends completed task messages to.
image
- [x] `/settings get` to get the current server settings
image


I'll start cleaning up the code, here are my plans so far, but please mention anything that is missing.

Immediate
- [ ] address `TODO` comments
- [ ] backend text labeling
- [ ] switch logging to `loguru` and add more + better log messages
- [ ] switch bot config to `pydantic`
- [ ] switch nox sessions to precommit-hooks
- [ ] move example.py to a markdown file
- [ ] document codebase structure
- [ ] docker setup (maybe?)

Quality of life / UI
- [ ] improve bot message wording
- [ ] switch most (if not all) bot messages to embeds
- [ ] move all user facing messages (embeds) to a central place
- [ ] add better images to embeds
- [ ] `/help`
- [ ] `/tutorial`

Waiting on
- [ ] leaderboard / points

AlexanderHOtt:
@yk you created #40, did you mean that users can add text labels to task responses or to _any_ message?

discord-bot/bot/settings.py
```diff
@@ -0,0 +1,15 @@
...
+class Settings(BaseSettings):
+    """Settings for the bot."""
+
+    token: str = Field(env="TOKEN", default="")
+    declare_global_commands: int = Field(env="DECLARE_GLOBAL_COMMANDS", default=0)
+    owner_ids: list[int] = Field(env="OWNER_IDS", default_factory=list)
+    prefix: str = Field(env="PREFIX", default="./")
+
+    class Config(BaseSettings.Config):
+        env_file = ".env"
```
andreaskoepf, line 15: Could you please add `case_sensitive = False`, I guess it is default in pydantic but just to be explicit (since unix env names normally capitalized).

andreaskoepf (commented):
LGTM. API settings should be read from conf and I am still not sure about the large example file. Is this something special that you wrote specific for the project or is it something which is on the net and we could simply link to it? Currently it looks like a bit of junk (e.g. we also do not include full manual and examples to all other libs we use).

The comments that I made regarding the messages, external string-table/templates etc. could be addressed later after merging this. We should review the overall interaction concept first.

AlexanderHOtt:
About the example file, it’s something I wrote to help new contributors get up to speed quickly with hikari. There are some examples online, but they are spread all over the place. I want to include it somewhere, even if that is in the project wiki or a gist.

discord-bot/bot/settings.py
yk: 1. I don't think there's a need to inherit form `BaseSettings.Config`, I think they're merged automatically
2. maybe it's worth (in the future) to make a base settings class of our own where we have things like case sensitivity and the presence of a `.env` file in.

yk:
> About the example file, it’s something I wrote to help new contributors get up to speed quickly with hikari. There are some examples online, but they are spread all over the place. I want to include it somewhere, even if that is in the project wiki or a gist.

I think the ideal way would be to put it into a gist, or maybe you already have it somewhere in an open repo, then link to that in `discord-bot/README.md`. Although I'm not particularly against having examples, but ideally these come in the form of unit tests, which I can see is a bit tricky with discord. So no strong opinions, but if the file exists somewhere already or if a gist is fine, then I think a link would be perfect. Otherwise including it seems ok.

discord-bot/bot/extensions/tasks.py
AlexanderHOtt: the `tasks.py` file was for testing different types of user input. If we decide to go with select menus for rating, I will update the code in `work.py`. I'll rename `task.py` to `task_tests.py` to make that more clear.

discord-bot/bot/settings.py
AlexanderHOtt: > 1. I don't think there's a need to inherit form `BaseSettings.Config`, I think they're merged automatically
> 2. maybe it's worth (in the future) to make a base settings class of our own where we have things like case sensitivity and the presence of a `.env` file in.

When I don't inherit, I get an error from my type-checker
```
"Config" overrides symbol of same name in class "BaseSettings"
  "discord-bot.bot.settings.Settings.Config" is incompatible with "pydantic.env_settings.BaseSettings.Config"
  Type "Type[Config]" cannot be assigned to type "Type[Config]"

Pylance[reportIncompatibleVariableOverride]
```
The message is a bit cryptic, but because we are subclassing `pydantic.BaseSettings`, we need to keep the types the same (including `class Config` which inherits from `pydantic.BaseConfig`, but we can't just inherit from `pydantic.BaseConfig`, we have to inherit from `pydantic.BaseSettings.Config`). This "defining a class inside a class" always confused me a bit, but I guess it works. I could add a `# type: ignore` comment, but inheriting make more sense to me.

andreaskoepf (approved):
OK, ready for bot replacement.

Issue #103, opened 2022-12-28
Design a nice loading screen component


fozziethebeat:
When we're fetching a task for a contributor, we have a pretty boring loading component.  It would be nice to have a visually appealing loading screen component we can re-use in all of our task pages or in general whenever some part of the UI is loading data.

This should probably use a Chakra progress loader or progress bar with some additional wrapping around it.

jojopirker:
Somewhat like this? Text would be optional...

https://user-images.githubusercontent.com/19796430/209836633-58530458-9bed-4f17-942d-de4069c1fb43.mov

yk:
this is a very clean solution! for really long waits, we could think of displaying little hints or something like that, but I'd guess for what we need right now, this seems very neat!

could you PR this?

fozziethebeat:
Yup! this is a solid solution

PR #110, opened 2022-12-28
Initial Storybook setup

jojopirker:
These changes enable to use `npm run storybook` to start **storybook** as mentioned in #102. 
There is an example story for the header.
Maybe should be moved to the components folder when doing #99.

What is **not finished**: 
- DataMocking
- fokus only on `components/` rather than `pages/`

What is **finished**:
- TailwindCSS and ChakraUI integration

Issue #104, opened 2022-12-28
Load a new task for users when reaching the done state


fozziethebeat:
Right now in the task implementations we get to the end state and don't do anything useful.  It would be better to cleanly reset the task, fetch a brand new task, and let the user keep contributing to the same task type.

This should affect everything under `website/src/pages/create` and `website/src/pages/evaluate`.  Ideally this is some typescript library/hook that each of these pages can call to make sure everything is done correctly or a re-usable component that somehow handles it all.

PR #138, opened 2022-12-29
New loading screen component

jojopirker:
This is an implementation of a new loading screen component as mentionend in #103.  I also replaced the 
```if (tasks.length == 0)``` with the `isLoading` from `useSWRImmutable`. Also includes a story for storybook, which showcases usage with and without text.

### CODE REVIEW

File: website/src/pages/create/user_reply.tsx [hunk 1]
```diff
@@ -8,6 +8,7 @@ import poster from "src/lib/poster";
 import { Messages } from "src/components/Messages";
 import { TwoColumns } from "src/components/TwoColumns";
 import { Button } from "src/components/Button";
+import { LoadingScreen } from "@/components/Loading/LoadingScreen";
```
AbdBarho, line 11: please use `src` instead of `@`

AbdBarho (commented):
Thanks, looks good.

We wanted to refactor the `isLoading` and `task.length` in #104.

I see we merge #110 first and then this one?

website/src/components/Loading/Loading.stories.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+  title: "Example/LoadingScreen",
+  component: LoadingScreen,
+  parameters: {
+    layout: "fullscreen",
+  },
+};
+
+const Template = (args) => <LoadingScreen {...args} />; //<><div>text</div><div className="max-w-500 mt-40 z-1000 h-full relative"></div></>;
```
fozziethebeat, line 11: the commented out div can probably be deleted

website/src/pages/create/user_reply.tsx
fozziethebeat: I think it's fine if vs code is doing it.  I'm going to delete this mapping from tsconfig soon so it stops happening.

fozziethebeat (commented):
I agree, let's merge in the storybook change and then this.  I'll approve after #110 goes in.

AbdBarho (approved):

fozziethebeat (approved):

PR #125, opened 2022-12-28
New eslintrc.json

Jac-Zac:
Closes #121 

New eslintrc.json which should include [this](https://eslint.org/docs/latest/rules/sort-imports#options) and also [those](https://typescript-eslint.io/getting-started/)

PR #142, opened 2022-12-29
Update eslint config

### LINKED ISSUE #121: Add eslint rules for consistent imports
To have consistent imports, add eslint rules to encourage (and perhaps enforce?) a preferred ordering.

The rules in question:

https://eslint.org/docs/latest/rules/sort-imports

Stick with the default eslint sorting rules ([link](https://eslint.org/docs/latest/rules/sort-imports#options)), in case we are not happy with it, we can change it later.

Notes:
Should we reformat all files at once?

AbdBarho:
Continuation of #125 

Closes #121

Update eslint config & fix all linting errors

We warn instead of error for sort imports, since we can't just `eslint --fix`.

There are some other [tools](https://github.com/trivago/prettier-plugin-sort-imports) that can do this, we could integrate them later.


website/src/lib/prismadb.ts
```diff
@@ -1,7 +1,7 @@
 import { PrismaClient } from "@prisma/client";
 
 declare global {
-  var prisma;
+  const prisma;
```
yk, line 4: the `const` here breaks build.

k-nearest-neighbor (approved):
LGTM and builds on latest

k-nearest-neighbor:
I think you just need to rebase to fix the Header.tsx merge

k-nearest-neighbor:
> I think you just need to rebase to fix the Header.tsx merge

NM, I resolved it through the github UI

PR #144, opened 2022-12-29
add db seed-data check on backend startup

### LINKED ISSUE #133: Insert two dummy conversations in an empty database
Determine if posts table is empty upon backend startup. If so, insert 4 dummy posts to emulate two mini-conversations, e.g the two examples from the [Prompting Guide](https://github.com/LAION-AI/Open-Assistant/blob/main/Prompting_Guide.md). This allows a smoother operation of the `/tasks/` REST endpoint  (e.g. when a random task is requested) in the very early phase after deployment and during development.

It could be placed in [backend/main.py](https://github.com/LAION-AI/Open-Assistant/blob/main/backend/main.py) similar to the alembic update.

andreaskoepf:
This inserts dummy conversations into the posts table if they are not present during startup. This behavior can be disabled by setting the `USE_SEED_DATA` config param to `False`.

Resolves #133 

(Turned out to be more struggle than I initially thought with all the ids. Also good way to review current API maybe.)

### DIFF
--- a/backend/oasst_backend/config.py
+++ b/backend/oasst_backend/config.py
@@ -33,6 +33,7 @@ class Settings(BaseSettings):
 
     BACKEND_CORS_ORIGINS: List[AnyHttpUrl] = []
     UPDATE_ALEMBIC: bool = True
+    USE_SEED_DATA: bool = True
 
     @validator("BACKEND_CORS_ORIGINS", pre=True)
     def assemble_cors_origins(cls, v: Union[str, List[str]]) -> Union[List[str], str]:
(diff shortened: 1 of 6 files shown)


backend/oasst_backend/config.py
```diff
@@ -33,6 +33,7 @@ def assemble_db_connection(cls, v: Optional[str], values: Dict[str, Any]) -> Any
 
     BACKEND_CORS_ORIGINS: List[AnyHttpUrl] = []
     UPDATE_ALEMBIC: bool = True
+    USE_SEED_DATA: bool = True
```
yk, line 36: could we set this to false, but set to true in the docker-compose?

yk (approved):

backend/oasst_backend/config.py
andreaskoepf: good idea, will do.

andreaskoepf: yes, is done.

AbdBarho:
Thank you ❤, this was blocking us on the frontend!

PR #146, opened 2022-12-29
Adding a specific layout for Home

k-nearest-neighbor:
* Adds a specific layout to Home page, so it can have (header / footer / grid / etc) choices independent of other pages.
* Makes the header transparent on home (it was originally transparent, and recently became white in a change aimed at making the header look better for signed-in pages)

Use this pattern any time a page needs a different layout:

`ThePage.getLayout = (page) => ( ... )`

### DIFF
--- a/website/src/components/Header/Header.tsx
+++ b/website/src/components/Header/Header.tsx
@@ -5,6 +5,7 @@ import Image from "next/image";
 import Link from "next/link";
 import { signOut, useSession } from "next-auth/react";
 import { FaUser, FaSignOutAlt } from "react-icons/fa";
+import clsx from "clsx";
 
 import { Container } from "src/components/Container";
 import { NavLinks } from "./NavLinks";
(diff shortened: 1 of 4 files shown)


website/src/components/Header/Header.tsx
```diff
@@ -53,9 +54,10 @@ function AccountButton() {
   );
 }
 
-export function Header() {
+export function Header(props) {
+  const transparent = props.transparent ?? false;
   return (
-    <header className="bg-white">
+    <header className={`${clsx({ "bg-transparent": transparent, "bg-white": !transparent })}`}>
```
AbdBarho, line 60: nit: much simpler:
```jsx
<header className={clsx(!transparent && "bg-white")}>
```

k-nearest-neighbor: That's much better. done.

AbdBarho (approved):

PR #151, opened 2022-12-29
Added readable exception if no posts were found for task

mjagkow:
Added code for an error where a task could not be formed due to there are no suitable posts


andreaskoepf (approved):
OK great, thanks!

Issue #132, opened 2022-12-29
"Join Us" and "FAQ" buttons do nothing when on Sign Up page


billray0259:
Join Us and FAQ buttons in the navbar should navigate the user to the appropriate pages when clicked but do nothing when on the Sign Up page.

<img width="1282" alt="image" src="https://user-images.githubusercontent.com/31375073/209886910-cf43f0ec-a2c1-4843-850c-3403076b40f4.png">

fozziethebeat:
Correct! These buttons needs to be either removed or updated to actually do something.  Would you like to make a PR to remove them for now?

billray0259:
Yes! I can remove the buttons.

PR #152, opened 2022-12-29
Updated join-us and FAQ links

billray0259:
It was easier to update the links than to remove the buttons on the sign-up page. 

resolves LAION-AI/Open-Assistant#132

### DIFF
--- a/website/src/components/Footer.tsx
+++ b/website/src/components/Footer.tsx
@@ -27,7 +27,7 @@ export function Footer() {
                   <Link href="#" aria-label="Our Team" className="hover:underline underline-offset-2">
                     Our Team
                   </Link>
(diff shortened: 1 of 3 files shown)


AbdBarho (approved):
@billray0259 thank you!

This solves the problem for the login / sign up page, but not for other pages.

That being said, I think the problem is much bigger in how we structure the site. In our [index.tsx](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/index.tsx) we show two different pages if the user is signed in vs not.

I think it would make sense to leave `/` as non-logged user, and maybe add a `/dashboard` that shows the available tasks.

@billray0259 what do you think?

This is totally out of scope of this PR, I  will create a follow up issue. #156

PR #155, opened 2022-12-29
Added support for collective tasks

mjagkow:
- When frontend requests for a new task it can specify the task is "collective". 
- Collective tasks are not closed automatically when an interaction has been submitted for them. 
- This allows to collect multiple interactions in public groups. This could also be used in personal tasks (e.g. in DM) to allow for multiple responses for one task/message.
- Task can be closed manually by frontend using `/api/v1/tasks/close` method.


andreaskoepf (approved):
ok, looks good. I think easiest would be to use the expiry option, but it is of course ok to have an explicit endpoint to close the task.

PR #162, opened 2022-12-29
Added trusted api client flag

mjagkow:
Dummy api client entry is created with trusted=True on startup.
Also manually fixed existing migration IDs which resulted in broken alembic upgrade due to migraions were added in parallel PRs.


backend/alembic/versions/2022_12_28_1824-ef0b52902560_added_lang_column_for_iso_639_1_codes.py
```diff
@@ -2,7 +2,7 @@
 """Added lang column for ISO-639-1 codes
 
 Revision ID: ef0b52902560
-Revises: 3358eb6834e6
+Revises: d24b37426857
```
andreaskoepf, line 5: thanks for fixing

andreaskoepf (approved):
perfect, thanks!

Issue #90, opened 2022-12-27
Implement dark mode on website


LucianPetri:
# Dark Mode Implementation 

## Reasons
- Night Owl users like me would like to have a way to switch the websites appearance to something more dark for the protection of our eyes. 
- With this implementation we inevitably also have the possibility to more easily theme the website. 
- Also while doing this we should create a suite of components to be reused
- This will also speed up development for others that only implement logic and expect the theming to work properly

## Description
- dark mode is a theme complete with a color scheme and Context Provider
- all components in Chakra-UI are tied to the Chakra ThemeProvider inside ChakraProvider
- to switch between light and dark mode a basic switch should be placed in the Header
- a light theme and dark theme has to be created and placed inside the styles folder ( naturally ) 
- Chakra components must be added to the page layout and checked that backgrounds and foreground change correctly
- any custom components should be wrapped with Chakra Components and be made reusable
- the dark mode switch should look like a sun in light mode and a half moon in dark mode
- the color mode should be initially taken from system and kept afterwards in LocalStorage
- check for any SSR Issues that might arrise from hold the color mode value

### Pages
- [x] [Assistant Reply](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/create/assistant_reply.tsx)
- [x] [User Reply](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/create/user_reply.tsx)
- [x] [Grade Output](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/grading/grade-output.tsx)
- [ ] [Score Leaderboard](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/leaderboard/score-leaderboard.tsx)
- [x] [Summarize Story](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/summarize/story.tsx) 
- [x] [404](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/404.tsx)
- [x] [Index](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/index.tsx)
- [x] [Sign Up](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/auth/signup.tsx)
- [x] [Verify](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/auth/verify.tsx)

### Components
- [x] UserMenu
- TBD

### Reusable Components that need to be implemented
- TBD

[...]

PR #167, opened 2022-12-30
Dark mode implementation

LucianPetri:
Dark Mode Implementation Details Here:  #90


k-nearest-neighbor:
This is a great start. There are a few areas that need to be addressed:

- UI regressions: Changes to light mode theme; Header functionality and appearance
- UI that isn't incorporated into dark mode yet / well.
- Aesthetics.

I'm going to go through this PR and highlight these, probably committing.

fozziethebeat:
This is a pretty big PR touching a lot of files.  Let's hold off merging other react component changes until this is reviewed and we're comfortable with it (or deciding some other resolution strategy).

AbdBarho:
I also agree that this is a great start. thank you @LucianPetri.

I would like to propose a strategy for moving forward with this:

Ideally, all dark mode handling should be done either be configuring the theme directly, or by having some conditional styling for the components under `/components` only. This should also include font-size, shadows, gaps, padding, etc...

Everything under `/pages` should only care about what is being shown, not its design, by using the aforementioned components.

This should make our work future proof, and changes to the design both easy and global.

I am not sure if browser support is there yet, or if chakra ui supports it, but [Container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) sound really promising for paddings margins and such...

website/src/pages/index.tsx
```diff
@@ -21,12 +21,12 @@ const Home = () => {
           content="Conversational AI for everyone. An open source project to create a chat enabled GPT LLM run by LAION and contributors around the world."
         />
       </Head>
-      {session ? (
+      {false ? (
```
k-nearest-neighbor, line 24: This debugging hack snuck in to my commit. Will be setting it back to session.

website/src/styles/Theme/index.ts
```diff
@@ -0,0 +1,29 @@
+import { type ThemeConfig, extendTheme } from "@chakra-ui/react";
+import { containerTheme } from "./components/Container";
```
fozziethebeat, line 2: Rather than import each components theme here can we create a `website/src/styles/Theme/components/index.ts` file.

With that we can simplify this as 
```
import * as components from "./components";
```

And then down below we don't need to define a `const components = { ...}`.  We can just use the import

website/src/components/Loading/LoadingScreen.jsx
```diff
@@ -1,12 +1,12 @@
-import { Progress } from "@chakra-ui/react";
+import { Container, Progress } from "@chakra-ui/react";
 
 export const LoadingScreen = ({ text }) => (
-  <div className="bg-slate-100">
+  <Container>
```
fozziethebeat, line 4: If you're just replacing `div` elements without the most similar chakra element, i'd suggest `Box` instead of `Container`

fozziethebeat (commented):
This looks pretty close to being ready.  What's stopping us from handling the  merge conflicts and checking it in as is?

fozziethebeat:
I took a look at this branch.  I think we should just check this in as is and then revise the small UI issues in later (tinier) PRs.

The key things to fix before approval:
- Fix the merge the conflicts above
- Make sure the action boxes in tasks all are visible.  Right now `rank_assistant_replies` is visible and looks great.  The others done (i'm guessing because of how that box is styled).  I think the fix is as simple as copy whatever is done in `rank_assistant_replies` to the other tasks.

k-nearest-neighbor:
Alright. All checks passing and ready.

fozziethebeat (approved):
Great to finally get this in.

I'll make some small targeted fixes for things in a follow up PR

PR #171, opened 2022-12-30
Data management api

### LINKED ISSUE #160: Add REST endpoint to query & delete posts (messages)
In the DB we are dealing with message trees. We need REST endpoints to query single messages, conversations & message-trees.

`post_id` here means PK of `post` table in db. The frontend-provided (only in combination with api_client_id unique) id will be called `frontend_post_id`.

##  Required functionality

- query a message by `post_id`
- query a message by `frontend_post_id`
- query a conversation given a post_id (list of messages traced to tree-root in order (starting at root))
- query all messages of a message tree given a `post_id` (lookup `thread_id` of message -> query whole tree) (unordered)
- query children/descendants given a parent_post_id (with max_depth)
- query longest conversation of a message tree (by providing a `post_id` of any message in the tree)
- query messages of tree with most direct children (by providing a `post_id` of any message in the tree)

- query by user_id, ordered by date with conditions: max_count (limit), after_date (optional) & before end_date (optional)
- query by created_date, max_count (limit), start_date (optional), end_date (optional), only_roots (optional)

Only trusted api_clients:
- delete message (e.g. by setting deleted flag to true)
- delete all messages of a user (select all user's messages and perform delete them)
- query stats: number of total entries in post table, number of threads (message trees)

Optional (using existing functions):
- sample random messages
- sample random thread

mjagkow:
Resolves #160


andreaskoepf:
@mjagkow nice job for #160! I would suggest to the rename `management.py` into `messages.py`  and Yannic started with `/tasks` (plural) therefore I suggest we do the same with `/messages`.  What do you think about further extracting the `/frontend_messages` handlers in a separate `frontend_messages.py` file? Our file-structure would then resemble the request-paths more closely.

mjagkow:
Ok, souds good to me, will change it. It will also take some time to test those endpoints, as of now it is still raw draft.

backend/oasst_backend/prompt_repository.py
```diff
@@ -488,3 +496,178 @@ def close_task(self, post_id: str, allow_personal_tasks: bool = False):
...
+        """
+        Pick messages from a collection so that the result makes a linear conversation
+        starting from a message tree root and up to the given message.
+        Returns an ordered list of messages starting from the message tree root.
+        """
+        if isinstance(messages, list):
+            messages = {m.id: m for m in messages}
+        if not isinstance(messages, dict):
+            # This should not normally happen
+            raise OasstError("Server error", OasstErrorCode.SERVER_ERROR, HTTPStatus.INTERNAL_SERVER_ERROR)
```
andreaskoepf, line 511: maybe add also a unique id here, to quickly identify the error-site from the error code. The text can stay generic if we do not want to expose to the user what went wrong.

backend/oasst_backend/prompt_repository.py
```diff
@@ -488,3 +496,178 @@ def close_task(self, post_id: str, allow_personal_tasks: bool = False):
...
+            messages = [messages]
+
+        ids = []
+        for message in messages:
+            if isinstance(message, UUID):
+                ids.append(message)
+            elif isinstance(message, Post):
+                ids.append(message.id)
+            else:
+                raise OasstError("Server error", OasstErrorCode.SERVER_ERROR, HTTPStatus.INTERNAL_SERVER_ERROR)
```
andreaskoepf, line 649: same here for the error-site code, I know it is a bit inconvenient to add unique error codes...

backend/oasst_backend/prompt_repository.py
```diff
@@ -488,3 +496,178 @@ def close_task(self, post_id: str, allow_personal_tasks: bool = False):
...
+                parent_ids = self.db.execute(query).scalars().all()
+
+        self.db.commit()
+
+    def get_stats(self):
+        deleted = self.db.query(Post.deleted, func.count()).group_by(Post.deleted)
+        nthreads = self.db.query(None, func.count(Post.id)).filter(Post.parent_id.is_(None))
+        query = deleted.union_all(nthreads)
+        result = {k: v for k, v in query.all()}
+        return {
```
andreaskoepf, line 668: - maybe use here the `SystemStats` class from protocol instead dict?
- return type annotation for get_stats() is missing

backend/oasst_backend/prompt_repository.py
```diff
@@ -488,3 +496,178 @@ def close_task(self, post_id: str, allow_personal_tasks: bool = False):
...
+
+        if desc:
+            messages = messages.order_by(Post.created_date.desc())
+        else:
+            messages = messages.order_by(Post.created_date.asc())
+
+        if limit is not None:
+            messages = messages.limit(limit)
+
+        # TODO: Pagination could be great at some point
```
andreaskoepf, line 635: the problem with pagination is always that it is only really useful with for messages .. otherwise which of the 10000 pages does the user really want to see? Also it puts stress on the database to find correct elements of visible page based on counts which means queries everything before the elements that are on the page etc. Therefore I think in many cases it is actually easier to step by name or date incrementally forward backward - this allows to page forward + backward but not to specify a page index .. but you could jump by prefix or start/end-date...

backend/oasst_backend/exceptions.py
```diff
@@ -17,6 +17,7 @@ class OasstErrorCode(IntEnum):
     GENERIC_ERROR = 0
     DATABASE_URI_NOT_SET = 1
     API_CLIENT_NOT_AUTHORIZED = 2
+    SERVER_ERROR = 3
```
andreaskoepf, line 20: ;-) that one is not much different from GENERIC_ERRER

backend/oasst_backend/api/v1/users.py
```diff
@@ -0,0 +1,60 @@
...
+from starlette.status import HTTP_200_OK
+
+router = APIRouter()
+
+
+@router.get("/{user_id}/messages")
+def query_user_messages(
+    user_id: UUID,
+    api_client_id: UUID = None,
+    max_count: int = Query(10, gt=0, le=25),
```
andreaskoepf, line 21: I think we allow a higher value here, e.g. `le=1000` ...

backend/oasst_backend/api/v1/messages.py
```diff
@@ -0,0 +1,159 @@
...
+from starlette.status import HTTP_200_OK
+
+router = APIRouter()
+
+
+@router.get("/")
+def query_messages(
+    username: str = None,
+    api_client_id: str = None,
+    max_count: int = Query(10, gt=0, le=25),
```
andreaskoepf, line 24: I think we can safely query/send a bit more data, e.g. `le=1000`.

backend/oasst_backend/api/v1/frontend_users.py
```diff
@@ -0,0 +1,60 @@
...
+from starlette.status import HTTP_200_OK
+
+router = APIRouter()
+
+
+@router.get("/{username}/messages")
+def query_frontend_user_messages(
+    username: str,
+    api_client_id: UUID = None,
+    max_count: int = Query(10, gt=0, le=25),
```
andreaskoepf, line 21: same here ... `le=1000`

andreaskoepf (changes requested):
Looks really good! Had only minor comments..

backend/oasst_backend/exceptions.py
mjagkow: In my opinion GENERIC_ERROR is an error we do not understand, while SERVER_ERROR is error due to our fault.

backend/oasst_backend/prompt_repository.py
mjagkow: This comment is just a reminder, not a commitment. It is definitely a subject to discussion, no pagination is intended at this moment. 
Stepping using date is just another way to paginate. I believe there are ways to introduce pagination without putting much stress on the db. On the other hand loading arbitrary many entries from the table could affect performance as well.

backend/oasst_backend/prompt_repository.py
mjagkow: I'd rather not, this is a type of a situation which would never happen if 1) typing is enforced in the code and controlled with linters and 2) there's sufficient validation on fastapi/pydantic level. 

How would you describe this error case? I'd say "something wrong happened and we don't really know why" :)
If you'd describe it as "we expected one type/value somewhere, but got something else", then why would you give such amount of details to the frontend (which is the ultimate reciever of these errors so far unless you read ALL the server logs manually)? What would it be for?

If you really want to control what errors arise at which line than I'd recommend look towards Sentry and/or Prometheus+Kibana/Grafana.

backend/oasst_backend/api/v1/users.py
mjagkow: Changed to 1000

backend/oasst_backend/api/v1/messages.py
mjagkow: +

backend/oasst_backend/api/v1/frontend_users.py
mjagkow: changed to 1000

backend/oasst_backend/prompt_repository.py
mjagkow: Same as above, I'd leave it as is, we have too little information about what happened here.

backend/oasst_backend/prompt_repository.py
mjagkow: Changed return type and value

backend/oasst_backend/prompt_repository.py
andreaskoepf: It's more a general idea to have unique codes for all code-lines that throw an OasstError (independent of types). I have tried different things and made good experience with the unique call-site id approach .. otherwise you can only get this information from a stack-trace which normally isn't shown to the client.

backend/oasst_backend/journal_writer.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, db: Session, api_client: ApiClient, user: User):
         self.user = user
         self.user_id = self.user.id if self.user else None
 
-    def log_text_reply(self, task: Task, message_id: UUID, role: str, length: int) -> Journal:
+    def log_text_reply(self, task: Task, message_id: Optional[UUID], role: str, length: int) -> Journal:
```
andreaskoepf, line 57: not clear to me why the message_id became Optional now, it is essential for the journal entries. We need to clarify this via discord.

andreaskoepf (approved):
Ready for merge. We can address further points in an additional PR.

Issue #175, opened 2022-12-30
Add backend API rate limiting


andreaskoepf:
Add a throttling system to the backend REST endpoints that tracks and limits the interaction frequency of users/api-clients. This will help us to prevent or slow down users/automated systems form flooding our database. Preferably we would like to limit on multiple keys, e.g. user and api_client (each with different thresholds).

1. Review existing libraries such as [laurentS/slowapi](https://github.com/laurentS/slowapi) & [fastapi-limiter](https://github.com/long2ice/fastapi-limiter) and select one (please justify your choice).
2. Make necessary changes in to existing backend endpoints

GraemeHarris:
@andreaskoepf Happy to pick this up if free - will add some local load testing and report in the PR

kiritowu:
The following are my two-cents after reading through the source code and documentations for slowapi and fastapi-limiter:

### Storage Options
- slowapi: Redis (Cluster), Memcached, MongoDB, In-built Memory
- fastapi-limiter: Redis

Since In-built Memory and MongoDB is eliminated due to scalability issue and relative latency (as compared to memory-based alternatives), the two storage options left are [Redis vs Memcached](https://aws.amazon.com/elasticache/redis-vs-memcached/) in which Redis seems to be a better alternative due to its Replication capabilities and Lua scripting.

### Code Usage
- slowapi: Python Decorator
- fastapi-limiter: FastAPI Dependencies

One limitation of slowapi is that the `request` argument must be [explicitly stated](https://github.com/laurentS/slowapi#limitations-and-known-issues) and hinted as `fastapi.Request` in the endpoints parameters like this.
```python
@limiter.limit("5/minute")
async def myendpoint(request: Request)
    pass
```
After reviewing the existing backend code, the request is now type hinted as sub-classes from `pydantic.BaseModel` to be benefited from fastapi's automated body conversion. However, this will run into problem with slowapi's syntax compatibility since it checks whether is the `request` argument a subclass for `starlette.Request` in these [few lines](https://github.com/laurentS/slowapi/blob/d52da267222dd54bc13d391ea8a6b84e8924eed4/slowapi/extension.py#L653-L657).

Solving this issue will require some changes to the naming convention of all our current routes like this:
```python
@router.post("/", response_model=protocol_schema.AnyTask)  # work with Union once more types are added
def request_task(
    *,
    db: Session = Depends(deps.get_db),
    api_key: APIKey = Depends(deps.get_api_key),
[...]

PR #187, opened 2022-12-30
Add basic redis setup for rate limiting

GraemeHarris:
As per https://github.com/LAION-AI/Open-Assistant/issues/175 redis setup to be used for rate limiting - maybe BE cache down the line. 

I have updated the docker compose file to include redis + redis insights. They're quite basic for now with a lot of settings that can be tweaked as per need / feedback. 

Ansible changes might need more testing


docker-compose.yaml
```diff
@@ -27,6 +27,29 @@ services:
...
+    volumes:
+      - ./redis.conf:/usr/local/etc/redis/redis.conf
+      - redis:/data
+  # insights host - redis:6379
+  redis-insights:
+    image: redislabs/redisinsight:latest
+    ports:
+      - 8001:8001
+    volumes:
+      - redis:/data
```
yk, line 51: do you think it's necessary to define a volume here? seems like for a test setup it's fine if the data goes down with the container, especially bc we don't use this redis as long-term storage

yk:
also, some merge conflicts

docker-compose.yaml
GraemeHarris: Fair point. Removed 👍

yk (changes requested):
could you remove the ansible part? so we can get this merged. can you also make a new issue for including redis in ansible? I'll take that myself and do it later.

andreaskoepf:
@yk is this ready for merge?

kiritowu:
@GraemeHarris do you intend to add `redis` (and `redis-insights`) to Line 7 in the `docker-compose.yml` such that both `redis` and `redis-insights` will be started up for `backend-dev` purpose?

yk (approved):
Thank you!

PR #188, opened 2022-12-31
Fixes for breaking changes in #182

k-nearest-neighbor:
This PR gathers any fixes needed from https://github.com/LAION-AI/Open-Assistant/pull/182.

### DIFF
--- a/website/src/pages/api/new_task/[task_type].ts
+++ b/website/src/pages/api/new_task/[task_type].ts
@@ -62,7 +62,7 @@ const handler = async (req, res) => {
       "Content-Type": "application/json",
     },
     body: JSON.stringify({
-      post_id: registeredTask.id,
+      message_id: registeredTask.id,
     }),
   });
   const ack = await ackRes.json();
--- a/website/src/pages/api/update_task.ts
+++ b/website/src/pages/api/update_task.ts
@@ -52,8 +52,8 @@ const handler = async (req, res) => {
         display_name: token.name || token.email,
         auth_method: "local",
       },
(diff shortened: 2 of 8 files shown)


k-nearest-neighbor:
Here is a summary of the renamings from #182 grouped by current status of fix


- **Fixed** "user_reply" -> "prompter_reply"
- **Fixed** "rank_user_replies" -> "rank_prompter_replies"

The API endpoints for these tasks was updated. For now I left the UI and filenames using the name "user". (e.g., the UI still says "**Reply As User**" instead of "**Reply as Prompter**") as I think that while "prompter" is a good model name, it might make for awkward UX as a user-facing string. However, if we want it to match in the UI ("**Reply as Prompter**"), I don't object. Let's just confirm first that we want the *UX* to call humans "prompters" instead of "users".

- **Fixed** post_id -> message_id  (in TaskAck, TaskClose)
- **Fixed** user_post_id -> user_message_id

Updated in `[task_type].ts` and `update_task.ts`.

- **Fixed** "post_rating" -> "message_rating"
- **Fixed** "post_ranking" -> "message_ranking"

Updated in `rank_user_replies.tsx` `rank_initial_promps.tsx`, `rank_assistant_replies.tsx` `update_tasks.tsx`

- **Fixed** "text_reply_to_post" -> "text_reply_to_message"

Updated in `assistant_reply.ts` `summarize_story.tsx` `user_reply.tsx`

fozziethebeat (approved):
Thanks for updating this!

Issue #42, opened 2022-12-19
Implement Text Labels in web frontend


yk:
Text labels are described in #40 .

Make sure the user can submit those for every post they see.

fozziethebeat:
This feature is now open for work.

I'll make a web API point to submit labels that complies with the backend change and then document it.  Then we can generalize the component in the current rating task and add it to every task.

BitterKanegul:
Hi @fozziethebeat , @yk , I would love to take this up!

fozziethebeat:
Ideally this will make the checklist code in the current [grading task](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/pages/grading/grade-output.tsx) a re-usable component that is revealed as a popover (hidden behind a flag icon).  Then, when a user has checked one or more tags, they get submitted and associated with the task id.

To be complete, this will require both the UI side and and the web rest end point setup so that the tags are sent to the FastAPI backend.

BitterKanegul:
One feature that was discussed in https://github.com/LAION-AI/Open-Assistant/issues/40 was allowing parts of the generated text to be annotated with various labels.

Do you think highlights and selection for the text is a way to go/ convert the text into a set of tags(each word is a tag) and allow the user to click on each?

yk:
I think most text labels don't necessarily make sense to restrict to individual words, so I'd prefer the approach where the user range-selects some text. I think preference should always be given (ui-wise) to putting labels onto an entire post, because being able to map labels not only to text, but to posts inside conversations gives us some extra possibilities to create datasets, but I also think sometimes it's not appropriate to label an entire post, therefore I suggested the ability to annotate this kind of arbitrary text.

BitterKanegul:
Interesting! I'll try to make it by default select an entire post, and if the user wishes, can change the selection to a substring of the post. Multiple labels on the same post/substring is also possible?

yk:
> Interesting! I'll try to make it by default select an entire post, and if the user wishes, can change the selection to a substring of the post. Multiple labels on the same post/substring is also possible?

sounds like it could work :) yes, any label can be assigned a value from 0 to 1 (ideally, for binary labels just use 0 and 1)

fozziethebeat:
I would say for the MVP and first PR we should just support text labels to an entire post.  After the MVP we can get fancy with sub-post labeling.  And we should be able to add multiple tags (i.e. a checkbox) for any given post.

PR #189, opened 2022-12-31
#42 Adding popover to flag text with labels

BitterKanegul:


website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+  Checkbox,
+  PopoverCloseButton,
+  PopoverArrow,
+  Flex,
+  Spacer,
+  Slider,
+  SliderTrack,
+  SliderFilledTrack,
+  SliderThumb,
+} from "@chakra-ui/react";
```
fozziethebeat, line 19: Can these be ordered alphabetically as follows:

1. Import any `@something` imports
2. Import any external libraries
3. Import local code from `src/`

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <QuestionMarkCircleIcon
+          className="flex h-5 w-5 ml-3 text-gray-400 group-hover:text-gray-500"
+          aria-hidden="true"
+        />
+      </a>
+    );
+  }
+
+  return (
+    <>
```
fozziethebeat, line 127: I think we can drop the surrounding `<> ...</>` since this is all in a single `<li>`

fozziethebeat (changes requested):
This is a great start.  I added some small cleanup requests then I'll approve.

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+  SliderFilledTrack,
+  SliderThumb,
+} from "@chakra-ui/react";
+import useSWRMutation from "swr/mutation";
+import { QuestionMarkCircleIcon, FlagIcon } from "@heroicons/react/20/solid";
+import { useState } from "react";
+
+export type TextFlagProps = {
+  text: string;
+  label_map;
```
AbdBarho, line 26: `label_map: Record<string, string>`?

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+  Slider,
+  SliderTrack,
+  SliderFilledTrack,
+  SliderThumb,
+} from "@chakra-ui/react";
+import useSWRMutation from "swr/mutation";
+import { QuestionMarkCircleIcon, FlagIcon } from "@heroicons/react/20/solid";
+import { useState } from "react";
+
+export type TextFlagProps = {
```
AbdBarho, line 24: This is not referenced anywhere?

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+      setValue: setValue,
+      isChecked: isChecked,
+      setChecked: setChecked,
+    };
+  });
+  const fetchData = () => {
+    let label_map: Map<string, number> = new Map();
+    label_refs
+      .filter((flagRef) => flagRef.isChecked)
+      .map((refs) => {
```
AbdBarho, line 64: maybe you want `forEach`? or replacee the entire thing with  `reduce`?

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const { trigger, isMutating } = useSWRMutation("/api/v1/text_labels", poster, {
+    onSuccess: () => {
+      setIsEditing.off;
+    },
+  });
+
+  const submitResponse = (label_data) => {
+    trigger(label_data);
+  };
+  const label_refs: FlagRefs[] = TEXT_LABEL_FLAGS.map((flag) => {
```
AbdBarho, line 49: this works now just fine because the flags are all constant, but in-case the number of flags is dynamic (i.e. comes from the server) this will error because of different number of hook calls.

Nothing to do here, just keep in mind for the future.

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,205 @@
...
+      </li>
+    </>
+  );
+}
+interface textFlagLabels {
+  attributeName: string;
+  labelText: string;
+  additionalExplanation?: string;
+}
+const TEXT_LABEL_FLAGS: textFlagLabels[] = [
```
AbdBarho, line 166: this is the same as in `rate_summary.tsx` right? should we extract it to some shared variable?

website/src/components/FlaggableElement.tsx
BitterKanegul: ahh I thought since we pass the props to the HTML element when used, need to export it!

website/src/components/FlaggableElement.tsx
BitterKanegul: thank you! have removed

website/src/components/FlaggableElement.tsx
BitterKanegul: thanks for the input! thought i was being clever, but yes I have changed to useState on array

website/src/components/FlaggableElement.tsx
BitterKanegul: thank you, I have made the changes

website/src/components/FlaggableElement.tsx
BitterKanegul: I have removed it now!

website/src/components/FlaggableElement.tsx
fozziethebeat: I think the simpler thing to do is in a follow up PR delete the code in `rate_summary.tsx` when this component is added to it.

fozziethebeat:
Assuming that a follow on PR will use this component in each of the tasks, this looks good once pre-commits are green.

yk:
@BitterKanegul run `pre-commit run --all-files` until it passes to make the linters happy.

website/src/components/FlaggableElement.tsx
```diff
@@ -0,0 +1,220 @@
...
+    },
+  });
+
+  const submitResponse = (label_data) => {
+    trigger(label_data);
+  };
+  const [checkboxValues, setCheckboxValues] = useState(new Array(TEXT_LABEL_FLAGS.length).fill(false));
+  const [sliderValues, setSliderValues] = useState(new Array(TEXT_LABEL_FLAGS.length).fill(1));
+
+  const fetchData = () => {
```
fozziethebeat, line 39: small detail, does this need to be a separate function? It seems to only be called when calling submitResponse.  We could simplify this by putting it all in submitResponse

fozziethebeat (changes requested):
`npm run build` reports a few warnings and errors.  Please make sure those are fixed and no warnings or errors are reported.

website/src/components/FlaggableElement.tsx
BitterKanegul: thank you! I've added the same and ran precommit and npm run build. No issues now!

BitterKanegul: image

fozziethebeat (approved):
Great start! I look forward to seeing this show up in the other tasks

PR #192, opened 2022-12-31
Update all the task pages to fetch a new taks after submission

### LINKED ISSUE #104: Load a new task for users when reaching the done state

fozziethebeat:
Closes #104 

When a user submits an answer to a task and gets to the `done` state we reveal a `Next Task` button that will fetch a new task for them to work on in the same type.

This leverages `useSWR`'s mutate function to refetch tasks.

### DIFF
--- a/website/src/pages/create/assistant_reply.tsx
+++ b/website/src/pages/create/assistant_reply.tsx
@@ -18,9 +18,8 @@ const AssistantReply = () => {
 
   const inputRef = useRef<HTMLTextAreaElement>(null);
 
-  const { isLoading } = useSWRImmutable("/api/new_task/assistant_reply ", fetcher, {
+  const { isLoading, mutate } = useSWRImmutable("/api/new_task/assistant_reply ", fetcher, {
     onSuccess: (data) => {
-      console.log(data);
       setTasks([data]);
     },
   });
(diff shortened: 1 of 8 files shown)


website/src/pages/create/assistant_reply.tsx
```diff
@@ -52,6 +56,7 @@ const AssistantReply = () => {
   }
 
   const task = tasks[0].task;
+  const endTask = tasks[tasks.length - 1];
```
AbdBarho, line 59: what is the advantage of having an array of tasks?

AbdBarho (approved):

website/src/pages/create/assistant_reply.tsx
fozziethebeat: This was originally based on how the demo `text-frontend` handled tasks.  After submitting the prototype was returning a potentially different task.

Now that it's changed to always return `task_done` after submission, I want to simplify all this in another PR that reduces code duplication.

PR #199, opened 2022-12-31
Fix linter warnings

croumegous:
Fixed all warnings in `next build/next lint`

Added a new npm command: `npm run fix` to run format and lint code.
To make review easier I have made 2 commits, the first one adds eslint plugins to automatically handle import sort and unused import.
And a second commit which fixes the remaining warning manually, so this commit is more "opinionated" and needs more attention during the review.
(I have added some `//eslint-disable-next-line` for some part, where it seems to be the best option)

(There is a lot of changes so it might create conflicts for other people working on /website)

### DIFF
--- a/website/.eslintrc.json
+++ b/website/.eslintrc.json
@@ -6,19 +6,9 @@
     "next/core-web-vitals"
   ],
   "rules": {
-    "simple-import-sort/imports": "error",
-    "simple-import-sort/exports": "error",
-    "@typescript-eslint/no-unused-vars": "off",
-    "unused-imports/no-unused-imports": "error",
-    "unused-imports/no-unused-vars": [
-      "warn",
-      {
-        "vars": "all",
-        "varsIgnorePattern": "^_",
-        "args": "after-used",
-        "argsIgnorePattern": "^_"
-      }
-    ]
+    "unused-imports/no-unused-imports": "warn",
+    "simple-import-sort/imports": "warn",
+    "simple-import-sort/exports": "warn"
   },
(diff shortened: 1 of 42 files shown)


website/.eslintrc.json
```diff
@@ -6,6 +6,19 @@
...
+    "unused-imports/no-unused-vars": [
+      "warn",
+      {
+        "vars": "all",
+        "varsIgnorePattern": "^_",
+        "args": "after-used",
+        "argsIgnorePattern": "^_"
+      }
+    ]
+  },
```
AbdBarho, line 22: I know that customizing eslint is always very opinionated... imo there is no need for these rules.

I will have to wait for opinions of other team member...

AbdBarho (commented):
@croumegous thanks for the contribution!

could you please rebase? easier to review.

With the exception of my comment it looks fine.

website/.eslintrc.json
croumegous: I've changed it to keep the less rule as possible to make it work:
```
"unused-imports/no-unused-imports": "warn",  //mandatory to automatically remove unused import  
"simple-import-sort/imports": "warn", //mandatory to automatically sort import  
"simple-import-sort/exports": "warn"  //mandatory to automatically sort export
```

fozziethebeat: I think these three small additions look good.

fozziethebeat (approved):
Things look a lot cleaner now, thank you~

fozziethebeat:
Please fix the merge issues and then I'll re-review

PR #201, opened 2022-12-31
Update Prompting_Guide.md

koustubhavachat:
following two points added 
- Provide accurate and reliable information using credible sources and references as appropriate
- Avoid providing vague or incomplete responses, or giving opinions or personal advice unless specifically requested


yk (approved):
thank you!

PR #207, opened 2022-12-31
data argumentation in docs and notebooks

GravermanDev:
added the changes I did previously but but this time this should be compatible 

I added my essay for data argumentation of essays, it turns any essay into a set of instructions. I also did some much needed work on docs. Added a file about data argumentation so anyone can check what that means and how they can contribute. I also moved prompting guide to docs/ as I think it should be there, makes it a little easier to find for someone looking to get more info.

Added another notebook for essay revision, it generates data helpful for revising essays. I also added markdown files related to the changes explaining the files in more detail


GravermanDev:
@yk I accidently closed the last PR, but this should work now

yk (approved):

Issue #168, opened 2022-12-30
Can't fetch tasks when logging in as a debug users


fozziethebeat:
The website writes a copy of a fetched task to a local database and associates it with the user id fetching the task.

It seems that when using debug credentials, no user entry is created.  I think we can use one of NextAuthJS's hooks to make this work correctly but i'll need to check.

PR #210, opened 2022-12-31
saving user to the dabase when using the debug credentials provider

yk:
Solves #168 

note: a bit hacky, but next-auth makes it quite troublesome to do this in a clean way when JWT is active, and we need JWT because the credentials provider only supports that.


k-nearest-neighbor (approved):
🚢

PR #230, opened 2023-01-01
website: Drag handles for ranking items

### LINKED ISSUE #223: Add drag handles for rankable (sortable) items
Adding drag handles to the dragable items being ranked would make it clearer that they can be dragged.

This was originally suggested by @k-nearest-neighbor in [#158 ](https://github.com/LAION-AI/Open-Assistant/issues/158#issuecomment-1368148998_)
> [...] there should be a drag handle. Something like [RxDragHandleDots2](https://react-icons.github.io/react-icons/search?q=rxdraghandledots2) or [MdDragHandle](https://react-icons.github.io/react-icons/search?q=mddraghandle). On the left or right side is debatable. (both?)

Also the rankable items currently have dot points, that can probably also be removed to avoid confusing the user.

othrayte:
Adds drag handles, to the rank prompts and replies pages of the website.
Closes #223

Removed the dot points that were appearing next to each item. Restrict ranking items to vertical motion to avoid breaking the webpage layout; there is an even better option to restrict to the window but that didn't seem to work on mobile (firefox) were the issue looks the most unprofessional.

Also enabled the keyboard sensor so that the items can be moved without a mouse. To use select one of the drag handles with `TAB` or the mouse to give it focus, press `ENTER` or `SPACE` to start moving the item and use the arrow keys (up or down) to reorder the items, press `ENTER` or `SPACE` again when done, `ESCAPE` cancels the action. There is a full description at https://docs.dndkit.com/guides/accessibility#keyboard-shortcuts which also mentions we might want to customise the screen reader prompt.

image

### DIFF
--- a/website/src/components/Sortable/Sortable.tsx
+++ b/website/src/components/Sortable/Sortable.tsx
@@ -2,6 +2,7 @@ import { Flex } from "@chakra-ui/react";
 import { closestCenter, DndContext, PointerSensor, TouchSensor, useSensor, useSensors } from "@dnd-kit/core";
 import type { DragEndEvent } from "@dnd-kit/core/dist/types/events";
 import { arrayMove, SortableContext, verticalListSortingStrategy } from "@dnd-kit/sortable";
+import { restrictToVerticalAxis } from "@dnd-kit/modifiers";
 import { ReactNode, useEffect, useState } from "react";
 
 import { SortableItem } from "./SortableItem";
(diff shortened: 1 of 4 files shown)


website/src/components/Sortable/Sortable.tsx
```diff
@@ -1,7 +1,21 @@
...
+} from "@dnd-kit/core";
 import type { DragEndEvent } from "@dnd-kit/core/dist/types/events";
-import { arrayMove, SortableContext, verticalListSortingStrategy } from "@dnd-kit/sortable";
+import {
+  arrayMove,
+  SortableContext,
+  sortableKeyboardCoordinates,
+  verticalListSortingStrategy,
+} from "@dnd-kit/sortable";
+import { restrictToVerticalAxis } from "@dnd-kit/modifiers";
```
fozziethebeat, line 18: small nit, could you move this import up above sortable so it's in order?

fozziethebeat (changes requested):

website/src/components/Sortable/Sortable.tsx
othrayte: Sure, just to confirm it should be alphabetically by @scope / unscoped name, then by scoped package name? so:
* a
* @b/a
* @b/e
* c

fozziethebeat: I think with the new eslint change that goes in `@` is always first.  So it'd be

```
@b/a
@b/e
a
c
c-foo
```

fozziethebeat (approved):

PR #235, opened 2023-01-01
Add an error handler to the discord bot

AlexanderHOtt:
Most of these errors only happen with prefix commands, so they can be a bit hard to trigger for testing, but this code is straight forward so there shouldn't be any bugs. I could add logging, but exceptions are already logged to the console.


andreaskoepf (approved):
ok

Issue #243, opened 2023-01-01
Move OasstApiClient & OasstError to oasst_shared, parse Oasst standard exception


andreaskoepf:
!! wait for changes of https://github.com/LAION-AI/Open-Assistant/pull/216 to be integrated

1. `OasstApiClient` which is currently part of the discord bot should be moved to `oasst_shared` to be used by different python frontends (e.g. later text-frontend, testing etc.)
2. The exception class used in the server code `OasstError` should be moved to `oasst_shared` and all locations that currently import `oasst_backend.exceptions` have to be patched, e.g. to `oasst_shared.exceptions`.
3. The OA server supplies an error code + status message when standard OasstErrors happen. Instead of just using `response.raise_for_status()` the OaastApiCient should check (e.g. in `post()`) if the response (for non 2XX http status cases) contains a standard OASST JSON object (with `message` and `error_code` fields) and if so raise a corresponding `OasstError` exception (including the http-status and Oasst-error code). 

Json serialization of OasstError:
```
404 Error: Not Found

Response body:

{
  "message": "Task for task_id=UUID('3fa85f64-5717-4562-b3fc-2c9a3f66afa6') not found",
  "error_code": 2100
}

Response headers:

 content-length: 103
 content-type: application/json
 
```

jack-michaud:
Happy to take a crack at this. Feel free to assign to me

jack-michaud:
Step 1 & 2 are done in #287 and #289. Waiting to get them merged to work on step 3.

yk:
> Step 1 & 2 are done in #287 and #289. Waiting to get them merged to work on step 3.

merged now :)

jack-michaud:
Step 3 is ready to go in #300

PR #244, opened 2023-01-01
ci: add Test API Contract workflow

jack-michaud:
Adds a CI workflow which calls the discord bot tests. I called this "Test API Contract" because the only tests in it are the contract tests. I imagine this workflow can also perform the contract tests for the website too. Once there are Discord bot unittests and OasstApiClient is moved into oasst-shared (#243) we can point to a new test script in this workflow and create a separate Discord bot workflow.

### DIFF
--- a/.github/workflows/test-api-contract.yaml
+++ b/.github/workflows/test-api-contract.yaml
@@ -1,0 +1,33 @@
+name: Test API Contract
+
+on:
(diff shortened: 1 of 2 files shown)


yk (commented):
LGTM, thank you!

yk (approved):

PR #246, opened 2023-01-01
Changed the regex in utils.py and reworked WebGPT loading

agoryuno:
The regex was missing references with >2 ids and 2 ids with a space after the comma. Replaced the for loop for loading the WebGPT dataset up by using dataset.map() instead and flattening the resulting dataset with a list comprehension Should be about 3 times faster this way.


agoryuno:
Screwed up a little (a lottle)

PR #248, opened 2023-01-01
Changed the regex in utils.py...

agoryuno:
Changed the regex in utils.py - the original  was missing references with 2+ ids and spaces after commas. Also changed the way the WebGPT is loaded to use dataset.map() instead of a loop, which should be about 3x faster.


yk:
@theblackcat102 could you have a look?

theblackcat102:
@yk looks good, it fixes reference with space in between.

theblackcat102:
@agoryuno can you remove [these](https://github.com/agoryuno/Open-Assistant/blob/main/model/reward/instructor/trainer.py#L12) unused import? pre-commit is not happy about it

theblackcat102 (approved):
looks good

yk:
Thank you! @agoryuno could you run `pre-commit run --all-files` and then commit & push? to make linters happy

agoryuno:
Strange. I struggled with pre-commit for an hour last night and was sure I won ) I'll do it tomorrow. I've refactored it some more, making it run in Colab, as well as from cli. It was a fairly large refactor so I put it in a separate repo for now. It'd be great if @theblackcat102 could take a look at it: https://github.com/agoryuno/instructor

bitplane:
General feedback about PRs: It's best to create a feature branch for your changes in your own repo, that way we can be sure that the pull request doesn't contain unrelated changes that you've merged into your own default branch.

andreaskoepf:
@agoryuno Could you please resolve the conflict in `model/reward/instructor/trainer.py`?

theblackcat102 (approved):

theblackcat102 (commented):
Just need to resolve the conflict and should be good, need to merge before I can fix the validation bug for rankgen model #397

agoryuno:
Forgot to check if this was merged already before committing changes. Crap!

agoryuno:
I'll reopen a new one with all the changes at once, since I was going to anyway.

theblackcat102:
@agoryuno can you push the PR for your current first? Cause we need to run a working reward-model first and the code for training is changing rapidly, so the later the PR the more conflict there is.

PR #249, opened 2023-01-01
Add context to compose file so it runs in older "docker-compose"

bitplane:
Looks like the reason it was failing in the older `docker-compose` tool but not `docker compose` is because recent versions use the current working dir as a default build context, while older versions need it to be explicitly defined.

This should fix running it on machines or in tools that haven't upgraded to `docker compose` and still call `docker-compose` instead.


yk (approved):
thank you!

Issue #47, opened 2022-12-21
Create API endpoints that return leaderboards


yk:
The frontends will want to display leaderboards. The backend needs to be able to compute these, and should expose API endpoints for the frontends to request their data.

Leaderboards can be made for multiple things: Global leaderboards, time-based (weekly,daily,...), task-based (being the assistant, making prompts, ...), and others. The API endpoints should have the flexibility to allow for this and the backend implementation should start with one or two basic ones (e.g. global and daily).

jasonmhead:
This task perhaps needs expanded and split out with some feature goals and 
I could help design and document the API endpoints.

At a higher level it could be good to use the OpenAPI standard and use tools like https://openapi-generator.tech/

yk:
Yes, I absolutely agree, we should probably split this up. So I guess this issue could mainly track the discussion around what endpoints are even needed.

> At a higher level it could be good to use the OpenAPI standard and use tools like https://openapi-generator.tech/

I believe the framework we use, fastAPI, already generates OpenAPI-compatible endpoints and their specs by default (i.e. https://fastapi.tiangolo.com/tutorial/metadata/#openapi-url)

> Would the "High-Level Protocol Architecture" document provide enough information to design the endpoints?

No, for the leaderboards, that's not enough. The protocol only defines the main interaction between frontend and backend with respect to the tasks to be fulfilled, but says nothing about leaderboards. That will have to be an entirely different set of (generic) endpoints.

danielpatrickhug:
Hi @yk, I was wondering if you had any thoughts on the scheduling aspect. I was thinking of the backend implementation. Potentially, Celery could be used for scheduling and updating the person_stats table for the leaderboard endpoint.

yk:
> Hi @yk, I was wondering if you had any thoughts on the scheduling aspect. I was thinking of the backend implementation. Potentially, Celery could be used for scheduling and updating the person_stats table for the leaderboard endpoint.

I think initially, we can get away with just running a SQL query to compute the stats on the fly. In the future I think it makes sense to use some scheduling system like celery.

yeungadrian:
Hi @yk , I'd like to take this up. Here's some thoughts / question on an initial implementation:
- New Router: Called leaderboards (or could put it under the current stats)
- Initial Proof of Concept endpoints: (one endpoint per stat), copying the phrasing I'm seeing in the UI
    - GET /api/v1/leaderboards/user/{timespan}
[...]

PR #250, opened 2023-01-01
#47- Create API endpoints that return leaderboards

yeungadrian:
#47 
 Leaderboard endpoints proof of concept

- New leaderboards router
- Two endpoints /api/v1/leaderboards/create/assistant & /api/v1/leaderboards/create/prompter
- Function in prompt_repository to calculate user scores. ( Count of Messages created (not deleted) per user )

image

There's a couple of ways to structure the endpoints. Using this structure we could end up with:
 /api/v1/leaderboards/create/prompter/{timespan}
... create/combined, create/assistant
 /api/v1/leaderboards/evaluate/prompter/{timespan}
... evaluate/user, evaluate/combined ...


yk (changes requested):
Very nice, very clean thank you! I think this is a good first step.
Could you rename the route prefix to be `/experimental/leaderboards`? just so we have nobody fully depend on that yet

yeungadrian:
Thanks @yk . That makes sense

andreaskoepf:
Thanks a lot for the first stats-based leaderboard. This is definitely a metric we should include but alone it encourages the user to flood the db with texts. I will try to write down current ideas that are floating araound for the user scoring.

Regarding online-calculation: Could you pehaps create a litte benchmark test that measures how long the message-count leaderboard query takes for of a fake-DB with 500k messages (each 500 charaters long) and 1k users thtat randomly are the authors of the fake messages? That would be helpful to get a first idea of the performance. It would also be a good test-script that we could use for other scenarios.

andreaskoepf (approved):
we can start with this.

yk (approved):

PR #252, opened 2023-01-02
Remove utf8 coding pragma from pre-commit and python files

bitplane:
We don't need the pre-commit pragma in Python 3, and since Python 2 has been officially dead for 3 years we're a Python 3 codebase.

It looks like a big change because of the number of files impacted, but it's low risk and there shouldn't be any conflicts.


yk (approved):
congrats, you just became our top-committer ;)

PR #265, opened 2023-01-02
Stop users from starting more than one task at a time

AlexanderHOtt:
The bot now keeps a `set` of user ID's for everyone currently doing a task. This gets removed after a user times out or doesn't want to do another task.

### DIFF
--- a/discord-bot/bot/bot.py
+++ b/discord-bot/bot/bot.py
@@ -35,6 +35,9 @@ async def on_starting(event: hikari.StartingEvent):
 
     bot.d.oasst_api = OasstApiClient(settings.oasst_api_url, settings.oasst_api_key)
 
+    # A set of user id's that are currently doing work.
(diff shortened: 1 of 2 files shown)


yk:
I don't have a good overview over what the code looks like, but could we store a map user -> task, so that at least internally we have an idea which task the user is working on? we could then log this internally when we yell at the user, and when there's a problem and a user says "hey it tells me I already have a task, but I don't", we can go and debug.

andreaskoepf (approved):
ok, looks good.

AlexanderHOtt:
> I don't have a good overview over what the code looks like, but could we store a map user -> task, so that at least internally we have an idea which task the user is working on? we could then log this internally when we yell at the user, and when there's a problem and a user says "hey it tells me I already have a task, but I don't", we can go and debug.

We could use a weakref map from user id to task. Should we give the user an option to cancel their old task?

PR #269, opened 2023-01-02
Add performance tracking to the bot.

AlexanderHOtt:
There is currently no way to visualize the data, but a script or command can be added if necessary.

### DIFF
--- a/discord-bot/bot/api_client.py
+++ b/discord-bot/bot/api_client.py
@@ -40,6 +40,7 @@ class OasstApiClient:
         self.session = aiohttp.ClientSession()
         self.backend_url = backend_url
         self.api_key = api_key
+        self._calls = 0
 
         self.task_models_map: dict[TaskType, Type[protocol_schema.Task]] = {
             TaskType.summarize_story: protocol_schema.SummarizeStoryTask,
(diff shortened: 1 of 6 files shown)


yk (approved):
LGTM
Two independent questions:
- why is `_post` now private?
- is it at least conceivable that we use some kind of db-framework in the future (e.g. sqlmodel+alembic, etc.) instead of raw sql, or is there a technical reason against this?

discord-bot/bot/api_client.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, backend_url: str, api_key: str):
         self.session = aiohttp.ClientSession()
         self.backend_url = backend_url
         self.api_key = api_key
+        self._calls = 0
```
andreaskoepf, line 43: What is the idea of this call-counter, when could ist be useful?

andreaskoepf (changes requested):
This PR adds perf DB logging of the bot process. This is a feature without a clear requirement, nice to have. regarding api-client: I rather would not have the counter in the api-client .. also I do not see where this metric would really help. It is planned to move the api-client into oasst-shared library. Logging of the process stats could be merged if you consider it important for the bot-operation.

AlexanderHOtt:
> LGTM Two independent questions:
> 
> * why is `_post` now private?
> * is it at least conceivable that we use some kind of db-framework in the future (e.g. sqlmodel+alembic, etc.) instead of raw sql, or is there a technical reason against this?

I see it as users should only interact with the direct methods like `fetch_task`, `ack_task`, and `post_interaction`. This is not meant to be used as a regular http client.

AlexanderHOtt:
Logging bot perf will become important when I start tuning the cache settings. Hikari's cache is very powerful, but we need to be careful because some categories (like presence) take up a lot of memory.

I guess tracking API calls isn't too useful right now. If you really don't want it, I can remove it.

andreaskoepf:
> Logging bot perf will become important when I start tuning the cache settings. Hikari's cache is very powerful, but we need to be careful because some categories (like presence) take up a lot of memory.

OK, sounds reasonable.
 
> I guess tracking API calls isn't too useful right now. If you really don't want it, I can remove it.

OasstApiClient has been moved ot oaast-shared. I suggest simply using it as it is right now. I hope exception parsing will also soon be added so that cllients see `OasstError´ exceptions with message and internal code.

PR #286, opened 2023-01-02
Clean up data augmentation doc

olliestanley:
Just resolves some typos, primarily the use of "argumentation" instead of "augmentation" which may have led to confusion


yk (approved):
haha I'm just seeing this now 🙃 thank you!

PR #288, opened 2023-01-02
added data schemas md

yk:


andreaskoepf (approved):
thanks for schemas. Igor created similar structures for the backend API .. maybe you could refactor that to also become ConversationTreeNode & ConsersationTree .. e.g. adding user and date inso as metadata and not direct members.

PR #289, opened 2023-01-02
refactor: move OasstError into oasst_shared.exceptions

jack-michaud:
Fixes step two of #243

- refactor: move OasstError into oasst_shared.exceptions
- fix: update OasstError import
- fix: linter errors


andreaskoepf (approved):
Sweet and simple! thx

PR #292, opened 2023-01-02
Add missing requirement to discord-bot

onegunsamurai:
When starting container with nonempty envs , python was missing `pydantic[dotenv]`


andreaskoepf (approved):
good catch

Issue #276, opened 2023-01-02
Account page (website) says "null"


othrayte:
If you sign in to the website with either email or the debug login (other types not tested) and go the account page (/account, or click on the user menu top right) the first content on the page is 'null'.
![image](https://user-images.githubusercontent.com/241372/210233172-a8aa1aaf-17bf-41be-8d78-f3619786ca17.png)

If you login with debug login the location where the email address would appear is also blank so you only see 'null' \n "Edit Username". We should probably show the debug username there like we do in the account menu top right.

The page probably also needs some additional labels or something on it to make it clearer what the content is.

Jac-Zac:
This merge doesn't complitly fix the issue for me since when I try to edit the username it never updates. And also it gives me an error. I have made some changes in my pull request to fix this

PR #294, opened 2023-01-02
#276 - Fixed null username

Klotske:
### Fixes issue #276 
Used code from edit page to show current username or the abscence of it.
Removed the update handler because it was duplicating Account "Edit" page functionallity.

**Before:**
Before
**After:**
After

### DIFF
--- a/website/src/pages/account/index.tsx
+++ b/website/src/pages/account/index.tsx
@@ -2,7 +2,7 @@ import { Button } from "@chakra-ui/react";
 import Head from "next/head";
 import Link from "next/link";
 import { useSession } from "next-auth/react";
-import React, { useState } from "react";
+import React from "react";
 
 export default function Account() {
   const { data: session } = useSession();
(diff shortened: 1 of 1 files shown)


website/src/pages/account/index.tsx
```diff
@@ -2,24 +2,10 @@ import { Button } from "@chakra-ui/react";
...
-import React, { useState } from "react";
+import React from "react";
 
 export default function Account() {
   const { data: session } = useSession();
-  const [username, setUsername] = useState("null");
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const handleUpdate = async () => {
-    const response = await fetch("../api/update", {
```
fozziethebeat, line 13: I double checked and `/api/update` doesn't exist.  It's called `/api/username`.  I'll fix this in a follow up PR where the user explicitly clicks a button to update their username.

fozziethebeat (approved):

Issue #298, opened 2023-01-02
invalid json response


croumegous:
Oops seems like my last PR: https://github.com/LAION-AI/Open-Assistant/pull/295 broke frontend.

When we try to get a new task we get a 500 error (for every type of task).
logs:
```shell
FetchError: invalid json response body at http://backend:8080/api/v1/tasks/2fb434b1-41bc-4c6a-83a5-0d9e3d69adac/ack reason: Unexpected end of JSON input
    at /app/node_modules/next/dist/compiled/node-fetch/index.js:1:51220
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async handler (/app/.next/server/pages/api/new_task/[task_type].js:105:5)
    at async Object.apiResolver (/app/node_modules/next/dist/server/api-utils/node.js:363:9)
    at async NextNodeServer.runApi (/app/node_modules/next/dist/server/next-server.js:474:9)
    at async Object.fn (/app/node_modules/next/dist/server/next-server.js:736:37)
    at async Router.execute (/app/node_modules/next/dist/server/router.js:252:36)
    at async NextNodeServer.run (/app/node_modules/next/dist/server/base-server.js:383:29)
    at async NextNodeServer.handleRequest (/app/node_modules/next/dist/server/base-server.js:321:20)
    at async Server.<anonymous> (/app/server.js:19:5) {
  type: 'invalid-json'
}
```
This is because we now send a HTTP 204 without content on /ack therefore frontend is not happy and create invalid-json error.

Easy fix is to rollback to empty 200 response.
But might be better to fix frontend. (however don't know if the 204 response affect discord bot)

PR #299, opened 2023-01-02
fix(frontend): fix invalid json error when HTTP response is 204

croumegous:
Fix https://github.com/LAION-AI/Open-Assistant/issues/298


fozziethebeat (approved):
Good catch on keeping things in sync!

PR #300, opened 2023-01-03
feat: handle OasstError in OasstApiClient

jack-michaud:
Fixes step 3 of #243

- feat: add OasstErrorResponse to protocols
- feat: handle OasstError in OasstApiClient
- test: add test for handling OasstError
- test: add test for unhandled api error

Embarrassingly, failed tests were not surfacing in CI, so that's fixed in 789593b. This issue effectively hid the fact that the tests were failing to run because of a missing dependency in oasst-shared, and that's fixed in fbcb0a0.


yk (approved):
Looks solid, thank you!

Issue #78, opened 2022-12-27
Train a reward model based on RankGen


andreaskoepf:
Add a reward-head (linear projection of the model's embedding to a scalar value) to [martiansideofthemoon/rankgen](https://github.com/martiansideofthemoon/rankgen) & [RankGen paper](https://arxiv.org/abs/2205.09726)) and train it on human-feedback (good-bad example pairs) of the [openai/summarize-from-feedback](https://github.com/openai/summarize-from-feedback) dataset (see the [Learning to summarize from human feedback](https://arxiv.org/abs/2009.01325) paper for details about the objective).

- place your training code in a new `model/reward/rankgen` folder
- please use wandb for experiment tracking, measure at least loss + accuracy
- try to avoid modifying the original model, if possible aggregate the existing model (i.e. add the existing model as a member of the new model class)
- compare with results from #77

Background:
We want to implement the RLHF stack for Open-Assistant in parallel to our data collection effort. As a temporary fill-in we use existing RLHF datasets like [OpenAI's learning to summarize](https://openai.com/blog/learning-to-summarize-with-human-feedback/) for model development. Instructor was proposed as a promising base-model candidate for a reward model. 

You could use [bits of reward model training code](https://github.com/andreaskoepf/laion_sum_data/blob/main/reward_model/train.py) that I trained a couple of weeks ago which contains data loading code for the summarize-from-feedback data as inspirations. If you like, you can of course use of a framework like pytorch_lightning.

bth5032:
I'd be happy to take a look at this. 

If I understand correctly, you'd like me to

1. Start with rankgen pretrained model
2. Use the dot product method between posts explained in rankgen paper to score the summaries from the SFF dataset 
3. Treat the dot product score as rewards for the RM loss from SFF

A few further questions:
1. Do we want to then implement the policy optimization step as well?
2. What is the baseline generative model we are using, e.g. to do comparisons with #77 
3. What kind of compute resources do you expect this to require?

andreaskoepf:
> 2\. Use the dot product method between posts explained in rankgen paper to score the summaries from the SFF dataset

I am not sure if the method described in the rankgen paper can be directly applied here (I only skimmed the paper)? If you know more, let me know... My first thoughts were to take the rankgen embedding and project it to a scalar reward and then finetune with cross-entropy loss.
 
> A few further questions:
>     1. Do we want to then implement the policy optimization step as well?

That will be the topic of a separate issue. 

[...]

PR #313, opened 2023-01-03
Bth5032/78 blackcat trainer

bth5032:
#78 

As per discussion with @theblackcat102 I built the rankgen trainer on top of their framework ([wandb](https://wandb.ai/bobakhashemi/reward-model/runs/2sgitozt?workspace=user-bobakhashemi)). The model seems to be training now in fp32. Apparently t5 has some issue with fp16. Blackcat suggested scaling the weights might help as in [1](https://github.com/LAION-AI/Open-Assistant/blob/main/model/reward/instructor/trainer.py). This could be a good next step if we want to move with this model. Until then, I think this code is worth comitting since it shows how to add new models which are not `AutoModelForSequenceClassification`

### DIFF
--- a/model/reward/instructor/configs/rankgen-t5-base-fp16.yml
+++ b/model/reward/instructor/configs/rankgen-t5-base-fp16.yml
@@ -1,11 +1,8 @@
 model_name: kalpeshk2011/rankgen-t5-base-all
-# model_name: kalpeshk2011/rankgen-t5-xl-all
-# model_name: kalpeshk2011/rankgen-t5-xl-pg19
-# model_name: kalpeshk2011/rankgen-t5-large-all
 tokenizer_name: google/t5-v1_1-base
 learning_rate: 6e-6
 gradient_checkpointing: false
-fp16: false
+fp16: true
 gradient_accumulation_steps: 16
 per_device_train_batch_size: 2
 warmup_steps: 600
--- a/model/reward/instructor/configs/rankgen-t5-base.yml
+++ b/model/reward/instructor/configs/rankgen-t5-base.yml
@@ -2,8 +2,9 @@ model_name: kalpeshk2011/rankgen-t5-base-all
 tokenizer_name: google/t5-v1_1-base
 learning_rate: 6e-6
 gradient_checkpointing: false
+fp16: false
 gradient_accumulation_steps: 16
-per_device_train_batch_size: 3
+per_device_train_batch_size: 2
 warmup_steps: 600
 freeze_layer: 20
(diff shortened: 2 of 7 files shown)


theblackcat102:
@bth5032 your wandb loss looks good, but I wonder why is the accuracy validation missing?

Not trying to nickpick anyway, but just curious why not assign the default value for tokenizer_name is same as model_name at the [argument_parser](https://github.com/bth5032/Open-Assistant/blob/bth5032/78-blackcat-trainer/model/reward/instructor/utils.py#L94)  then we can remove this if else at the trainer [here](https://github.com/bth5032/Open-Assistant/blob/bth5032/78-blackcat-trainer/model/reward/instructor/trainer.py#L200)

But overall the code looks good

theblackcat102 (approved):
looks good

sanagno (approved):
Thanks a lot!

yk:
@bth5032 thank you! could you run `pre-commit run --all-files` to make linters happy?

bth5032:
> @bth5032 thank you! could you run `pre-commit run --all-files` to make linters happy?

Thanks! 

> Not trying to nickpick anyway, but just curious why not assign the default value for tokenizer_name is same as model_name at the [argument_parser](https://github.com/bth5032/Open-Assistant/blob/bth5032/78-blackcat-trainer/model/reward/instructor/utils.py#L94) then we can remove this if else at the trainer [here](https://github.com/bth5032/Open-Assistant/blob/bth5032/78-blackcat-trainer/model/reward/instructor/trainer.py#L200)

Ah, I normally use hydra/omegaconf for this, didn't realize you had that utility function, will update and push another. I cleaned that up

> @bth5032 your wandb loss looks good, but I wonder why is the accuracy validation missing?

I've been trying to figure that out myself. The `compute_metrics` function never runs for me... I found [this thread](https://discuss.huggingface.co/t/trainer-never-invokes-compute-metrics/11440/6) and tried setting up `evaluation_strategy=IntervalStrategy.STEPS` and a few other things. But ultimately I verified the condition they listed [here](https://github.com/huggingface/transformers/blob/8afaaa26f5754948f4ddf8f31d70d0293488a897/src/transformers/trainer_callback.py#L419-L423) is true. However I still don't ever see `compute_metrics` evaluating...

I'll keep looking into it and submit a PR if I figure it out.


@theblackcat102 @sanagno should be ready for merge.

theblackcat102 (commented):
looks good

theblackcat102 (commented):
models.py is fine

PR #321, opened 2023-01-03
257 backend on aws

fozziethebeat:
This creates a basic Copilot configuration that runs the backend on AWS using a simple Postgres database.


backend/oasst_backend/config.py
```diff
@@ -44,4 +44,5 @@ def assemble_cors_origins(cls, v: Union[str, List[str]]) -> Union[List[str], str
         raise ValueError(v)
 
 
+print("cats")
```
yk, line 47: 🐈 🐈 🐈

yk (approved):

backend/oasst_backend/config.py
fozziethebeat: that was to force a re-deployment lol.  This is deleted now.

Issue #305, opened 2023-01-03
Where to contribute datasets


iurimatias:
As part of phase 1, demonstration data is collected. I would like to contribute such data:
- should it go into this repo or somewhere else?
- should it be the data itself or only the scripts that generate such data ?

bitplane:
I think just the scripts that generate the data, with the data itself being collected and put in S3 buckets or on HuggingFace or whatever.

We really need an example data collection package within the main repo, with an example script that people can use as a starting point. This could either be as module or a sub-package. The danger of not having this structure is we'll end up with tons of data collection scripts that have to be run manually and have a range of dependencies and manual setup steps, that output data to different places.

I'll add a ticket for that if there isn't one already.

bitplane:
Added #331

bitplane:
And closed it in favour of #165 :)

lewtun:
Working on addressing this in https://github.com/LAION-AI/Open-Assistant/pull/324

marianna13:
Hi everyone! I already have a working pipeline for YouTube subtitles extraction and I already got subtitles for 1.5M+ videos. Regarding copyright, I don't think that subtitles have strict copyright infringements (see [here](https://opensource.stackexchange.com/a/10721)).

Issue #165, opened 2022-12-30
Set up an initial framework for data collection, storage, cleaning, and accessing


yk:
Many people are collecting data, for example scraping websites, etc. We need a framework (guidelines, how-to's, documented processes, etc.) on how that data does not get lost and gets made into something useful. Ideally, the people who collect the data provide the code to collect, clean, and process the data as a PR to the repo (can be some new subfolder like `datasets/`). I.e. they do not provide the data correctly, but the code, and along that, they provide clear instructions on how to use it. (Ideally the code comes in CLI-like scripts, can easily be built with e.g. `typer` or `click`). Then, we need a storage location (set up a LAION s3 bucket) and a method of running these scripts to actually collect, store, and process the data, such that it is available for downstream applications (ML training).

yk:
will assign to @lewtun as soon as he accepts invite ;)

lewtun:
A few initial thoughts on how we could standardise the data collection / data processing efforts:

#### 1. Create a dedicated [GitHub project board](https://github.com/LAION-AI/Open-Assistant/projects?query=is%3Aopen) to track which datasets to add / WIP / done

Using Discord to coordinate this work it pretty hard, so having an overview would help. We could use a similar format to the 🤗 Datasets [issues](https://github.com/huggingface/datasets/issues/5371), which log basic info needed for anyone to work on adding the dataset.

@yk I don't have permissions to create a new project board - if you agree, could you create one for this?


#### 2. Use the 🤗 Datasets library for loading / processing

* The dataset loading scripts will be standardised, i.e. easy to reproduce with a one-liner `load_dataset(...)` call
* Has [S3 support](https://huggingface.co/docs/datasets/v2.8.0/en/filesystems#cloud-storage)
* Allows people to easily push curated datasets to the Hugging Face Hub at some later date with the `push_to_hub()` method ([docs](https://huggingface.co/docs/datasets/v2.8.0/en/package_reference/main_classes#datasets.Dataset.push_to_hub))
* Has plenty of high-performance features baked in (caching, multiprocessing) and plays nice with 🤗 Tokenizers (which I assume we'll be using)

#### 3. Focus first on dataset hosting / loading

In principle, a cleaned or processed dataset in 🤗 Datasets is just another dataset, so my suggestion is to first focus on having a good workflow for adding raw datasets and then extending that for whatever downstream analysis is needed.

lewtun:
[...]

PR #324, opened 2023-01-03
Add dataset creation guide

lewtun:
This PR adds a small guide on how to upload datasets to the OpenAssistant org on the Hugging Face Hub. 

To simplify the process, I've created a Space ([link](https://huggingface.co/spaces/OpenAssistant/dataset-generator)) that creates a template loading script that people can edit. Note that although there are easier ways to add data to the Hub (e.g. by dumping JSON or CSV files), this approach guarantees the loading scripts can also run on S3 if needed later on. 

For now, I haven't added any instructions about the column formats for specific tasks like instruction-fine tuning, but that can be done in a second step once there's some agreement.

Closes https://github.com/LAION-AI/Open-Assistant/issues/305 https://github.com/LAION-AI/Open-Assistant/issues/165


yk:
Looks really nice, thank you!

I see two potential workflows:
a) people do this largely by themselves
b) people submit code to scrape/parse/clean/upload and "we" run them

a) would certainly be easier, but I tend to favor b) because we could make sure things are uniform and work as intended when people submit their code via PRs. I don't think a hybrid "upload-yourself-but-also-commit-code" would work well, as we'll have no guarantee the code to upload actually matches the one in the repo (and since people tend to do lots of fixes, it probably will be very different quickly).

What do you think? If b) sounds good, maybe we'll adjust the guide a bit to have one section for "data admins" (i.e. the section that's already there) and one section for "I have a dataset that I want to submit" that details the steps to submit the DS upload code.

also, `pre-commit run --all-files`

lewtun:
OK, I like option (b) too. I think a clean way to do this is to have a folder structure like:

```
datasets
├── README.md
└── dataset_1
    ├── README.md
    ├── load.py
    ├── preprocess.py
    ├── push_to_hub.py
    ├── requirements.txt
    └── scrape.py
```

Here each `dataset_i` folder has the same set of scripts and we can provide a simple template for people to copy-paste. I'll amend the PR in this direction :)

lewtun:
> What about we provide some kind of pre-commit that will make all neccessary checks, that users will need to run on their side before creating the PR with a dataset. Basically just like the one we have rn In Open Assitant repo.

We can certainly do this, although we probably won't be able to check the heavy stuff like scraping / downloading large corpora

yk:
sounds great, thank you for working this out 👍

openassistant/utils/configs.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+from dataclasses import dataclass
+
+import datasets
+
+
+@dataclass
+class OpenAssistantConfig(datasets.BuilderConfig):
```
lewtun, line 8: This class ensures all Hub datasets have the same config structure

openassistant/utils/schemas/language_modeling.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+"""
+Language Modeling Schema
```
lewtun, line 3: Based on The Pile which just has two fields - happy to extend if needed

openassistant/templates/dataset_card.md
```diff
@@ -0,0 +1,28 @@
+---
+license: mit
+tags:
```
lewtun, line 3: Any other tags we should add here for discoverability / filtering?

openassistant/templates/dataset_card.md
```diff
@@ -0,0 +1,28 @@
+---
+license: mit
```
lewtun, line 2: Just pick a permissive license at random

openassistant/templates/hub.py
```diff
@@ -0,0 +1,22 @@
...
+    """BuilderConfig for OpenAssistant datasets."""
+
+    name: str = None
+    version: datasets.Version = None
+    description: str = None
+    schema: str = None
+    subset_id: str = None
+
+
+lm_features = datasets.Features(
```
lewtun, line 17: These "features" are the schema associated with each dataset. I need to check out `data_schemas.md` to integrate them here as well.

For language modelling, I just picked the schema from The Pile (happy to change it)

lewtun:
Alright, I revamped this PR significantly to allow us to have some standardisation going forward:

* Created an `openassistant` Python "lib" where we can store all the dataset loading scripts + templates
* Expanded the guide for people to commit just the loading scripts

I could take a stab at adding a simple dataset as an example if someone can point me to one :)

yk (approved):
Amazing, thank you so much. I think all your choices make total sense.

I'll merge this bc lots of people are looking for it, we can adjust as we go :)

PR #325, opened 2023-01-03
readme typo; Use `docker-compose` instead of `docker compose`

tijszwinkels:


yk:
as far as I know, `docker-compose` is the old, external binary, and has been included into the main docker ecosystem, thus the new way is `docker compose`

yk:
https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose

yk:
I'll close this, as it's better to use the up to date version `docker compose`. If something doesn't work using that (update docker to newest version first), let us know

PR #327, opened 2023-01-03
Update .devcontainer to use the universal devcontainer

kostiak:
### Suggestion to use the universal devcontainer for development on github codepaces

Hey, I was experimenting with github codespaces and a way to run `docker compose up --build` in it. Using the universal devcontainer I was able to get the following workflows working:

**Tested worflows:**

1. `docker compose up --build`
2. the website workflow (detailed inside website/README.md) including:
2.1 `docker compose up frontend-dev --build --attach-dependencies`
2.2 `npm install`, `npx prisma db push`, `npm run dev` inside `website/`

**Known issue:**

When signing in or out, after the signin/signout you are redicrected to localhost:3000 instead of back to the github codespace forwarded url. A workaround I've been using is to go back to the forwarded url. So for example if I log in - I get forwarded to localhost:3000, then I go back to the forwarded url and refresh the page and I'm already logged in and can use the website.

Let me know what you think.


yk:
thanks a lot for looking into this. do you have a bit of insight into what the current dev container is doing? and why the default one would be an improvement?

kostiak:
@yk The current container tries to spin up the `frontend-dev` service. The default devcontainer env gives you access to things like `nvm`, `npm`, `node`, `python`, `git`, `docker` etc. out of the box - basically a standard dev env and once it's up, you are able to spin up everything using `docker compose up --build`  and other workflows from the already existing `README.md`s just like you would do on your local env.

That way, the workflow of using codespaces and local envs is more similar. Using this .devcontainer setup, you get a remote dev environment that's supposed to mimick the way you would work locally. The added advantage is because it doesn't need to build anything before spinning up, it's faster to start. 

One possible disadvantage is that you need to wait for the docker build to happen inside the already running codespace, but again, that's similar to how you would work locally.

yk:
I like this. @andrewm4894 would you agree to this? The default container would indeed be more universal. We could amend the main README to have people use the devcontainer if they want to see the end-to-end demo, along with the steps to run the docker build.

Maybe another question is whether we'd at least already run some commands in the dev container to get people going faster. Examples:
- pip install
- npm install
- docker build
- ?

andrewm4894:
also should maybe install `pre-commit` into it too etc.

andrewm4894:
we could just layer in whatever pip stuff we need like pre-commit etc into a .Dockerfile (which builds on `mcr.microsoft.com/vscode/devcontainers/universal` image) within .devcontainer and have it use that. 

actually i see pre-commit as an available feature so could also just do it that way: https://containers.dev/features

ghcr.io/devcontainers-contrib/features/pre-commit:2
https://github.com/devcontainers-contrib/features/tree/main/src/pre-commit

.devcontainer/devcontainer.json
```diff
@@ -1,10 +1,4 @@
...
-  "service": "frontend-dev",
-  "dockerComposeFile": "../docker-compose.yaml",
-  "forwardPorts": [3000],
-  "customizations": {
-    "vscode": {
-      "extensions": ["GitHub.copilot"]
-    }
-  }
+  "name": "Open-Assistant",
+  "image": "mcr.microsoft.com/vscode/devcontainers/universal"
```
andrewm4894, line 3:
```suggestion
  "image": "mcr.microsoft.com/vscode/devcontainers/universal",
  "features": {
    "ghcr.io/devcontainers-contrib/features/pre-commit:2": {"version": "2.0.1"}
  }
```

andrewm4894: we could i think add `pre-commit` like this.

andrewm4894 (approved):

kostiak:
Ok, got it. Let me take another look at this and I'll report what I find.

andrewm4894:
pre-commit via features seems to work:

image

.devcontainer/devcontainer.json
```diff
@@ -1,10 +1,7 @@
...
-  "dockerComposeFile": "../docker-compose.yaml",
-  "forwardPorts": [3000],
-  "customizations": {
-    "vscode": {
-      "extensions": ["GitHub.copilot"]
-    }
+  "name": "Open-Assistant",
+  "image": "mcr.microsoft.com/vscode/devcontainers/universal",
+  "features": {
+    "ghcr.io/devcontainers-contrib/features/pre-commit:2": {"version": "2.0.1"}
```
andrewm4894, line 5:
```suggestion
    "ghcr.io/devcontainers-contrib/features/pre-commit:2": {
      "version": "2.0.1"
    }
```

my bad, might need be something like this to make pre-commit happy

andrewm4894: can see it complaining here: https://github.com/LAION-AI/Open-Assistant/actions/runs/3830383352/jobs/6518170592

kostiak: @andrewm4894 
Also when I tried to run it, it gives me the following error:
```[ERROR] The hook `black` requires pre-commit version 2.9.2 but version 2.0.1 is installed.  Perhaps run `pip install --upgrade pre-commit`.```

I'll try to change the version and report back with that.

andrewm4894:
actually looks like maybe we need a higher version of pre-commit for black and so might need to go the dockerfile layer approach instead perhaps.

andrewm4894:
@kostiak hey - "latest" seems to work for pre-commit. 

I'm playing about in here: https://github.com/andrewm4894/Open-Assistant/tree/devcontainer-improvements/.devcontainer

eg using latest seems to let me run pre-commit in the codespace then: 
image

kostiak:
@andrewm4894 Are you able to succesfully run `pre-commit run --all-files` after that?

andrewm4894:
> @andrewm4894 Are you able to succesfully run `pre-commit run --all-files` after that?

yep - apart from this one i broke by adding a `post_create_command.sh` script that we can use to script any custom steps like npm install etc.

seems from docs that `postCreateCommand` maybe better way then having to build or layer a custom image. So was going to see about adding such a script to see if can work. 

image

andrewm4894:
This seems to work as an option to just have one post create script and can put whatever we need/want in there: 

https://github.com/andrewm4894/Open-Assistant/tree/devcontainer-improvements/.devcontainer

image

andrewm4894:
made this draft PR to make it clear stuff i'm playing around with. https://github.com/LAION-AI/Open-Assistant/pull/333

andrewm4894:
@kostiak hey - i think my PR (https://github.com/LAION-AI/Open-Assistant/pull/333) maybe should be good to merge - main difference is using `post_create_command.sh` as might be cleaner and easier to maintain just the one script we can add stuff to. 

I would propose that once that is merged you could add the npm website stuff into `post_create_command.sh` (and anything else we might need/want) in a follow on PR. 

Apologies for breaking off into a separate PR in my fork - just had some time today to work on it so went for it.

kostiak:
Added onCreate/postCreate commands for npm, docker compose build and pre-commit. Now the image gets built before the docker is open and you can simply run `docker compose up` and the whole docker spins up pretty fast now.

Also added an attribute to port 3000 to open a browser when the port opens. It takes a couple of minute from when the port opens to when the whole env is fully loaded and the website interface is displayed.

@andrewm4894 No problem, let me know how you want to do this, I can just close this PR and create a new one after yours is in.

andrewm4894:
@kostiak cool, I'll try see if we can merge my PR and then you can do a follow on fresh one to build on it. 

Closing this one.

andrewm4894:
@kostiak merged: https://github.com/LAION-AI/Open-Assistant/pull/333

kostiak:
@andrewm4894 Thanks, I just spun up an instance from origin/main and it seems to work great, a nice starting point for someone who doesn't already have a dev env setup and ready to go.

I'll see what I can add to it and will try to get a PR up when I can. So far from my experiments it looks like including docker compose build in the pre-load might be taking more time than saving. I'll try to look into it more when I can.

PR #312, opened 2023-01-03
Added Dashboard and Dark Mode

rsandb:
Splits out the landing page for a user with a new design.

Closes #156

PR #328, opened 2023-01-03
Make menu icon visible in dark mode

chs20:
Set the stroke of the menu icon based on the current color mode.

Before:
before

After:
after


AbdBarho (commented):
thanks @chs20 for contributing!

We have a big PR containing many changes to different component including dark mode changes.
I am not sure if this is also addressed there: https://github.com/LAION-AI/Open-Assistant/pull/312

This is a small change so I think we can merge regardless, what do others think?

fozziethebeat (approved):
I approve. The dashboard change may take a while to get ready so we can add this in.

Issue #275, opened 2023-01-02
Implement create_initial_prompt for website


othrayte:
There is a way to rank initial prompts but we seem to be missing a page to create initial prompts, which I think is a thing we can do in the discord bot.

AbdBarho:
Do we have the backend for this?

yk:
At least in the protocol, we have a task type called `rank_initial_prompts` or something

othrayte:
We also already have the website front end for ranking the initial prompts (which functions and I assume records the results in the backend) just seemingly no way to create them, that is what originally prompted me to create this issue.

fozziethebeat:
Thank you for pointing this out.

In the backend, I believe the task type is called `initial_prompt`.  It doesn't really return anything other than a task id.    I'll assign it to @othrayte :)

Issue #274, opened 2023-01-02
Automate Cypress (UI component & E2E) tests


othrayte:
To make the most use of our UI and E2E tests for the website we should automate their use in the CI pipeline, otherwise they won't stay working and aren't flagging issues.

As mentioned in our [README for Cypress](https://github.com/LAION-AI/Open-Assistant/blob/main/website/cypress/README.md#component-and-e2e-testing-with-cypress) we should be able to automate the running of Cypress based tests as a CI step. That page mentions we should be able to do it for free, and after some very quick initial searches it looks like it might be free so long as we don't use Cypress Cloud, and [publish screenshots etc as artifacts](https://github.com/cypress-io/github-action#artifacts). Initial searching also found [the official GitHub action](https://docs.cypress.io/guides/continuous-integration/github-actions#Basic-Setup).

Issue #159, opened 2022-12-29
Re-evaluate the design and descriptions of tasks for a better understanding


AbdBarho:
From comment https://github.com/LAION-AI/Open-Assistant/pull/127#pullrequestreview-1232736041

Consider adjusting the tasks' description and possibly design

For example, the `rank_user_replies` task currently shows different replies, but it does not show what is being replied to? critical context is missing.

Go through all tasks and evaluate whether a change is required and what:
* Assistant reply & User reply
* Rank Assistant replies & Rank User replies
* Rank initial prompts
* Summarize story
* Rank summary

yk:
yes you're right, there should be the entire thread up to that point so the rater knows the context of the replies they're asked to rank!

AbdBarho:
## Assistant reply & User reply


![210204023-25594e51-8c55-4af9-be6a-860e75a0b16f](https://user-images.githubusercontent.com/24505302/210204471-a6c1fdbc-64fe-4e71-916d-efd5ef50fa6b.jpg)


Currently the "speaker" is color coded, maybe it makes sense to add some text to show the speaker? Mockup (design could be better of course):

![Screenshot 2023-01-02 084310](https://user-images.githubusercontent.com/24505302/210204969-8708d9da-eb05-4264-b7a7-c8611aa8c329.jpg)

Maybe using [tags](https://chakra-ui.com/docs/components/tag#with-custom-element) and icons?

yk:
only color-coding is not really suitable, because it needs a legend and potentially people with visual impairments could have trouble discriminating.

"User" is called "Prompter" to avoid confusion internally, but towards the actual user, idk what makes more sense.

I like the approach with avatars the most, maybe even avatars on different sides like many chat-apps

othrayte:
I think avatars is the clearest and most representative of what the conversations actually look like, thus making it the clearest for a new user to understand.

For bonus points we could AI generate some avatars to represent the prompter (kindof a joke, but maybe a lower effort set of N pregenerated avatars wouldn't actually be too much work :smile: ).

Jakolo121:
We can use the "profile Picture" for the prompter. The prompter is already familiar with the Icon. 
![image](https://user-images.githubusercontent.com/34916464/210546642-1e645d8a-721c-4297-9642-7a5702303966.png)

fozziethebeat:
I like where this is going

fozziethebeat:
[...]

PR #329, opened 2023-01-03
Website implement create initial prompt

othrayte:
In think this implements #275, thought it seemed a bit easy so maybe I missed something.

I think someone should look over the instructions on the page, maybe we can come up with something better.

Also the e2e tests got broken when the data-cy attributes disappeared from the elements. Note sure what happened, I can even find them being removed in the git history :man_shrugging:.

### DIFF
--- a/website/cypress/e2e/create/initial_prompt.cy.ts
+++ b/website/cypress/e2e/create/initial_prompt.cy.ts
@@ -1,0 +1,26 @@
+import { faker } from "@faker-js/faker";
+
+describe("creating initial prompts", () => {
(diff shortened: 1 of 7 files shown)


AbdBarho (approved):
@othrayte thank you!

tested it locally and it works.

I think we should activate the e2e tests as part of the CI.

for the instruction, we have #159.

othrayte:
@AbdBarho, agree totally, I started working on #274 as soon as I submitted this PR, I'm getting there but might need some help with using docker compose in GitHub actions, or else some better way to run the stack to test it. If that fits anyone's skillset message me in the open assistant discord.

Issue #229, opened 2023-01-01
Design an informative `/about` on the website.


fozziethebeat:
@yk has written a [comprehensive slide deck](https://docs.google.com/presentation/d/1n7IrAOVOqwdYgiYrXc8Sj0He8krn5MVZO_iLkCjTtu0/edit#slide=id.p) with a roadmap and philosophy.  We should copy parts of this and make a really nice looking `/about` page on the website.

brentbagapuro:
Hi, I'd like to work on this issue. I can take parts from the slide deck you mentioned and copy them onto different sections of the page. I noticed the homepage currently is using a Tailwind UI template (Pocket), so I can take inspiration from there, as well as integrate other Tailwind UI components as needed.

fozziethebeat:
Great! Assigned this to you.

brentbagapuro:
@fozziethebeat  Shall I include the slides below the "Getting to MVP" part?

fozziethebeat:
I think for a first draft we can skip those as they're more operational.

brentbagapuro:
Thanks for the quick response! I'll submit a PR in a few.

PR #335, opened 2023-01-03
about page

brentbagapuro:
This change resolves issue [#229](https://github.com/LAION-AI/Open-Assistant/issues/229): Design an informative /about on the website

### DIFF
--- a/website/src/components/Roadmap.tsx
+++ b/website/src/components/Roadmap.tsx
@@ -1,0 +1,78 @@
+import { Container } from "./Container";
+
+const Roadmap = () => {
+  return (
+    <Container className="">
+      <div className="py-32">
+        <h2 className="text-4xl mb-16">Our Roadmap</h2>
+        <div className="flex flex-col items-center space-y-8 md:space-y-0 md:items-start md:flex-row md:justify-between">
+          <div className="flex flex-col items-center space-y-4">
+            <div className="h-[5rem] w-[5rem] border-4 border-[#a72a1e] rounded-full flex items-center justify-center">
+              <p className="font-bold text-[#a72a1e] text-center">ASAP</p>
+            </div>
+            <h4 className="font-bold text-xl text-[#a72a1e] text-center max-w-[10rem]">Minimum Viable Prototype</h4>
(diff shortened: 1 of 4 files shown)


website/src/components/Roadmap.tsx
```diff
@@ -0,0 +1,78 @@
+import { Container } from "./Container";
+
+const Roadmap = () => {
+  return (
+    <Container className="">
+      <div className="py-32">
+        <h2 className="text-4xl mb-16">Our Roadmap</h2>
+        <div className="flex flex-col items-center space-y-8 md:space-y-0 md:items-start md:flex-row md:justify-between">
```
AbdBarho, line 8: @brentbagapuro thanks for the contribution!

I don't think it is necessary for this page since it only shows info with little interaction. But at some point we have to migrate to using chakra-ui.

For any future pull requests, prefer chakra-ui instead of raw html if possible ❤.

I cannot judge the content so I will wait for other reviewers from the team. But it looks good to me.

brentbagapuro: Thank you for you review @AbdBarho ! Appreciate it.

I understand. Was only basing on current layout of the homepage as it's mostly Tailwind styling as I did here. I also consulted with @fozziethebeat about my approach to use Tailwind UI when asking to assign the issue to me.

Nevertheless, I'll make sure future PRs prioritize chakra-ui for building the UI.

Thank you again! ❤

fozziethebeat: I answered in the issue a bit too quickly.  @AbdBarho is correct, Chakra-UI components are preferred over raw tailwind whenever possible but as a first draft it's not a big deal.  We can switch things out later

fozziethebeat (approved):
I think the content is a good first start.  We can always spruce up the wording later.  In a follow up PR, can you replace more parts with Chakra components? Namely `Header` to make it a bit easier to see what the core text is.

website/src/components/Roadmap.tsx
brentbagapuro: Gotcha. I'll take note of that. Thanks!

PR #342, opened 2023-01-03
refactor + test: add OasstApiClient in website with tests

### LINKED ISSUE #204: Contract tests between API and website
As per discussion in #180, the communication between the website and API should be tested through a Contract test. This test should test that each call satisfies the OpenAPI schema inputs and endpoints. That way, changes to the endpoint contract will automatically be caught by these tests.

jack-michaud:
Fixes #204 

- refactor: move new task's oasst api fetching into OasstApiClient
- test: add contract tests for fetchTask and ackTask
- docs: Add TODOs to refactor and add more tests
  - #353 and #354
- ci: run frontend contract tests in Test API Contract workflow

### DIFF
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -1,0 +1,63 @@
+import { JWT } from "next-auth/jwt";
+
+class OasstError {
+  message: string;
+  errorCode: number;
+  httpStatusCode: number;
+
+  constructor(message: string, errorCode: number, httpStatusCode: number) {
+    this.message = message;
+    this.errorCode = errorCode;
+    this.httpStatusCode = httpStatusCode;
+  }
+}
+
+export default class OasstApiClient {
+  constructor(private readonly oasstApiUrl: string, private readonly oasstApiKey: string) {}
+
+  private async post(path: string, body: any): Promise<any> {
+    const resp = await fetch(`${this.oasstApiUrl}${path}`, {
+      method: "POST",
+      headers: {
+        "X-API-Key": this.oasstApiKey,
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(body),
+    });
+
+    if (resp.status == 204) {
+      return null;
+    }
+
+    if (resp.status >= 300) {
+      try {
+        const error = await resp.clone().json();
+        throw new OasstError(error.message, error.error_code, resp.status);
+      } catch (e) {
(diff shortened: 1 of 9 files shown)


website/src/pages/api/update_task.ts
```diff
@@ -36,6 +36,7 @@ const handler = async (req, res) => {
 
   // Send the interaction to the Task Backend.  This automatically fetches the
   // next task in the sequence (or the done task).
+  // TODO Move this into OasstApiClient.
   const interactionRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/tasks/interaction`, {
```
fozziethebeat, line 40: Are you tackling this in a follow up PR?

fozziethebeat (approved):
This looks good.  I'm approving given that you'll need others to review before it's ready to merge.  Please update the comment to reference the issue and ensure the issue isn't closed yet.

website/src/lib/oasst_api_client.ts
```diff
@@ -0,0 +1,63 @@
...
+      body: JSON.stringify(body),
+    });
+
+    if (resp.status == 204) {
+      return null;
+    }
+
+    if (resp.status >= 300) {
+      try {
+        const error = await resp.clone().json();
```
AbdBarho, line 34: quick question, why cloning here?

website/src/pages/api/update_task.ts
jack-michaud: bdcfae5

website/src/lib/oasst_api_client.ts
AbdBarho: Aha! makes sense.

I guess using the `content-type` of the response is not reliable enough?

would it make sense to do it like this
```js
const text = await resp.text();
try {
  const error = JSON.parse(text);
  throw new OasstError(error.message, error.error_code, resp.status)
} catch (e) {
  // not json
  throw new OasstError(text, 0, resp.status);
}
```
this is equivalent right?

jack-michaud: Truth be told, I didn't think to use `content-type`. Using a `try` `catch` is a little more defensive. 

The way you've suggested is clearer and equivalent. I'll switch it!

jack-michaud: 695e1f3

AbdBarho (approved):
Thanks! waiting on non-web devs approval.

yk (approved):
thank you very much, amazing!

Issue #272, opened 2023-01-02
Can't submit ranking on website until order changed


othrayte:
When ranking replies or prompts on the website the submit button is disabled until at least one reply or prompt has been moved. This means that in order to submit the order unchanged the user needs to move something and then move it back. I think the submit button should always be enabled so long as there is an active task.

Note: this actually made it slightly harder in writing the e2e tests as originally I had planned to just leave them in the initial ranking ans submit but in the end I had to find a way to have the test change the rank. It might also mean that there are additional undue ways that test can fail.

yk:
I think it's good that we prevent the user from just clicking "submit" right away. Maybe we could do this: if the user clicks submit without touching anything, make a small popup that essentially says "you haven't touched anything, are you sure this is the correct order?" that needs to be confirmed.

othrayte:
That's a good point, it's quite probably that new users would do it accidentally, no matter how good our instructions get.

fozziethebeat:
I like Yannic's proposal.

stephancill:
Hey, I'm giving this a shot. Any feedback would be appreciated.

When try to submit without changing anything the following happens:

1. Alert popup (modal is probably more appropriate if we go this route)
<img width="972" alt="Screenshot 2023-01-03 at 20 56 50" src="https://user-images.githubusercontent.com/5469870/210423220-f93f0dcb-837a-42e1-a85d-950122193c70.png">

2. Ranking is prepared and button changes
<img width="972" alt="Screenshot 2023-01-03 at 20 56 57" src="https://user-images.githubusercontent.com/5469870/210423457-5194d9a5-64fe-4b40-ae9d-0b7a8a512108.png">

I could remove the popup and just change the button as well if that's preferred.

yk:
@stephancill looks neat! I think the common approach is to do popups not via alert, but via custom components. I'm sure chakra-ui (which we use) has something prepared, could you give that a try? could you do a PR with your proposed changes?

othrayte:
For my 2 cents, I think having a popup of some kind is clearer than just changing the button text as that might get missed. On the other hand, I think the ideal would be for the popup to include the options to edit or submit anyway. That said take that with a grain of salt as I'm not all that familiar with what our chosen UI framework has in terms of pop-overs that we can use or how much work that would be to use an integrated popup.

stephancill:
Can we close this issue?

PR #345, opened 2023-01-03
Add warning when submitting unchanged ranking

stephancill:
Fix for #272.

When the list is unchanged and the user tries to submit, they are presented with this modal:
Screenshot 2023-01-04 at 01 19 43

Something that may be an issue is that once the modal is presented, if the user presses Cancel, and they click submit again, it won't be presented again. More information on L27 of website/src/components/Survey/TaskControlsOverridable.tsx. Not sure how important this is but will required additional refactoring to fix.

### DIFF
--- a/website/src/pages/evaluate/rank_assistant_replies.tsx
+++ b/website/src/pages/evaluate/rank_assistant_replies.tsx
@@ -4,7 +4,7 @@ import { useState } from "react";
 import { LoadingScreen } from "src/components/Loading/LoadingScreen";
 import { Sortable } from "src/components/Sortable/Sortable";
 import { SurveyCard } from "src/components/Survey/SurveyCard";
-import { TaskControls } from "src/components/Survey/TaskControls";
+import { TaskControlsOverridable } from "src/components/Survey/TaskControlsOverridable";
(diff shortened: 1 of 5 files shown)


website/src/pages/evaluate/rank_assistant_replies.tsx
```diff
@@ -27,6 +27,7 @@ const RankAssistantReplies = () => {
   const { trigger } = useSWRMutation("/api/update_task", poster, {
     onSuccess: async (data) => {
       const newTask = await data.json();
+      console.log("response", newTask);
```
fozziethebeat, line 30: Could we drop this?

fozziethebeat (changes requested):
Looks good with one small change needed.  Will approve after that and after verifying another PR didn't cause a merge conflict.

website/src/pages/evaluate/rank_assistant_replies.tsx
stephancill: Oops, my bad. Fixed.

andreaskoepf:
(I removed the backend label since the proposed change is in the website code.)

fozziethebeat (approved):

Issue #340, opened 2023-01-03
docusaurus site for docs?


andrewm4894:
might be overkill for now but on the other hand is very easy to get up and running and maybe easier to do now (and evolve sanely overtime) rather than later (i'd be able and happy to help as have used docusaurus for my work: https://learn.netdata.cloud/, https://blog.netdata.cloud/). 

https://github.com/facebook/docusaurus 
https://docusaurus.io/docs/deployment#deploying-to-github-pages

could probably host it on github pages or via netlify very easy too.

yk:
what would be the required steps for us? what would go in there, the `docs/` pages?

andrewm4894:
docs can still live in /docs as markdown files so easy to work on for anyone. 

just some extra docusarus specific [config stuff](https://docusaurus.io/docs/configuration) and a github action then to deploy what's in /docs to a github page.

so could start with as minimal or flat/obvious a structure as possible and just sort of package up something very similar to existing .md files (will a little bit of a [frontmatter header](https://docusaurus.io/docs/next/create-doc#doc-front-matter) on them). 

i probably could bash together a POC in a branch on my fork easy enough.

andrewm4894:
here is a branch with a default docusarus starter https://github.com/andrewm4894/Open-Assistant/tree/docs-site-poc

deployed to github pages here: https://andrewm4894.github.io/Open-Assistant/

github.io domain in reality would be https://LAION-AI.github.io/Open-Assistant/ but easy to get a custom domain too if wanted.

andrewm4894:
actually wait thats just default gh pages boiler - will need to do one or two little things to get it using docusaurus

yk:
sounds good! if you or someone else wants to take this on, all clear :)

andrewm4894:
made draft pr here: https://github.com/LAION-AI/Open-Assistant/pull/346

will work on it a little and ping when something ready to look at - 90% sure is easy enough and could be useful while also very minimal friction or overhead for contribution and changes. 

basic moving parts:

- /docs is now a docusaurus project (https://github.com/andrewm4894/Open-Assistant/tree/docs-site-poc/docs)
- gh action to deploy docs site to `gh-pages` branch (https://github.com/andrewm4894/Open-Assistant/blob/docs-site-poc/.github/workflows/deploy-docs-site.yaml)
[...]

PR #346, opened 2023-01-03
Docs site poc

andrewm4894:
related to this issue: https://github.com/LAION-AI/Open-Assistant/issues/340

you can see the docs site create by this branch here: https://andrewm4894.github.io/Open-Assistant/

changes:
- add docusaurus project under /docs
- add gh action to deploy changes under /docs
- run yarn install within devcontainer so users can just do `cd docs; yarn start` to work on docs and see impacts if they want.

to try this out

create codespace on https://github.com/andrewm4894/Open-Assistant/tree/docs-site-poc

```bash
cd docs
yarn install
yarn start
```

Then the docs site will by default be served on port 3000 so you can look in there and changes should be reflected automatically. 

So as you work on docs then you can see impact straight away. 

image


andrewm4894:
@yk i think this is pretty much ready for review. 

I see a few PR's updating .md files etc so if we decide to get a docs site like this going i can handle any conflicts we might have initially. Just thinking be handy to put it in place sooner (if that's decided) as i can imagine will be lots of docs changes. 

Last thing i will do is update /docs/README.md to try make it clear and obvious how to make additions or changes to the docs.

andrewm4894:
if we merge this then someone with access would just need to go to the github pages part of the repo settings and do this:

image

We would also need to make one last commit (or small follow on PR) - to update all the `andrewm4894` parts of the docusaurus config with `LAION-AI`. (just have to use andrewm4894 for now to make sure my example site builds from my fork).

docs/.gitignore
```diff
@@ -0,0 +1,32 @@
+# Dependencies
+docs/node_modules
```
bitplane, line 2: `node_modules` will ignore `docs/node_modules` and the same for the other dupes.

bitplane (commented):
I don't really have the skill level to review js/ts files so just stuck my dirty oar in about .gitignore while reading about docusaurus. It looks really cool. Nice work :)

andrewm4894:
@yk @andreaskoepf what you think about this - i'm happy to take on any donkey work around maintaining it, helping users add to it and things like that. Just thinking a decent docs site might help onboard people easier over next while and maybe help them to not have to dig through .md files on GH (which they could still do too if they wanted since this really just about packaging up existing .md files to given them a bit of an obvious structure).

.devcontainer/post_create_command.sh
```diff
@@ -1,2 +1,7 @@
 # ensure pre-commit is installed
 pre-commit install
+
+# run yarn install in docs folder
+cd docs
+yarn install
+cd ..
```
yk, line 7: maybe we can leave this away? how much time does it add?

docs/docusaurus.config.js
```diff
@@ -0,0 +1,112 @@
...
+// Note: type annotations allow type checking and IDEs autocompletion
+
+const lightCodeTheme = require("prism-react-renderer/themes/github");
+const darkCodeTheme = require("prism-react-renderer/themes/dracula");
+
+/** @type {import('@docusaurus/types').Config} */
+const config = {
+  title: "Open Assistant",
+  tagline: "Build the assistant of the future!",
+  url: "https://andrewm4894.github.io",
```
yk, line 11: this probably needs to become the main url :)

yk (approved):
Very cool, thank you!

.devcontainer/post_create_command.sh
andrewm4894: Takes a few seconds, maybe worth having in default container to make it super easy for people to build docs site in a codespace if they want. Of course all they read need do is change a .md file and the workflow will build or fail so if not like they need to. 

But maybe handy to have there to make docs contributions as frictionless as possible.

docs/docusaurus.config.js
andrewm4894: Yep, I can change all that in one last commit now before we merge this. Was just leaving like that so my site was working. 

Will make follow up commit to this branch to replace all the andrewm4894 with LAION-AI

andrewm4894:
I will resolve conflicts and finalize the urls in one last commit. Will ping again once done.

Issue #315, opened 2023-01-03
Create script to fill database with mock-data


andreaskoepf:
Write a python script that fills the backend postgresql DB with random message-trees and users. Insert 500k messages (each on averge 500 characters long) of 1k randomly selected mock-users (all numbers parameters). See the [seed data generation script](https://github.com/LAION-AI/Open-Assistant/blob/a926703eec5ba58468b3d06aa9accfce1920a215/backend/main.py#L91) for inspiration. Report the query time of the current leaderboard stats query on your system.

Background:
We discussed some possible strategies for dealing with DB queries. e.g. querying user stats live/online vs. incrementally updating peristent views (e.g. for user stats). We would like to test the performance of some naive count-queries, e.g. as in the [ get_user_leaderboard()](https://github.com/LAION-AI/Open-Assistant/blob/a926703eec5ba58468b3d06aa9accfce1920a215/backend/oasst_backend/prompt_repository.py#L715-L721) of the PromptRepository class.

- This test code could be extended in the future to include message reactions, deleted messages and more.

Nil-Andreu:
Hello!
I suppose that the messages have to have words that make sense of could be only random characters?
All those have to be created on the fly each time, or maybe created once and store in a json. So when we fill the database, it is always the same messages?

yk:
> Where do we have to extract the messages of the sample?
> All those have to be created on the fly each time, or maybe created once and store in a json. So when we fill the database, it is always the same messages?

I think for creating this many messages, it's ok if we just either use the same few messages over and over again, or we use something like `faker` to create fake data on the fly. makes no sense to store thousands and thousands of messages in json just to do load tests.

Nil-Andreu:
Okay cool would like to contribute! Will take a look into that @onegunsamurai.

Nil-Andreu:
I am working on this [PR](https://github.com/LAION-AI/Open-Assistant/pull/417), which wants to:
- Store api clients
- Store users & messages with PromptRepository

andreaskoepf:
related https://github.com/LAION-AI/Open-Assistant/issues/322

nolofi:
this is what i got from Chat GPT: 

import psycopg2
import random
import string
import time

def random_string(string_length=500):
    """Generate a random string of fixed length."""
    letters = string.ascii_letters + string.digits
    return ''.join(random.choice(letters) for i in range(string_length))

def insert_messages(conn, cursor):
    """Insert 500k messages into the database."""
[...]

PR #349, opened 2023-01-03
[WIP] Fill PostgresDB Script for #315

onegunsamurai:
This script should fill DB with: 
- 1000 users (Done)
- 500K Random message trees (WIP)

### DIFF
--- a/backend/filldb.py
+++ b/backend/filldb.py
@@ -1,0 +1,153 @@
+import random
+import string
+import string
+import uuid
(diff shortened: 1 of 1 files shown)


andreaskoepf:
I had recommended to use the PromptRepository class. The script as it is now will be outdated in 2-3 days since we will have further change of the DB models, i.e. it creates another dependency that needs to be maintained. Can't merge this as is is.

Issue #339, opened 2023-01-03
in `evaluate/rank_user_replies` i would need to see what the replies are in response to


andrewm4894:
I would need to see what it was these replies were responding to to properly rank them.

<img width="960" alt="image" src="https://user-images.githubusercontent.com/2178292/210434785-77a2c43d-0d57-4e9e-b54a-e7c77cab3f4c.png">

PR #350, opened 2023-01-04
Context messages

kostiak:
Adding a potential fix for #339 

It looks like this now:
image

Let me know if I need to change anything.

### DIFF
--- a/website/src/components/ContextMessages.tsx
+++ b/website/src/components/ContextMessages.tsx
@@ -1,0 +1,19 @@
+import { Box } from "@chakra-ui/react";
+import { Message } from "./Messages";
+
+export const ContextMessages = ({ messages } : { messages: Message[]}) => {
+    console.log(messages)
+    return (
+        <Box className="flex flex-col gap-1">
+            {messages.map((message, i) => {
+                return (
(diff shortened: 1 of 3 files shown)


website/src/components/ContextMessages.tsx
```diff
@@ -0,0 +1,17 @@
+import { Box } from "@chakra-ui/react";
+import { Message } from "./Messages";
+
+export const ContextMessages = ({ messages }: { messages: Message[] }) => {
+  return (
+    <Box className="flex flex-col gap-1">
+      {messages.map((message, i) => {
+        return (
+          <Box key={i}>
+            <span>{message.is_assistant ? "Assistant: " : "User: "}</span>
```
fozziethebeat, line 10: Later this could look good with a Chakra label or some icon.

fozziethebeat (approved):
Functional and simple! In it goes

website/src/components/ContextMessages.tsx
kostiak: Complete agree. Got something basic working, and left the design for someone who's stronger in that area.

PR #352, opened 2023-01-04
[fix] rename old summary human feedback dataset to new one

theblackcat102:
Can verify that all test cases have finished without any errors and new source share the same format as the old one.


andreaskoepf (approved):
thnx for standardization

sanagno (approved):
Thanks for sharing this, I was actually looking into this today as well

PR #356, opened 2023-01-04
Create role field in website User table and be able to seed it#236

Jason-Kim1023:
I was trying to test this, but I do not know how to set up the backend to test whether or not the roles updated.  Also, the discord/email authentication does not seem to be setup, so I do not know where the admin discord/emails would be held.  Currently, the admin data would have to be saved in a config file, but that seems unrealistic (I would expect it to be held in some database with the admins email/discord). Also, the signin button doesnt seem to go anywhere, so I don't know how to test whether or not the callback is working or updating.  Made a draft PR for thoughts and whether or not I should wait for these changes to be made by others.  @fozziethebeat

### DIFF
--- a/website/src/pages/api/auth/[...nextauth].ts
+++ b/website/src/pages/api/auth/[...nextauth].ts
@@ -59,7 +59,7 @@ if (boolean(process.env.DEBUG_LOGIN) || process.env.NODE_ENV === "development") 
   );
 }
 
-export const authOptions: AuthOptions = {
+const authOptions: AuthOptions = {
   // Ensure we can store user data in a database.
   adapter: PrismaAdapter(prisma),
   providers,
@@ -70,6 +70,49 @@ export const authOptions: AuthOptions = {
   },
   session: {
     strategy: "jwt",
+  },
+  callbacks: {
+    async signIn(user, account) {
+      // Check if the user should be marked as an admin.
+      const adminUsers = process.env.ADMIN_USERS.split(",");
+      if (account && account.providor === "discord") {
+        const discordID = account.id;
+        const discordUsername = account.username;
+        const discordIdAndUsername = `${discordUsername}#${discordID}`;
+        if (adminUsers && (adminUsers.includes(discordIdAndUsername))) {
+          // Mark the user as an admin.
+          await prisma.user.update({
+            where: {
+              id: user.id,
+            },
+            data: {
+              role: "ADMIN",
+            },
+          });
+        }
+      }
+      if (adminUsers && (adminUsers.includes(user.email))) {
+        // Mark the user as an admin.
+        await prisma.user.update({
+          where: {
+            id: user.id,
+          },
+          data: {
+            role: "ADMIN",
+          },
+        });
+      }
+      else {
+        await prisma.user.update({
+          where: {
+            id: user.id,
+          },
+          data: {
+            role: "general",
+          },
+        });
+      }
+    },
   },
 };
 
--- a/website/src/pages/api/auth/[...nextauth].ts
+++ b/website/src/pages/api/auth/[...nextauth].ts
@@ -1,3 +1,4 @@
+import "./config.js"
 import { PrismaAdapter } from "@next-auth/prisma-adapter";
 import { boolean } from "boolean";
 import type { AuthOptions } from "next-auth";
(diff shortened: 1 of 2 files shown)


fozziethebeat:
I see our developer docs are out of date.

In testing, were you able to boot up the docker setup? At the project's root directory you should be able to do:
```
docker compose up frontend-dev --build
```

Then in a separate terminal, do:
```
cd website/
npx prisma db push
npm run dev
```
That gets the full backend running and a web specific database.

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -71,6 +72,48 @@ export const authOptions: AuthOptions = {
...
+        }
+      }
+      if (adminUsers && adminUsers.includes(user.email)) {
+        // Mark the user as an admin.
+        await prisma.user.update({
+          where: {
+            id: user.id,
+          },
+          data: {
+            role: "ADMIN",
```
Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:102]: small nit, can this be `admin` rather than `ADMIN`?

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -71,6 +72,48 @@ export const authOptions: AuthOptions = {
   session: {
     strategy: "jwt",
   },
+  callbacks: {
+    async signIn(user, account) {
+      // Check if the user should be marked as an admin.
+      const adminUsers = process.env.ADMIN_USERS.split(",");
```
Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:78]: Rather than generating this on every `signIn` call can you do it at the top level and turn this into a map keyed by provider that points to a set of admin IDs?

Something like:
{
 'discord' -> Set('fozziethebeat, andreaskopf'),
 'email' -> Set('fozziethebeat@gmail.com'),
}

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -71,6 +72,48 @@ export const authOptions: AuthOptions = {
...
+      if (account && account.providor === "discord") {
+        const discordID = account.id;
+        const discordUsername = account.username;
+        const discordIdAndUsername = `${discordUsername}#${discordID}`;
+        if (adminUsers && adminUsers.includes(discordIdAndUsername)) {
```
Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:83]: Could you turn these two checks into a function? That way we can do

const isAdminuser = checkAdminAccount(adminUsers);
if (isAdminUser) {
  // Do prisma update
}

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -71,6 +72,48 @@ export const authOptions: AuthOptions = {
...
+        // Mark the user as an admin.
+        await prisma.user.update({
+          where: {
+            id: user.id,
+          },
+          data: {
+            role: "ADMIN",
+          },
+        });
+      } else {
```
Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:105]: We can probably skip this

website/src/pages/api/auth/[...nextauth].ts
Comment (fozziethebeat) [reply website/src/pages/api/auth/[...nextauth].ts:83]: Also, could we only do the prisma write if the user doesn't already have the `admin` role set? Ideally we write to the database just once.

fozziethebeat (commented):
Some small suggestions to simplify the code.

Jason-Kim1023:
yah, Ill get on this as soon as possible.  I am deathly sick right now, so I may have to take a few days break and I will get back on this issue.  Essentially, you want me to make a map of the provider as the key and the set of userID (discord id or email) as the value at the top level and use that to compare with the current user's authentication method, correct?  when I get back on it, ill proceed with this - I am just going to make a map at the top and then in the function just do a comparison between the user's authentication method/the map and if its in the map then set the user's role

fozziethebeat:
> 

Oof, do rest well.  Quite a few people seem to be getting sick.

For injecting the list of admin IDs, we can use an environment variable (something like `process.env.ADMIN_IDS`)

fozziethebeat:
Based on a call we had, I need to get this feature in a bit faster to unblock other features.  I reworked the changes you have in this PR and found there's a few parts that need to change and made a new PR.  Thank you for initially looking into this~

fozziethebeat:
I've submitted by own PR that meets the needs of the request.  I'll be closing this PR

PR #359, opened 2023-01-04
Added Dashboard

rsandb:
Cleaned the git history to re-PR. The only change that's been made from the previous PR was adding account settings back to the user menu. Otherwise it's unchanged

### DIFF
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,6 +1,0 @@
-{
-  "name": "Open-Assistant",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {}
-}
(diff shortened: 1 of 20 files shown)


k-nearest-neighbor (approved):
From the other PR:

Runs great and code looks good. This obviously is large enough that it's hard to nitpick so we can do any fixes in follow ups. But I don't see much jumping out as needing fixes.

There is one thing on the header: the items should be constrained to w-max-7xl with flex and justify-between. That will get the logo and the auth stuff to align with the content. But we can fix in a follow up.

fozziethebeat (approved):
Will merge as soon as the `node_modules` folder at the root is deleted.

rsandb:
> Will merge as soon as the `node_modules` folder at the root is deleted.

done!

PR #367, opened 2023-01-04
ER-Diagrams in data_schemas.md

jojopirker:
I wanted to better understand the databases involved so I made some ER-diagrams. May be helpful for other people as well so I though I might share it. I necessary I can also share to raw xml files (from [draw.io](draw.io)).


fozziethebeat (approved):
I love documentation~

andreaskoepf:
Thanks a lot for the documentation/diagram! I am a bit worried that it will be outdated during the next days since there will be a couple of changes. But I guess that is always the problem with documentation. @yk should we merge it anyway?

jojopirker:
We can also wait a few days, and I'll update it then.

docs/data_schemas.md
```diff
@@ -203,3 +203,27 @@ message RankingExample {
 }
 
 ```
+
+## Databases
+
+Open-Assistant uses two databases, one for the backend and one for the frontend.
```
andreaskoepf, line 209: there is actually a third one for the bot which is a small sqlite db. So backend and frontends all have their own persistent state.

andreaskoepf (approved):
In a future update you could mention the sqllite db of the discrod-bot. Schema is still microscopic .. right now, will probably grow a bit over the next days.. https://github.com/LAION-AI/Open-Assistant/blob/main/discord-bot/bot/db/schema.sql

andreaskoepf:
> We can also wait a few days, and I'll update it then.

I'll merge it now. Woulb be really cool if you check in 1-2 weeks and update it to the latest state! :-) Thanks for your contribution!

Issue #273, opened 2023-01-02
Backend (and website) accept empty replies


othrayte:
I'm assuming this is a bug and empty assistant/prompter replies should be filtered out, or should they, is there a valid reason to allow empty replies? The same probably applies to initial prompts but there is no way to create them on the website at the moment.

GraemeHarris:
I'm happy to take this on to quickly fix - just need to double check how we intend to end prompt dialogues

yk:
> I'm happy to take this on to quickly fix - just need to double check how we intend to end prompt dialogues

I think this is largely independent. There's never really an "end", as the conversation could always continue

GraemeHarris:
Have updated the `TextReplyToMessage` protocol class with pydantic to ensure the text string is not blank, but I think while we're here I could also update the `MessageRating` and `MessageRanking` classes to constrain for a non-zero rating value and for the ranking list to not be empty?

Issue #173, opened 2022-12-30
Add a description to the github repository


AbdBarho:
This should make the project easier to find through github search and such.

![Screenshot 2022-12-30 094527](https://user-images.githubusercontent.com/24505302/210051683-961f7f8e-b1be-4141-bc3e-e112d83ef2f6.jpg)

Nil-Andreu:
What about:
"Revolutionary chat based language model for extraordinary conversations."

yk:
> "Revolutionary chat based language model for extraordinary conversations."

very market-y ;) I'll go with the statement from the roadmap so far, but open to ideas

yk:
done

PR #373, opened 2023-01-04
Bugfix #173 - Empty task interactions

GraemeHarris:
Added some pydantic validation updated to the protcol types to prevent empty text, lists and ratings less than zero for user replies: https://github.com/LAION-AI/Open-Assistant/issues/273


andreaskoepf (approved):
ok, we are getting pydantic.. ;-) ...

Issue #338, opened 2023-01-03
add tooltips in FE app as much as possible


andrewm4894:
i think it could be useful to have tooltips or maybe even "i" icons to get as much info or context i might need from within the app 

for example i'm not really sure what "Reply as user" means, what is it, what is expected of me, etc.

this could help as users can maybe get 80% of context they need from within the app. 

perhaps a popup model that could be used throughout the app might work.

<img width="955" alt="image" src="https://user-images.githubusercontent.com/2178292/210433105-86ee9a7b-b303-402d-9099-02f4655e4a27.png">

andrewm4894:
another example - im not sure what these flags are for

<img width="947" alt="image" src="https://user-images.githubusercontent.com/2178292/210434299-596a6fc7-47b2-4025-80bd-a903ad9b8d61.png">

Shtoner:
Just made the solution in a reusable function component with JS and CSS hover, does it need to converted to TypeScript? 
import utilStyles from '../styles/utils.module.css'

export default function Tooltips({children,props}){
    return(
        <div className={utilStyles.tooltip}
        title={props}>
                <main>{children}</main>

        </div>
    )
}

Shtoner:
TBH it might not even be worth it to add a reusable component and you could just use the CSS needed by adding the tooltip className directly into JSX along with needed data attribute(changed from title so there wouldn't be a duplicate that's not styled).
CSS here 
```
.tooltip:hover::after{
  background: rgba(0,0,0,.8);
  border-radius: 5px;
  color: #fff;
  content: attr(data);
  padding: 5px 15px;
  position:fixed;
  z-index: 3;
}
```

fozziethebeat:
Chakra has a [tooltip](https://chakra-ui.com/docs/components/tooltip) component.  Please use that and start adding it everywhere that's sensible.

When making the changes, please update just a few components at a time (or one task and it's relevant components).  That'll keep the PRs small and fast to review and merge.

fozziethebeat:
I don't think this is essential for the MVP.  It's a nice to have but more of a usability improvement.  Removing this from the milestone to cleanup the high priority items.

fozziethebeat:
I'm going to close this super bug in favor of specific features needing tooltips

PR #374, opened 2023-01-04
#338 - Tooltip for Report button

Klotske:
Worked on #338

As requested by comment https://github.com/LAION-AI/Open-Assistant/issues/338#issuecomment-1370184588 added tooltip for the report buttons

**Before:**
Before
**After:**
After

### DIFF
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -15,6 +15,7 @@ import {
   SliderThumb,
   SliderTrack,
   Spacer,
+  Tooltip,
   useBoolean,
   useId,
 } from "@chakra-ui/react";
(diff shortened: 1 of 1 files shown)


andrewm4894:
cool, thanks, might even be able to use this to try maybe add some myself elsewhere perhaps. 

one thing is what does "Report" mean? is it to report or flag something as sensitive or inappropriate, or broken in some sense? 

just asking as im actually not sure so wondering if something we can say thats more specific without getting too wordy.

website/src/components/FlaggableElement.tsx
```diff
@@ -69,11 +70,15 @@ export const FlaggableElement = (props) => {
     >
       <Grid templateColumns="1fr min-content" gap={2}>
         <PopoverAnchor>{props.children}</PopoverAnchor>
-        <PopoverTrigger>
-          <Button h="full">
-            <FlagIcon className="w-4 text-gray-400 group-hover:text-gray-500" aria-hidden="true" />
-          </Button>
-        </PopoverTrigger>
+        <Tooltip hasArrow label="Report" bg="red.600">
+          <div>
```
AbdBarho, line 74: is this div necessary?

Klotske: Yeah. The Tooltip component refuses to work if its child element is a PopoverTrigger. Same thing in reverse. Tried several approaches but only this one worked.

fozziethebeat: I concur, the Chakra docs show using `Box` which is basically `div`.

fozziethebeat (approved):

Issue #307, opened 2023-01-03
display the latest N messages on the website


fozziethebeat:
The backend now provides a `/api/v1/messages/` endpoint that returns by default the 10 latest messages stored.   We should present these somewhere on the home page when a user is logged in.

To finish this we need:
- [ ] A new API route that fetches messages
- [ ] A new component that use `useSWRImmutable` to fetch some messages and then presents them in a decent format, probably a table.

When presenting the messages, we should present the `text` and `is_assistant` fields.  The client side should retain the `id` and `parent_id` for later features.

fozziethebeat:
I think we can treat them separately.  This is more structured as a news feed of messages that have been submitted globally.  I expect the messages in this feed would be unrelated to each other.

Visually they *could* share the same components but I think it's also sensible for them to have distinct styles since their purposes diverge.

jojopirker:
I would like to try this!

fozziethebeat:
It's yours!

AbdBarho:
I will close this, it seems to be fixed in #375

PR #375, opened 2023-01-04
#307 new page to display recent messages

jojopirker:
Initial implementation for #307. 
Looks like this now: 
Screenshot

Unfortunately the flagging currently does not work..


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const bgColor = useColorModeValue(idx % 2 === 0 ? "bg-slate-800" : "bg-black", "bg-sky-900");
+
+  return (
+    <FlaggableElement text={item.text} post_id={item.id} key={`flag_${item.id}`}>
+      <HStack>
+        <Avatar
+          name={`${item.isAssistant ? "Assitant" : "User"}`}
+          src={`${item.isAssistant ? "/logos/logo.png" : "/images/temp-avatars/av1.jpg"}`}
+        />
+        <Box className={`p-4 rounded-md text-white whitespace-pre-wrap ${bgColor} text-white w-full`}>{item.text}</Box>
```
AbdBarho, line 15: does this work in dark mode?

website/src/components/Messages/MessageTable.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <Box
+      backgroundColor={backgroundColor}
+      boxShadow="base"
+      dropShadow={accentColor}
+      borderRadius="xl"
+      className="p-6 shadow-sm"
+    >
+      {isLoading ? (
```
AbdBarho, line 16: could you please move the `isLoading` check to the parent component?

website/src/pages/api/messages/user.tsx
```diff
@@ -0,0 +1,30 @@
...
+  //TODO: add params if needed
+  const params = new URLSearchParams({
+    username: token.sub,
+  });
+
+  const messagesRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/messages?${params}`, {
+    method: "GET",
+    headers: {
+      "X-API-Key": process.env.FASTAPI_KEY,
+      "Content-Type": "application/json",
```
AbdBarho, line 21: you arguably don't need content-type of a GET because it has no body.

website/src/pages/messages/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+              </Box>
+            </SimpleGrid>
+          </Box>
+        </Box>
+      </Box>
+    </>
+  );
+};
+
+MessagesDashboard.getLayout = (page) => (
```
AbdBarho, line 59: is this layout used somewhere else? I remember seeing something similar, if you find a similar layout please move it to its own component

AbdBarho (changes requested):
Thanks for the PR!

website/src/components/Messages/MessageTable.tsx
jojopirker: https://user-images.githubusercontent.com/19796430/210630496-7b82c897-2f84-468c-be6c-7e9f2f2f302f.mov

sure, i just had it in there to show small spinners... but could also change it to show a loading screen

website/src/pages/api/messages/user.tsx
jojopirker: removed it

website/src/pages/messages/index.tsx
jojopirker: is removed

website/src/components/Messages/MessageTable.tsx
jojopirker: now removed... but still looks the same :)

rsandb:
Can you make the padding between the title and div consistent with the widgets on the dashboard?

fozziethebeat (approved):
It looks like all the other requested changes have been made.  I approve but won't merge until @AbdBarho gets a chance to look again.

AbdBarho (approved):

PR #376, opened 2023-01-04
Automate e2e test of the website

### LINKED ISSUE #274: Automate Cypress (UI component & E2E) tests

othrayte:
At the moment these are enabled for PRs that change the website and for pushes to main. Note it does take 3-10mins to run.
Closes #274 (though it only runs the e2e tests and doesn't run the component tests at the moment, but we only have one at the moment anyway).

See https://github.com/othrayte/Open-Assistant/pull/1 and https://github.com/othrayte/Open-Assistant/pull/2 for an example of what a passing and failing PR looks like after this change. Note that you can both see in the log what failed and download images of the failure and videos of the run.

Edit: Please pay particular attention to when the workflow is triggered, what I have is just my initial assumptions and even I'm now thinking it might need changing -othraye.


.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - website/**
+  pull_request:
+    paths:
+      - website/**
```
jack-michaud, line 11: Even though this is an e2e test, I think we're okay just running when `website/**` changes. Interface changes in the backend should be caught by the light weight contract tests. Just calling it out because I think what you have makes sense. 

(Though I'm biased because I wrote the contract tests ;) )

jack-michaud (approved):
Took a look because I've worked with Github actions before and I'm also writing tests. 

Thanks for providing failing and successful builds! I love that you can view the failing tests in the artifacts. So cool.

.github/workflows/test-e2e.yaml
othrayte: Hmmm, now you've made me think about it again 😅. The main reason for it is so that we ignore untested changes in the bot or docs and such. I would probably be in favour of triggering on backend changes, as a kind of smoke test it exists to catch the case that you never expect to fail, including incompatible changes deeper in the backend than just the interface.

jack-michaud: You have a good point. It would be safer to run them for back end changes as well, just in case. A smoke test has value the contract test doesn't provide. It just stinks that it takes so long to run

othrayte: In the end all questions about when to trigger it really need to be decided by @yk or @andreaskoepf as it impacts the project fairly broadly. I realise now I should have called that out in the PR description.

fozziethebeat: I think changes to `website` and `backend` make sense.

bitplane (approved):
Looks good to me :)

othrayte:
Additional examples:
* Not triggering if no web or backend change https://github.com/othrayte/Open-Assistant/pull/3
* Triggering on backend change https://github.com/othrayte/Open-Assistant/pull/4

yk:
do I see this correctly, this will build & push the images on ever single push to main? because so far we're only doing that on release. I'm fine with building all the time, but is it possible that you also change the release workflow, specifically what needs to happen is that the ansible playbook must take as a parameter the image tag (default latest) that must be deployed, and then the release workflow must pass the release version tag to the ansible playbook. I.e. when we make the release "v0.1.0", then the ansible playbook should deploy v0.1.0, and not latest (as it does now).

let me know if you can do this. if not, we can just merge this and take care of ansible later.

Second possibility: don't push the docker images, but build them locally in the workflow and use them from there

Third possibility: run the e2e tests as a cron job every night

othrayte:
@yk, that's something I didn't consider, though if I'm reading https://github.com/docker/metadata-action#latest-tag correctly the latest tag should not be updated on the push and pr builds, just the release (tag) one. So maybe no changes are required.

If we are happy to have the extra builds there I think it provides a performance benefit as we test the browsers in parallel but use the same docker images. I don't see any issues with passing a specific tag to the ansible job, we even have the tags that are now provided by the docker build workflow that we can pass to ansible to select the exact image we just built (which is how the e2e tests do it).

othrayte:
I've just raised an alternative simpler PR as an option that should have less potential for fallout. Take a look. #439

othrayte:
I assume no-one has any objections to declining this now that we have merged the alternative https://github.com/LAION-AI/Open-Assistant/pull/439

fozziethebeat:
No objections~

PR #380, opened 2023-01-04
TaskSelectionTooltips

Shtoner:
This adds tooltips into the task selection page once signed in.

### DIFF
--- a/website/src/components/TaskSelection/TaskOption.tsx
+++ b/website/src/components/TaskSelection/TaskOption.tsx
@@ -1,6 +1,7 @@
-import { Card, CardBody, Flex, Heading } from "@chakra-ui/react";
+import { Card, CardBody, Flex, Heading, Tooltip } from "@chakra-ui/react";
 import Image from "next/image";
 import Link from "next/link";
+
 
 export type OptionProps = {
   img: string;
@@ -12,6 +13,8 @@ export type OptionProps = {
 export const TaskOption = (props: OptionProps) => {
   const { alt, img, title, link } = props;
   return (
+    
+    <Tooltip label={label}>
     <Link href={link}>
       <Card
         maxW="300"
@@ -35,5 +38,6 @@ export const TaskOption = (props: OptionProps) => {
         </CardBody>
       </Card>
     </Link>
+    </Tooltip>
   );
 };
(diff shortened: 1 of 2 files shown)


website/src/components/TaskSelection/TaskOption.tsx
```diff
@@ -12,6 +13,8 @@ export type OptionProps = {
 export const TaskOption = (props: OptionProps) => {
   const { alt, img, title, link } = props;
   return (
+    
+    <Tooltip label={label}>
```
notmd, line 17: You didn't deconstruct the label props. Also, please add the type to the `OptionProps`.

website/src/components/TaskSelection/TaskSelection.tsx
```diff
@@ -1,7 +1,7 @@
-import { Flex } from "@chakra-ui/react";
+import { Flex, Tooltip } from "@chakra-ui/react";
 import { useColorMode } from "@chakra-ui/react";
 import React from "react";
-
+import styles from '../../styles/Home.module.css'
```
notmd, line 4: this style is not used anywhere, you should remove it.

website/src/components/TaskSelection/TaskSelection.tsx
```diff
@@ -26,46 +26,57 @@ export const TaskSelection = () => {
...
           alt="Reply as Assistant"
           img="/images/logos/logo.svg"
           title="Reply as Assistant"
           link="/create/assistant_reply"
+          label='Respond to the user as the AI'
+
         />
       </TaskOptions>
-      <TaskOptions key="evaluate" title="Evaluate">
+      <TaskOptions key="evaluate" data-size="Evaluate">
```
notmd, line 49: Accident?

Shtoner: Indeed. I saw you said this is currently not being used but I'll leave it up to you if it should be committed for possible future cases if we go back to these components or if they will be deleted all together.

fozziethebeat:
> I believe the TaskSelection component is not currently being used. With PR #359 we have a dashboard page that already has self-explanatory entries. Here is an example: Example

This is correct.  This change now needs to go into `website/src/components/Widgets/TaskOption.tsx`

PR #381, opened 2023-01-04
344: Create tasks for text labels

olliestanley:
This adds backend support for text label tasks for initial prompts and replies. Also adds support for the tasks to the discord-bot and the text-frontend, but not the website


backend/oasst_backend/prompt_repository.py
```diff
@@ -282,18 +282,27 @@ def store_task(
                 payload = db_payload.AssistantReplyPayload(type=task.type, conversation=task.conversation)
 
             case protocol_schema.RankInitialPromptsTask:
-                payload = db_payload.RankInitialPromptsPayload(tpye=task.type, prompts=task.prompts)
```
yk, line 285: nice catch :)

backend/oasst_backend/prompt_repository.py
```diff
@@ -282,18 +282,27 @@ def store_task(
...
                 payload = db_payload.RankAssistantRepliesPayload(
-                    tpye=task.type, conversation=task.conversation, replies=task.replies
+                    type=task.type, conversation=task.conversation, replies=task.replies
                 )
 
+            case protocol_schema.LabelInitialPromptTask:
+                payload = db_payload.LabelInitialPromptPayload(type=task.type, prompt=task.prompt)
+
+            case protocol_schema.LabelPrompterReplyTask:
+                payload = db_payload.LabelPrompterReplyPayload(type=task.type, reply=task.reply)
```
yk, line 301: shouldn't `conversation` be a member here?

backend/oasst_backend/prompt_repository.py
```diff
@@ -282,18 +282,27 @@ def store_task(
...
+
+            case protocol_schema.LabelAssistantReplyTask:
+                payload = db_payload.LabelAssistantReplyPayload(type=task.type, reply=task.reply)
```
yk, line 304: same

oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -256,9 +283,10 @@ class TextLabel(str, enum.Enum):
     slang = "slang"
 
 
-class TextLabels(BaseModel):
+class TextLabels(Interaction):
```
yk, line 286: this will add a `user` member to `TextLabels`, which has some effects on other parts and might break protocol, although I agree the change is necessary. Thus, a few other places need to be changed:
- the prompt repository -> store_text_labels, the user needs to be stored. this probably even requires a database schema migration to save the user (like in the other interactions)
- the LabelTextRequest endpoint (api/v1/text_labels) should be adjusted
- website and bot requests that depend on the text_labels endpoint should be adjusted

I realize this is a lot. if you feel like this might be a bit too much, just change this back to `BaseModel` and we'll take it from there

text-frontend/__main__.py
```diff
@@ -11,6 +11,31 @@
 # debug constants
 USER = {"id": "1234", "display_name": "John Doe", "auth_method": "local"}
 
+VALID_LABELS = [
```
yk, line 14: eventually we should probably send the valid labels in the task to the client

yk (commented):
Thank you very much, really nice work. I've left some comments

andreaskoepf:
@olliestanley would it be possible for you to join our OA discord server and ping me? https://discord.gg/HFCPfugy

backend/oasst_backend/prompt_repository.py
andreaskoepf, line 285: good catch!

text-frontend/__main__.py
andreaskoepf: That's a good idea. Could we maybe also add a couple of positive labels?

andreaskoepf: ok, I see we have at least ... "creative", "beautiful",  "informative",

olliestanley: I used the same list which was already present in the `TextLabels` enum in `protocol.py`. Suggest we move any changes to the labels list to a separate issue/PR to avoid this one getting too broad

olliestanley: With the latest changes the backend should be sending the list of valid labels as part of the task. I still need to test to verify this works

yk (approved):
LGTM now, thank you very much! maybe let's wait for andreas' final comments

yk:
also, once we merge, please post this into the breaking-changes channel and tag the web-team

olliestanley:
Message ID is now sent out with the task as requested by @andreaskoepf and I have tested using the `text-frontend` to successfully complete label tasks. Will wait for any final comments from Andreas before merge and can then notify the web team of the breaking change in the `text_labels` endpoint

backend/alembic/versions/2023_01_05_1745-20cd871f4ec7_added_user_to_textlabels.py
```diff
@@ -0,0 +1,30 @@
...
+Create Date: 2023-01-05 17:45:15.696468
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "20cd871f4ec7"
+down_revision = "8d269bc4fdbd"
```
andreaskoepf, line 14: we need to fix the alembic tree, if you touch this PR .. could you please change down_revision to the latest?

backend/oasst_backend/models/text_labels.py
```diff
@@ -15,6 +15,9 @@ class TextLabels(SQLModel, table=True):
             pg.UUID(as_uuid=True), primary_key=True, default=uuid4, server_default=sa.text("gen_random_uuid()")
         ),
     )
+    user_id: UUID = Field(
+        sa_column=sa.Column(pg.UUID(as_uuid=True), sa.ForeignKey("user.id"), nullable=False, primary_key=True)
```
andreaskoepf, line 19: could it be `primary_key=True` here is a mistake?

backend/oasst_backend/prompt_repository.py
```diff
@@ -388,12 +411,12 @@ def insert_reaction(self, task_id: UUID, payload: db_payload.ReactionPayload) ->
     def store_text_labels(self, text_labels: protocol_schema.TextLabels) -> TextLabels:
         model = TextLabels(
             api_client_id=self.api_client.id,
+            message_id=UUID(text_labels.message_id),
```
andreaskoepf, line 414: if we require that the sent message id it always the read server message UUID we should change the type to `TextLabels.message_id` to UUID.

oasst-shared/oasst_shared/schemas/protocol.py
andreaskoepf: Could you please make a notification in the #breaking-changes for this?

oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -256,12 +287,13 @@ class TextLabel(str, enum.Enum):
...
     """A set of labels for a piece of text."""
 
+    type: Literal["text_labels"] = "text_labels"
     text: str
     labels: dict[TextLabel, float]
-    message_id: str | None = None
+    message_id: str
```
andreaskoepf, line 296: as mentioned above I think `message_id` should get type UUID

andreaskoepf (changes requested):
overall looks good .. added some minor comments.

backend/alembic/versions/2023_01_05_1745-20cd871f4ec7_added_user_to_textlabels.py
olliestanley: This should be fixed now but feel free to double check it's correct as I'm not too familiar with alembic

backend/oasst_backend/models/text_labels.py
olliestanley: Yep good spot, removed

backend/oasst_backend/prompt_repository.py
olliestanley: Have changed this to UUID

oasst-shared/oasst_shared/schemas/protocol.py
olliestanley: Will do on merge

andreaskoepf (approved):
ok, lets merge this! :-)

PR #383, opened 2023-01-04
Implement task selection

MattAlexMiracle:
This implements a simple self-stabilizing task selection system:
- ranking
- answering
- prompting
- voting

The goal of the algorithm ``task_selection`` is to balance prompting vs voting, such that the number of active prompts stays constant.
Further, we try to get tasks done as soon as possible, meaning that in the case a prompt is ready for ranking, we will preferentially (with probability `p`) rank first.

If the algorithm decides to "answer" a prompt, a second function ``next_answer_task`` exists that, given a dictionary of prompt-ids and existing number of answers, decides which prompt to provide an answer to next.
This also prefers close-to-done prompts so that we can start ranking quickly.

If the selected task doesn't exist (i.e. there are no prompts to answer) I would just re-sample which task to choose at the moment (though with the automatic balancing I'm not entirely sure whether this is even possible).


andreaskoepf (approved):
very nice, thank you!

PR #402, opened 2023-01-05
Starting point to make the documentation more modular and better README.md

Jac-Zac:
I have added a table of content to the README.md and the logo at the start, please tell me if there is anything that you want me to modify. Also I have created a CONTRIBUTING.md and moved the `I’m in! Now what?` section into it. I think this is a good starting point if in the future we want to add more documentation in general, I feel like this makes the github documentation easier to read and a bit cleaner


yk (approved):
thank you!

PR #404, opened 2023-01-05
Detoxify Rest API Client

Nil-Andreu:
Create a REST API Client to get the Detoxify Classification from HuggingFace.


backend/oasst_backend/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Raises:
+        OasstError: error in the response from the API
+
+    Returns:
+        Dict[str, Any]: _description_
+    """
+    headers: Dict[str, str] = {"Authorization": f"Bearer {hf_token}"}
+    payload: Dict[str, Any] = {"inputs": msg}
+    response: Response = requests.post(HUGGINGFACE_TOXIC_ROBERTA_URL, headers=headers, json=payload)
```
andreaskoepf, line 25: Could you please use `aiohttp` instead of `requests`? see https://github.com/LAION-AI/Open-Assistant/blob/894219423d72ae77f984f99b584a9c84ef73632d/oasst-shared/oasst_shared/api_client.py#L44

backend/oasst_backend/api/v1/hugging_face.py
```diff
@@ -0,0 +1,30 @@
...
+def get_text_toxicity(
+    msg: str,
+    hf_token: str,
+    api_client: ApiClient = Depends(deps.get_trusted_api_client),
+) -> List[List[ToxicityClassification]]:
+    """Get the Message Toxicity from HuggingFace Roberta model.
+
+    Args:
+        msg (str): the message that we want to analyze.
+        hf_token (str): the token for the huggingface access.
```
andreaskoepf, line 22: The HF token could better be stored in the backend  configuration, e.g. as a member of our pydantic [Settings classs](https://github.com/LAION-AI/Open-Assistant/blob/main/backend/oasst_backend/config.py).

backend/requirements.txt
```diff
@@ -6,6 +6,7 @@ numpy==1.22.4
 psycopg2-binary==2.9.5
 pydantic==1.9.1
 python-dotenv==0.21.0
+requests==2.28.1
```
andreaskoepf, line 9: don't forget to remove...

andreaskoepf (changes requested):
Thank you very much for the Detoxify call & endpoint!
Requesting the following changes:
- use aiohttp
- store hf key in settings class instead of query-param

andreaskoepf (approved):
ok, thanks. we probably should close the http session but I can add this later.

Nil-Andreu:
> ok, thanks. we probably should close the http session but I can add this later.

Added this in the [PR](https://github.com/LAION-AI/Open-Assistant/pull/411).

PR #405, opened 2023-01-05
Added Redis and Redis Insight to Ansible playbook.

rasdani:
I basically converted the [changes](https://github.com/LAION-AI/Open-Assistant/pull/187/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9) in the ```docker-compose.yaml``` to the ansible playbook and tweaked it a bit.
I further added an ansible inventory file for local testing and docs.


rasdani:
ran `pre-commit` before adding the `README.md` and forgot to run again

yk:
Very nice, thank you so much. One question I had was about the config file. Does the docker plugin copy this automatically or do we need to copy it first to the target machine in order to mount it?

rasdani:
I recall vaguely the config file beeing mounted from the working directory into the docker container.
If lines 28 and 29 in the playbook work like they would in a docker setup, the config should be mounted:
`        volumes:
          - "./redis.conf:/usr/local/etc/redis/redis.conf"
`

yk:
yes, my question is whether the mount is from the working directory on the managed node (in which case we'd have to add some copy action before) or whether ansible does this somehow magically from the working directory of the control node

rasdani:
Ah, I see. I tested this by spinning up an ubuntu container on my local machine an let ansible ssh into it (don't recommend this, annoying). 
When we provide ansible with a file to mount as a volume, an empty directory gets mounted in the redis instance on my fake remote host. After trying to mount a directory containing a test file as volume, I couldn't see the file in the redis instance either.
However all of the above works fine when testing on localhost.

Sharp-eyed of you pointing this out!
I stumbled upon something with file/copy in the ansible docs and could dig deeper, when I find time.
Who is in charge of deploying to prod anyway?

rasdani:
figured it out :)

yk:
thank you :) could you remove your inventory file? then it's perfect

rasdani:
you're welcome, see you on youtube ;)

yk:
thank you :)

yk (approved):

PR #411, opened 2023-01-05
Session Close when called HuggingFace Client

Nil-Andreu:


backend/oasst_backend/utils/hugging_face.py
```diff
@@ -46,4 +46,6 @@ async def post(self, input: str) -> Any:
         # Get the response from the API call
         inference = await response.json()
 
+        await session.close()
```
andreaskoepf, line 49: Could you please use a `async with` statement for the session (it will then be closed automatically) so that the client is also closed in case of an exception?

andreaskoepf (changes requested):
please use `async with` ...

backend/oasst_backend/utils/hugging_face.py
Nil-Andreu: Sure thanks! Done

andreaskoepf (approved):
thanks for the super-quick update!

Nil-Andreu:
> thanks for the super-quick update!

Cheers! Thanks for the feedback!

PR #417, opened 2023-01-05
Fill DB with Mock Data

Nil-Andreu:
Script to fill the database with mock data.


Nil-Andreu:
I am wondering what I am doing wrong when making the git pull from the upstream branch, as then it shows duplicate commits from mine.

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -35,15 +35,17 @@ async def post(self, input: str) -> Any:
...
-        response = await session.post(self.api_url, headers=self.headers, json=payload)
-
-        # If we get a bad response
-        if response.status != 200:
-            raise OasstError("Response Error Detoxify HuggingFace", error_code=OasstErrorCode.HUGGINGFACE_API_ERROR)
-
-        # Get the response from the API call
-        inference = await response.json()
+        async with aiohttp.ClientSession() as session:
+            payload: Dict[str, str] = {"inputs": input}
```
andreaskoepf, line 39: this is still strange. If you cannot resolve the strange git merge problem, could you maybe just create a new branch on current HEAD or main and add the two files filldb.py and the filldb.md ?

docs/filldb.md
```diff
@@ -0,0 +1,26 @@
+# Fill Database
+
+## Requirements
```
andreaskoepf, line 3: I am not sure why one would need to/want to set the skip api check to false? Could you explain or remove this section if it is not necessary? thx,

backend/filldb.py
```diff
@@ -0,0 +1,190 @@
...
+
+from oasst_backend.api.deps import api_auth
+from oasst_backend.database import engine
+from oasst_backend.models import ApiClient
+from oasst_backend.prompt_repository import PromptRepository
+from oasst_shared.schemas.protocol import User as ProtocolUser
+from sqlmodel import Session
+
+
+# TODO: Should put settings into False
```
andreaskoepf, line 16: Not sure what this means. Please clarify/remove.

backend/filldb.py
```diff
@@ -0,0 +1,190 @@
...
+class FillDb:
+    """Class object which fills the database: Api Client, Users & Messages."""
```
andreaskoepf, line 18: It says "... & Messages" but I couldn't find the code that actually inserts messages? Is this still open?

andreaskoepf (changes requested):
Thnx for using PromptReposity & cmd line args! Insertion of Messages seems to be still missing. Disabling of skpi-api-key-check in md-file is not clear.

andreaskoepf:
Just noticed this is still in Draft stage .. sry for premature review.

Nil-Andreu:
> Just noticed this is still in Draft stage .. sry for premature review.

No worries! Thanks for the feedback!

Nil-Andreu:
> Just noticed this is still in Draft stage .. sry for premature review.

It is now on code review, but do not know why the Tests are failing.

yk:
@danielpatrickhug does this interfere with what you did in #613 ?

Nil-Andreu:
> @danielpatrickhug does this interfere with what you did in #613 ?

No, in fact it uses the realistic data and the work done for the creation of the messages.

danielpatrickhug:
@yk Hi, no, it doesn't look like it directly interferes w/ #613 it basically moves the seed_data logic into a FillDB, and the logic looks the same for message insertion except for a couple of changes to how the test users and API client is created. on startup, the seed_data function is still called in main.py and the realistic data should be inserted the same as it was in #613. 

From what I remember about the MockDB ticket, it was supposed to be for adding 500,000+ messages and 1000+ users and testing the leaderboard query time. Filldb should also be able to be used with a much larger dataset to accomplish that, however during instantiation of filldb, you should be able to pass in the path to the JSON file storing the messages, currently, it is hard coded.

Nil-Andreu:
> @yk Hi, no, it doesn't look like it directly interferes w/ #613 it basically moves the seed_data logic into a FillDB, and the logic looks the same for message insertion except for a couple of changes to how the test users and API client is created. on startup, the seed_data function is still called in main.py and the realistic data should be inserted the same as it was in #613.
> 
> From what I remember about the MockDB ticket, it was supposed to be for adding 500,000+ messages and 1000+ users and testing the leaderboard query time. Filldb should also be able to be used with a much larger dataset to accomplish that, however during instantiation of filldb, you should be able to pass in the path to the JSON file storing the messages, currently, it is hard coded.

Okay could do not hardcode the path to the dataset, so we could then pass another dataset that is bigger.
For the dataset of 500k+, should I use something similar as the dataset of `anthropic's helpful test dataset` that you used? If so, could you share which was the script you used to prepare the dataset?

backend/filldb.py
```diff
@@ -0,0 +1,290 @@
...
+    user_message_id: str
+    parent_message_id: Optional[str]
+    text: str
+    role: str
+
+
+class FillDb:
+    """Class object which fills the database: Api Client, Users & Messages."""
+
+    def __init__(self, db_engine, num_api_clients: int = 10, num_users: int = 10, use_seed: bool = True, seed: int = 0):
```
momegas, line 34: Should the default be something like 500K messages?
Or is it taking a lot of time?

backend/filldb.py
```diff
@@ -0,0 +1,290 @@
...
+
+                # First for that API Client, should create some random users
+                ur = UserRepository(db=db, api_client=api_client)
+
+                # Create new usser with that api client
+                ur.lookup_client_user(client_user=random_user)
+
+        return self.users
+
+    def fill_messages(self):
```
momegas, line 97: Some more docs would be very helpful here.

momegas (approved):
Looks good overall. I'm new to this project, so someone more experienced should have a. better look

Nil-Andreu:
Heey @andreaskoepf @yk, does the filling of the database already accomplished? So we can close this PR or I can keep working in it.

PR #418, opened 2023-01-05
Create openbugger_example.ipynb

furlat:
Added an example for OpenBugger that can bug python code either in string or from a imported class to generate training samples for interactive code debugging. The code will clone https://github.com/furlat/OpenBugger and pip install the package to load the SyntaxBug and LogicBug classes.


yk:
Thank you very much! could you add a small markdown file like the other notebooks have? just so people can see what's in there in case they don't have a jupyter running or aren't on github. it can be super brief.

furlat:
Added the coommit I hope is fine

yk (approved):
Wow, this is very thorough, thank you very much!

Issue #353, opened 2023-01-04
Move update task interaction into OasstApiClient


jack-michaud:
**Note**: Waiting for https://github.com/LAION-AI/Open-Assistant/pull/342 to be merged.

The website directly calls the backend through `fetch` [here](https://github.com/LAION-AI/Open-Assistant/blob/a95e71d6f9b8c46d8f6d875cb1cce12bfd549e63/website/src/pages/api/update_task.ts#L39-L57) when it should be called through the `OasstApiClient`.

1. Create a new method in `website/src/lib/oasst_api_client.ts`'s `OasstApiClient` which makes the API call done [here](https://github.com/LAION-AI/Open-Assistant/blob/a95e71d6f9b8c46d8f6d875cb1cce12bfd549e63/website/src/pages/api/update_task.ts#L39-L57).
2. Replace the original `fetch` call with the method call on `OasstApiClient`. 
3. Write a test in `website/cypress/contract/oasst_api_contract_tests.cy.ts` which calls that function and checks that the response is not null. An example of such a test can be found [here](https://github.com/LAION-AI/Open-Assistant/blob/a95e71d6f9b8c46d8f6d875cb1cce12bfd549e63/website/cypress/contract/oasst_api_contract_tests.cy.ts#L7-L15)

fozziethebeat:
I haven't been tracking what this `OasstApiClient` does.  Could you outline what it's core methods are?

fozziethebeat:
Nevermind, seeing the PR, i see this is a new thing you're creating.  I like it.

fozziethebeat:
Before closing this out, can you also do some refactoring to tackle #105?  This basically needs to change how you set the `user` field in the body being posted.

jack-michaud:
> Before closing this out, can you also do some refactoring to tackle #105?  This basically needs to change how you set the `user` field in the body being posted.

Sure, I can supply the JWT in the constructor of OasstApiClient so the client can be aware of the current user.

PR #421, opened 2023-01-05
#353 - Add OasstApiClient interaction with task

Klotske:
#353
My attempt at adding a `taskInteraction` method with its use in `api/update_task.ts` and contract test.

This is my first time working with cypress, but I hope I understood the idea.

### DIFF
--- a/website/src/pages/api/update_task.ts
+++ b/website/src/pages/api/update_task.ts
@@ -1,4 +1,5 @@
 import { getToken } from "next-auth/jwt";
+import OasstApiClient from "src/lib/oasst_api_client";
 import prisma from "src/lib/prismadb";
 
 /**
@@ -18,6 +19,8 @@ const handler = async (req, res) => {
     return;
   }
 
+  const oasstApiClient = new OasstApiClient(process.env.FASTAPI_URL, process.env.FASTAPI_KEY);
+
   // Parse out the local task ID and the interaction contents.
   const { id, content, update_type } = await JSON.parse(req.body);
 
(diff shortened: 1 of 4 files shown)


website/src/pages/api/update_task.ts
```diff
@@ -18,6 +19,8 @@ const handler = async (req, res) => {
     return;
   }
 
+  const oasstApiClient = new OasstApiClient(process.env.FASTAPI_URL, process.env.FASTAPI_KEY);
```
fozziethebeat, line 22: Instead of creating a new client object every render, can we create a single client object in `oasst_api_client.ts` and export that?  This is similar to how we do things with the prisma client.

fozziethebeat (changes requested):

website/src/pages/api/update_task.ts
jack-michaud: We could do this, but if we wanted to add the 'user' to the constructor (as mentioned in https://github.com/LAION-AI/Open-Assistant/issues/353#issuecomment-1370478078) it would need to be reworked. 

Is there some sort of request context that we could add the api client to?

fozziethebeat: We don't need to provide `user` in the constructor of the client.  Keeping it as an argument to the method calls seems fine.  The cleanup we need is just for the usage of `user` to check what type of user is it and send a different user ID if they're authenticated via discord.

NextJS Middleware seems too intrusive since we don't need this to happen for every single API route.

Klotske: Might be fixed by 3c330dc
Added prisma-like global variable init & changed exports so both `oasstApiClient` and `OasstApiClient` class could be used.

fozziethebeat (approved):

PR #422, opened 2023-01-05
devcontainer: post create run npm install in /website

andrewm4894:
Small pr to run npm install in `/website` so users in devcontainer dont need to.

### DIFF
--- a/.devcontainer/post_create_command.sh
+++ b/.devcontainer/post_create_command.sh
@@ -1,2 +1,7 @@
 # ensure pre-commit is installed
 pre-commit install
+
+# npm install in /website
+cd website
+npm install
+cd ..


yk:
does this slow down startup for anyone? or everyone? or is it cached somehow? I'm sorry I have no clue of dev containers

andrewm4894:
> does this slow down startup for anyone? or everyone? or is it cached somehow? I'm sorry I have no clue of dev containers

will slow down initial startup by just a little the first time the devcontainer is built. so is for sure a trade off between spending time up front trying to prepare everything one could possibly want vs just letting them run it themselves when they need it. 

in this case its maybe 10-20 secs to essentially make sure step 5 [here](https://github.com/LAION-AI/Open-Assistant/tree/main/website#getting-everything-up-and-running) is for sure already done. 

eg if working on multiple FE feature branches it could be nice to have so can just hope in, spin up the containers, and get quilcky to `npm run dev`. 

So on balance i think ok, but agree there will be a limit to amount of stuff it would make sense to do in `post_create_command.sh`

andrewm4894:
p.s. i could imagine if we moved it into a .Dockerfile in .devcontainer that might maybe get cached easier as a layer in the docker image. if this was a longer thing we needed to do then i think we'd be better off going that way, maybe not worth it for this. 

will double check the docs on best practices to be sure.

kostiak:
Another option we can explore is having more than one devcontainer.json file.

According to https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/introduction-to-dev-containers it's possible to make more setups by putting them in subfolders of .devcontainer.

So one possibility here is to keep the general container as is and add a specific one for the web team (and then maybe the backend team, too)

andrewm4894:
@kostiak yep - that seems to work

image

closing this in favor of https://github.com/LAION-AI/Open-Assistant/pull/445

PR #439, opened 2023-01-06
Automate e2e test of the website (basic)

othrayte:
This is an alternative to #376 that strips down the CI step to reduce any potential for unforeseen changes to other CI steps and avoids spamming the ghcr with docker builds.

I decided to test this version to prove if all the extra work the other version does was warranted and it appears if might not have been. The performance of building the container on demand actually seems to be better than building it using the reusable workflow and the only downside is that tests in different browsers would be run sequentially rather than in parallel. In the original PR it turned out however that Cypress seemed to have more to trouble working with firefox than chrome and so I figure it might be better to just test one browser for now for E2E tests as that should be more than sufficient as a smoke test (detecting browser specific failures is not really the goal atm).

Example of tests passing: https://github.com/othrayte/Open-Assistant/pull/6
Example of tests failing: https://github.com/othrayte/Open-Assistant/pull/7

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -1,0 +1,37 @@
+name: E2E Tests (Website)
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - backend/**
+      - website/**
+  pull_request:
+    paths:
+      - backend/**
+      - website/**
+
+jobs:
+  test-e2e:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Start website, backend, etc
+        run: docker compose up ci --build -d
+      - name: Run Cypress tests
+        uses: cypress-io/github-action@v5.0.2
+        with:
+          browser: chrome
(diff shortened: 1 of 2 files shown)


.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,37 @@
...
+        run: docker compose up ci --build -d
+      - name: Run Cypress tests
+        uses: cypress-io/github-action@v5.0.2
+        with:
+          browser: chrome
+          working-directory: website
+      - uses: actions/upload-artifact@v3
+        if: failure() # NOTE: screenshots will be generated only if E2E test failed
+        with:
+          name: cypress-screenshots-${{ matrix.browser }}
```
AbdBarho, line 31: `${{ matrix.browser }}` should not be there right?

othrayte: Fixed.

yk (changes requested):
can we also take `oasst-shared` into the paths?

yk (approved):
thank you!

Issue #309, opened 2023-01-03
Create a way to navigate to and view conversations branches given a message id


fozziethebeat:
Given a `message_id` we can fetch the full conversation tree it resides within and the various branches.

It's not obvious how to best do this so please experiment with different approaches and describe them here in this issue before making a pull request.

Potential API methods to user:
- `/api/v1/messages/{message_id}/tree`
- `/api/v1/messages/{message_id}/children`
- `/api/v1/messages/{message_id}/descendants`

This could also leverage #308 as a base whereby someone can click a message in the conversation, see branches and then navigate to that branch conversation.

jojopirker:
Somewhat like this? I still have some bugs with it, but may be a start... 
its just a hacky solution to display children and parents, so would need to be reworked the be more flexible 



https://user-images.githubusercontent.com/19796430/210849009-df5c03d8-d67d-4646-8091-64e555d49aff.mov

fozziethebeat:
Yes! That's certainly functional and meets the key need.  I'd be happy with starting with that and then improving it visually later.

jojopirker:
I made a draft here #440 :)

fozziethebeat:
I think I accidentally self-assigned this.  i've now assigned it to @jojopirker

AbdBarho:
could this be closed by #440 ?

fozziethebeat:
Yes, I think that's sufficient enough to close this issue.  We can open future issues for specific improvements to it.

PR #440, opened 2023-01-06
#309 -  Message navigation

jojopirker:
Draft for an implementation for #309. 

Looks like this currently:
https://user-images.githubusercontent.com/19796430/210849009-df5c03d8-d67d-4646-8091-64e555d49aff.mov


website/src/pages/api/messages/[id]/parent.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const { id } = req.query;
+
+  if (!id) {
+    res.status(400).end();
+    return;
+  }
+
+  const messageRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/messages/${id}`, {
```
Comment (fozziethebeat) [inline website/src/pages/api/messages/[id]/parent.ts:19]: Should this include `/parent` in the URL?

fozziethebeat (commented):
Broadly this looks right! Tidy up as much as you think needed and I'll do a more careful review.

website/src/pages/api/messages/[id]/parent.ts
Comment (jojopirker) [reply website/src/pages/api/messages/[id]/parent.ts:19]: I believe the backend does not provide the parent function..

Comment (fozziethebeat) [reply website/src/pages/api/messages/[id]/parent.ts:19]: I read this a bit too quickly before going to take a bath.  You are right.  This seems to be the only way right now

jojopirker:
made a few changes to make it recursive... 
Looks the same if message tree is not big, but still has some visual quirks with a bigger trees (see attached video)

@fozziethebeat maybe you got some ideas how to properly style it...
I would leave it as a draft still but I think you could already take a look at the code, please :)

https://user-images.githubusercontent.com/19796430/211035618-0bb5383a-9bfd-44dd-a556-1627a6b15569.mov

fozziethebeat:
I don't see many code suggestions to make.  It's clean and pretty easy to understand.  Later I think we can fold the backend fetch calls into a library someone is working on but I'm also trying to negotiate not having these API routes anyways so it's not a huge issue.

Will approve once you feel it's ready

jojopirker:
should be ready now :)

Screenshot 2023-01-07 at 10 23 59

fozziethebeat (approved):

AbdBarho:
This actually looks sick!!!

PR #441, opened 2023-01-06
Creating a role field for Users #236

### LINKED ISSUE #236: Create role field in website User table and be able to seed it
We need a `role` field in the website's `User` table.  By default any new user should be given the `general` role.

We also need a seed script that lets us mark one or more users as admins in advance.  That seed script should let us configure via an environment variable the set of admins via their auth method (discord or email) and the username they use (their discord username or their email address).

When a user matching this short list signs up, they should automatically be given the `admin` role.

fozziethebeat:
This creates a new field in the `User` table for the user role and sets it to `general` by default.

Then, when users sign in, it checks against a shortlist of user IDs for each provider type, if the user id matches the shortlist, they're updated to be an admin.  That role is further forwarded to the user's token and session objects so it can be verified in the client side and at the API routes.

Closes #236

### DIFF
--- a/website/.env
+++ b/website/.env
@@ -1,3 +1,5 @@
+ADMIN_USERS = "credentials:admin,discord:root,email:admin@example.com"
+
 # The database created by running the jobs in /scripts/frontend-development/docker-compose.yaml
 DATABASE_URL=postgres://postgres:postgres@localhost:5433/oasst_web
 
--- a/website/prisma/schema.prisma
+++ b/website/prisma/schema.prisma
@@ -41,6 +41,7 @@ model User {
   email         String?   @unique
   emailVerified DateTime?
   image         String?
+  role          String    @default("general")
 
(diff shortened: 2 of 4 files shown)


AbdBarho (approved):
This perhaps not necessary for the MVP, but does it make sense to use an array of roles?
In this case, a standard user would have an empty roles array.

what do you think?

fozziethebeat:
> This perhaps not necessary for the MVP, but does it make sense to use an array of roles? In this case, a standard user would have an empty roles array.
> 
> what do you think?

I've tried that before in other projects and I think it was just kind of confusing.  Also most frameworks recommend a single role value per user.  Right now I think we have a strict hierarchy of abilities associated with roles so a single value is simpler and works well enough.  When we find out there's roles that don't fully subsume each other then we should investigate how to make an array of roles work well.

Issue #437, opened 2023-01-06
Drag and drop should not be constrained to the handle only


AbdBarho:
In ranking tasks, drag and drop should not be constrained to the handle only, but should be the entire text element.

https://github.com/LAION-AI/Open-Assistant/blob/42deae20f29492759c2f9d11ec088d875b45d476/website/src/components/Sortable/SortableItem.tsx#L24-L33

othrayte:
I suggest we make sure to keep the drag handle icon so that it is still clear than it can be dragged (and after this change will also be easy to use)

othrayte:
Also this element is interacted with in the end-to-end tests so make sure to run `npm run cypress:run` or `npm run cypress` to check that those still pass.

AbdBarho:
This is already implemented.

PR #443, opened 2023-01-06
#437 sortableItem draggable

jojopirker:
fix for #437

### DIFF
--- a/website/src/components/Sortable/SortableItem.tsx
+++ b/website/src/components/Sortable/SortableItem.tsx
@@ -1,5 +1,6 @@
 import { Button } from "@chakra-ui/react";
 import { useColorMode } from "@chakra-ui/react";
+import { DragOverlay } from "@dnd-kit/core";
 import { useSortable } from "@dnd-kit/sortable";
 import { CSS } from "@dnd-kit/utilities";
 import { PropsWithChildren } from "react";
@@ -21,15 +22,17 @@ export const SortableItem = ({ children, id }: PropsWithChildren<{ id: number }>
       : "bg-black hover:bg-slate-900 text-white ring-1 ring-white/30 ring-inset hover:ring-slate-200/50";
 
   return (
-    <li
-      className={`grid grid-cols-[min-content_1fr] items-center rounded-lg shadow-md gap-x-2 p-2 ${themedClasses}`}
-      ref={setNodeRef}
-      style={style}
-    >
-      <Button justifyContent="center" variant="ghost" {...attributes} {...listeners}>
-        <RxDragHandleDots2 />
-      </Button>
-      {children}
-    </li>
+      <li
+        className={`grid grid-cols-[min-content_1fr] items-center rounded-lg shadow-md gap-x-2 p-2 ${themedClasses}`}
+        ref={setNodeRef}
+        style={style}
+        {...attributes}
+        {...listeners}
+      >
+        <Button justifyContent="center" variant="ghost" {...attributes} {...listeners}>
+          <RxDragHandleDots2 />
+        </Button>
+        {children}
+      </li>
   );
 };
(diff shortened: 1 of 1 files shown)


website/src/components/Sortable/SortableItem.tsx
```diff
@@ -25,6 +25,8 @@ export const SortableItem = ({ children, id }: PropsWithChildren<{ id: number }>
       className={`grid grid-cols-[min-content_1fr] items-center rounded-lg shadow-md gap-x-2 p-2 ${themedClasses}`}
       ref={setNodeRef}
       style={style}
+      {...attributes}
+      {...listeners}
     >
       <Button justifyContent="center" variant="ghost" {...attributes} {...listeners}>
```
AbdBarho, line 31: the attributes could be removed from here right?

fozziethebeat:
I will leave this to @AbdBarho as I haven't looked into dnd-kit at all

kostiak:
One thing that might be a problem - the "..." button I added previously for long texts will not work with this change. Got a codespaces span up to verify.

website/src/components/Sortable/SortableItem.tsx
jojopirker: Left them in for the styling of the mouse pointer... but could also take it out and style manually

AbdBarho: yeah that would be great.

AbdBarho: oh sorry, I did not literally mean the `attributes` object, I meant all dnd stuff, I fixed it quickly.

I think it looks good with the mouse drag.

I will check the `...` problem mentioned below.

AbdBarho:
> One thing that might be a problem - the "..." button I added previously for long texts will not work with this change. Got a codespaces span up to verify.

yeah, with the `...` it is complicated since the DnD listens on pointer down, and since it listens on the entire subtree you can literally drag & drop through the modal, which is not great.

Possible solutions:
* `event.stopPropagation()`
* change the layout / position of the expand or `...` button.

@jojopirker @kostiak what do you think?

jojopirker:
I never used DnDKit before so I'm not sure...

kostiak:
The reason I added it as a `...` in the first place was because the whole row wasn't selectable like it will be after this PR. I have done some work with combining two buttons on top of each other like that in the past and it's not worth it. I propose we keep the text shortening as-is but move the button somewhere else. Or change the layout somehow.

@AbdBarho @jojopirker Any suggestions on where we can put the button now? I was initially thinking of putting it at the end, but that will just take a lot of space and will not be needed most of the time.

Another, harder solution might be to remove the button and somehow detect when you start moving a component but don't actually change it's position, and trigger the modal on that.

Maybe even turn it to some kind of tooltip on hover?

othrayte:
@kostiak, that later comment is what I have toyed with in the video demo I posted in the website channel (https://discordapp.com/channels/1055935572465700980/1055940465536532571/1061258548400619560).

I've pushed what I have to my fork (https://github.com/LAION-AI/Open-Assistant/compare/main...othrayte:Open-Assistant:collapse-on-sort) but keep in mind it was an experiment so there is a lot of other mess in there from other experiments. The main change I think you'd need is setting the activationConstraint on the sensors:
```ts
    useSensor(MouseSensor, { activationConstraint: { distance: 4 } }),
    useSensor(TouchSensor, { activationConstraint: { distance: 4 } }),
```

fozziethebeat:
Are we moving forward with this PR?  Or have we fixed the underlying problem with other PRs?

fozziethebeat:
This PR has been quiet for about a week now.  I think the underlying problem has already been fixed so I'm going to close this for now.  If it's going to be resumed, please re-open.

jojopirker:
just checked, no need to re-open! thx!

PR #444, opened 2023-01-06
fix hrefs on readme badges

andrewm4894:
my bad, was a bit hasty and forgot the hrefs - added here.


andreaskoepf (approved):
thanks for the fix! :-)

PR #445, opened 2023-01-06
add support for multiple devcontainers

andrewm4894:
changes:

- add `.devcontainer/frontend/` that just pre-runs `npm install` from within the `/website/` dir. 
- add note to `.devcontainer/README.md` about this.

this looks like a useful way to add multiple different types of devcontainers that might be useful over time and is much cleaner than what i was suggesting in #422


yk (approved):
Thank you, very cool!

PR #446, opened 2023-01-06
Fix message flagging in dark mode (#434)

### LINKED ISSUE #434: Message flagging in dark mode isn't functional
When a user switches to dark mode and tries to flag a message, the checkboxes and text isn't readable.

chs20:
Fixes #434.

Note that I didn't change the checkboxes, but I think they are well visible:

message


fozziethebeat (approved):

fozziethebeat:
Thank you~

Issue #360, opened 2023-01-04
Add message id to messages in conversation model (protocol.py)


andreaskoepf:
Users should be able to give feedback for messages of a conversation when they are displayed in the fontends, e.g. during tasks to add a reply or when completed work is displayed in a recent activity view or the bot log-channel. Currently only the text and a flag whether the message is a user or an assistant message are sent to the client.

Add either a message_id field directly or in a meta-data dict. Potentially adapt other models like the MessageTree also, e.g. move meta data into a sub field, potentially create a class for standard-meta data (like date, message-id, user-id, frontend-user, frontend-message-id).

-> Align potocol with [data_schemas](https://github.com/LAION-AI/Open-Assistant/blob/main/docs/data_schemas.md) docs.

andreaskoepf:
use case e.g. see 2.) of https://github.com/LAION-AI/Open-Assistant/issues/231#issuecomment-1369768813

andreaskoepf:
Until we have a better solution, please include a `message_id` and a separate`frontend_message_id` field in `ConversationMessage`. @yk check

melvinebenezer:
https://github.com/LAION-AI/Open-Assistant/issues/371 is dependent on this fix.

PR #449, opened 2023-01-06
(WIP)Conversation messages refactor add msg ids

danielpatrickhug:
Pull Request for #360 
- Still a WIP, still need to update main and potentially update ConversationMessage types to remove optional.
#### Here the current changes
- in oasst_shared/schemas/protocal.py I modified ConversationMessage adding the variable message_id and frontend_message_id
- in backend/oasst_backend/api/v1 I modified instances of `ConversationMessage` in tasks.py and utils.py to include the new variable when instantiating `ConversationMessage`
- Ran pre-commit to format

#### What still needs to be fixed:
- I'm still trying to understand how to fix what's going on in main.py any advice would be appreciated. @andreaskoepf mentioned using trace_conversation() in prompt_repository.


yk:
is the issue still persisting? if yes, what is it? I've checked out the PR and at least it runs for me

danielpatrickhug:
@yk the instantiation of ConversationMessage in main.py now takes in a frontend_message_id and message_id, however, I'm unsure how to pass in the message_id before the message is created(pr.store_text_reply. 

I set the variable to optional so that the task can be created with the dummy message. So I think after the message is created(on line 198) the tasks need to be queried (created on  line 188)and updated to include the frontend_message_id and message_id. But this seems overly complex. I feel as if I'm missing something.

yk:
I see, thanks for the explanation. very sure @andreaskoepf knows much more than I do. the entire message-ids are kinda screwed up and we're looking to simplify those a lot shortly

Issue #369, opened 2023-01-04
Going to the webpage should redirect to `/dashboard` for signin users


fozziethebeat:
Right now if a user goes to the main website, it goes to the signed out page which doesn't have any obvious call to actions for contributing.  They can click the "dashboard" link in the user dropdown but it's not totally obvious.

Fixes might be:
- Have an obvious call to action for signed in users
- Redirect to `/dashboard` for signed in users

Jakolo121:
I'm working on it!

Jakolo121:
My take would be this...

User is not authenticated:
![image](https://user-images.githubusercontent.com/34916464/210758244-ec0a9a8f-b388-40f9-8fbe-06e52a7176d0.png)

User is  authenticated:
![image](https://user-images.githubusercontent.com/34916464/210758561-1c9fb69b-2b66-4b72-b9f2-fce0538dee8d.png)

fozziethebeat:
Is there someway we can redirect the user to `/dashboard` when they're authenticated?  I feel like that'd be the smoothest flow and reduce the number if button clicks needed.

Klotske:
Should be a pretty easy thing to do.
We have NextAuth.js `useSession()` hook to determine if user is authenticated. Then we could use NextJS `useRouter()` hook and push the `/dashboard` route.

fozziethebeat:
That sounds most ideal!

Jakolo121:
Actually in the `Signin` function, there is a Redirect callback `callbackUrl` defined as `/dashboard`. By default only URLs on the same URL as the site are allowed, so either I added `http://localhost:3000/dashboard` or I add the redirect callback to `[...nextauth].ts` so that the `baseUrl` is appended to the `callbackUrl`.

fozziethebeat:
The signin flow is working correctly.  The part we need to fix is when an already signed in user goes directly to the website.  When that happens they won't go through the signin and redirect flow, instead they'd land at `/`.

So @Klotske's solution would be the most appropriate.  When a signed in user lands at `/` we should automatically redirect them to `/dashboard`.

PR #451, opened 2023-01-06
Going to the webpage should redirect to /dashboard for signin users #369

Jakolo121:
Added the CTA Component to Hero.tsx. 
The Button should display "Login" if the User is not logged-in and should redirect to `auth/signin`.
If the User is Logged-in and wants to visit the index page, one should be redirected to `/dashboard`.

### DIFF
--- a/website/src/components/Hero.tsx
+++ b/website/src/components/Hero.tsx
@@ -64,10 +64,9 @@ function CTA() {
   const router = useRouter();
   if (session) {
     router.push("/dashboard");
-  }
-  return (
-    <div>
-      {!session ? (
+  } else {
+    return (
+      <div>
         <Link href="/auth/signin">
           <button
             type="button"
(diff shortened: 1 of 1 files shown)


website/src/components/Hero.tsx
```diff
@@ -55,6 +59,28 @@ function BackgroundIllustration(props) {
   );
 }
 
+function CTA() {
+  const { data: session } = useSession();
+  const router = useRouter();
+  if (session) {
+    router.push("/dashboard");
```
fozziethebeat, line 66: Having a redirect buried in a small component feels very weird.  Can you instead do this redirect directly inside of `website/src/pages/index.tsx`?  I think all we need is the session check and redirect.  We don't need a new log in button as we have one in the header already.

fozziethebeat (changes requested):

Issue #433, opened 2023-01-06
Refactor the web tasks to present a server selected task type


fozziethebeat:
Right now users can select the task type they want but for data collection goals it's better for the server to decide what task is most important.

We should refactor the tasks such that they're all components and a single task page (`random`) will fetch a task type and content from the server and then present whatever that task content is.

This requires moving some react code around and consolidating things then simplifying the landing page.

fozziethebeat:
This is very high priority

andrewm4894:
Could maybe add task "points" concepts or something and have a way to see which are high value vs low value tasks etc. Eg maybe lean into the gamification stuff by putting a "price" or "value" on each task and then being able to use that to control or weight tasks and maybe even let users incentive to more higher value tasks. 

Might be a bit too fancy for right now, but on the other hand if each task had a "points" associated with it that could be a useful building block to add now. Is kinda a product sort of question maybe.

Unsure myself but just throwing idea out there.

fozziethebeat:
That's a good idea in the future.  For the MVP we want to avoid creating any new feature requests for the backend.  The backend right now doesn't have a way to check if a task type has actual tasks that need answers so the best option would be to just have a "Random" option on the dashboard and let the server pick the task which needs an answer.

Later on, your idea is the right direction to go towards but needs help in the backend.

kostiak:
Is there already an endpoint that will return a task_type?

Also, do we want the content (like "Given the following prompts, sort them from best to worst, best being first, worst being last.") to come from the backend or sit in the frontend?

kostiak:
@fozziethebeat I created a draft PR to suggest how we can refactor the pages into components. It's already quite a big change, so I didn't add the "random" page yet. Want to double check I'm going in the right direction.

What I did was extract the task definitions from `/Dashboard/TaskOption.tsx` into '/Tasks/TaskTypes.tsx` and modify it to be more extendable. Then I based the components on it. 

I created a `Task.tsx` which has the general task logic. `CreateTask.tsx`  and `EvaluateTask.tsx` have the specific things of those types need.

I also moved all the task pages into /tasks/ route.

fozziethebeat:
I think this sounds right but i'd have to look at a PR to be sure

kostiak:
@fozziethebeat uploaded a new draft PR, let me know if that's more in the right direction, thanks!

kostiak:
[...]

PR #455, opened 2023-01-06
Website tasks refactor

kostiak:
This is a draft refactor of the task pages into components, relating to (but not fully implementing) #433


website/src/components/Tasks/Task.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import CreateTask from "./CreateTask";
+import EvaluateTask from "./EvaluateTask";
+import { taskTypes } from "./TaskTypes";
+
+const Task = ({ taskTypeName, taskName }) => {
+  const taskType = taskTypes[taskTypeName][taskName];
+  const [tasks, setTasks] = useState([]);
+
+  const { isLoading, mutate } = useSWRImmutable(`/api/new_task/${taskType.endpoint}`, fetcher, {
```
fozziethebeat, line 18: The one big change I think is that this initial data fetch should be at the page level.  

That'll make the second PR easier.  The specific task pages would just fetch their known task type.  That code would be duplicated a bit but oh well.  The random task would fetch data from `/api/new_task/random` and then have a switch based on what task type gets returned.  It'd be able to render any of the available task types dynamically.  And then when we re-fetch a task, it should naturally just switch between different types.

fozziethebeat (changes requested):

website/src/components/Tasks/Task.tsx
kostiak: Thanks a lot for the feedback. I'm closing this PR and will open a new one once I have something to show.

PR #456, opened 2023-01-06
Supervised finetuning minor changes

sanagno:
Small changes

- small typo when masking with padding
- better collator with comments
- started quantization


theblackcat102 (approved):
can we force label_mask to be boolean at the collator stage? such that we don't need .bool() in trainer.py line 73 ?

theblackcat102:
but I think it's just minor so going to merge it

sanagno:
Good point, that how I had it but it didn't work with deepspeed, I assume it casts tensors in a specific ways. I am going to fix it in the next commit.

model/supervised_finetuning/models/gptj.py
```diff
@@ -0,0 +1,187 @@
+# Taken from https://github.com/sleekmike/Finetune_GPT-J_6B_8-bit/blob/master/gpt-j-6b-8-bit.py
+
+import torch
```
mrcabbage972, line 3: Maybe instead use bitsandbytes for a generic solution for 8-bit quantization? Are there any downsides to that?

sanagno: Not really! feel free to edit

mrcabbage972: Also I'm pretty sure I've previously seen a generic library that implements a whole pile of different adapter types, could be convenient.

mrcabbage972: @sanagno Opened PR [here](https://github.com/LAION-AI/Open-Assistant/pull/582)

Issue #435, opened 2023-01-06
Make it easier to submit longer responses on website


fozziethebeat:
The default textbox and instructions encourage fairly short message responses.  We should:
- [ ] Make the text area larger to encourage more writing
- [ ] Add some kind of instructions that encourage more comprehensive answers

AbdBarho:
idea from meeting:
length could be indicated by colored bars that fill up just like password strength.

Klotske:
Working on it

fozziethebeat:
I'll mark this as closed for now.  We've added the progress bar to indicate response length.

PR #457, opened 2023-01-06
#435 Added longer response incentive

Klotske:
#435 

As per https://github.com/LAION-AI/Open-Assistant/issues/435#issuecomment-1373392898 added new `TrackedTextarea` component, which implements basic word count progress bar.
The word tracker has three threshold points as required props.

The ref forwarding was getting complicated _(was not sure how to properly update the word counter when setting value with `inputRef.current.value` as it does not trigger re-rendering)_ so I decided to add `inputText` state to each page.

Also, we should deside on thresholds for the response's word count.

PS
All e2e tests ran perfectly.

### DIFF
--- a/website/src/components/Survey/TrackedTextarea.tsx
+++ b/website/src/components/Survey/TrackedTextarea.tsx
@@ -1,0 +1,35 @@
+import { Progress, Stack, Textarea, TextareaProps } from "@chakra-ui/react";
+
+interface TrackedTextboxProps {
+  text: string;
+  thresholds: {
+    low: number;
+    medium: number;
+    goal: number;
+  };
+  textareaProps?: TextareaProps;
+  onTextChange: (event: React.ChangeEvent<HTMLTextAreaElement>) => void;
+}
+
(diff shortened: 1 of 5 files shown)


Klotske:
Main issue with state addition is the constant re-rendering of the whole page. Not sure how to prevent such behaviour.

website/src/components/Survey/TrackedTextarea.tsx
```diff
@@ -0,0 +1,35 @@
...
+  thresholds: {
+    low: number;
+    medium: number;
+    goal: number;
+  };
+  textareaProps?: TextareaProps;
+  onTextChange: (event: React.ChangeEvent<HTMLTextAreaElement>) => void;
+}
+
+export const TrackedTextarea = (props: TrackedTextboxProps) => {
```
fozziethebeat, line 14: I like how this is its own component.  Later one we could add more to make the writing experience better.

A small nit I might fix in the future is making this a `forwardRef` so that we can avoid having a state variable.

fozziethebeat (approved):

PR #464, opened 2023-01-06
clean up docs github workflow

andrewm4894:
now that docs site pr is merged we should only trigger github docs site deploy workflow on `main` branch. `docs-site-poc` branch was just here as part of the PR testing, not needed anymore.


yk (approved):
thank you!

PR #468, opened 2023-01-07
add colab buttons to notebooks

andrewm4894:
add open in colab buttons to notebooks. seems like they all fairly self contained and able to just run in colab.

### DIFF
--- a/notebooks/code-bugger/openbugger_example.ipynb
+++ b/notebooks/code-bugger/openbugger_example.ipynb
@@ -1,6 +1,14 @@
 {
  "cells": [
   {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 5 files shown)


yk (approved):
thank you, that's pretty cool. did you do this by hand or with a script? I wonder if we could automate this

andrewm4894:
Just by hand as only a few so far. For one of them I had to add a pip install cell at top for `detoxify`. 

So maybe just trying to make it a standard practice as opposed to anything we can automate or adding a pre-commit for. 

Am sure could be ways to automate. 

If we really wanted to then a standard notebook folder structure that supports just opening in something like binder might be a good option https://mybinder.readthedocs.io/en/latest/index.html

I might have a look and maybe make a "example.ipynb" for others as a sort of template perhaps as I can see maybe a lot of different notebooks maybe coming in all with different styles etc. 

But I'd be happy to do some clean up later rather than adding any extra friction just yet. Will see, will think a bit.

andrewm4894:
Here is a nice binder example that might be sort of typical stuff we'd want in each notebook subfolder.

https://github.com/binder-examples/requirements

Issue #432, opened 2023-01-06
Factor out the dashboard page layout into a shared layout


fozziethebeat:
Right now `website/src/pages/dashboard` includes a nice `Sidebar` with multiple menu options and then presents the select page's contents.

This is a great design that would be nice to use in other scenarios such as for an admin specific page which would have different sidebar options.

To do this I think we need to:
- [ ] Make `website/src/components/Dashboard/SideMenu` it's own independent component that takes in a list of menu options as a property.
- [ ] Factor out the layout code in `website/src/pages/dashboard` into a layout component in `website/src/components/SideMenuLayout` that also takes in the list of menu options.
- [ ] Create several concrete layouts: `GeneralSidebarLayout` and `AdminSidebarLayout`, and maybe others.
- [ ] Use `GeneralSidebarLayout` in `website/src/pages/dashboard`
- [ ] Use `AdminSidebarLayout` in future admin pages but using a different color scheme and different sidebar options.

Klotske:
@fozziethebeat I think I could take this issue. Could you assign it to me?
Self-assigning needs Github's Collaborator role i think...

fozziethebeat:
Great!@ Yeah, the assigning part is a bit of a hassle.  After the MVP we might want to have a group discussion about simplifying it.

Also, since this is quite a few changes, I'd request this be done in a couple PRs rather than one giant PR.

AbdBarho:
could this be closed?

PR #469, opened 2023-01-07
#432 - Moved SideMenu to component & refactored layout

Klotske:
#432 
Should complete tasks 1 & 2. Moved the `SideMenu` to it's own component & included it in the `SideMenuLayout`.
Updated `getDashboardLayout` in NextJS's Layout resolver to work with new component.
Reworked `/dashboard` and '/messages' pages to use the updated `getDashboardLayout`.

Also spotted wierd issue with colorMode flashing while switching between `/dashboard` and `/messages`. Might look into it a little bit later.


AbdBarho (approved):
Thanks, looks good!

Issue #436, opened 2023-01-06
Implement Labeling Tasks


AbdBarho:
For tasks where the user has to assign labels to a given text, the backend should provide these labels, instead of hard-coding them in the frontend.

This is similar to the `Report` functionality.

Question:
Should the labels be part of the task delivered from the backend, or a separate endpoint?

olliestanley:
As part of #381 (merged today) the valid labels are sent as part of the task definition when starting a prompt/reply label task

AbdBarho:
@olliestanley thank for the info! these are 3 new tasks that needs to be added to the website, correct?
`label_initial_prompt` `label_prompter_reply` `label_prompter_reply`

olliestanley:
> @olliestanley thank for the info! these are 3 new tasks that needs to be added to the website, correct?
> `label_initial_prompt` `label_prompter_reply` `label_prompter_reply`

Yep! (but one is label_assistant_reply)

PR #477, opened 2023-01-07
Label Initial Prompt

AbdBarho:
Part of #436

I started to consolidate the shared code with regards to fetching and mutating to its own hook.
The amount of code duplication is starting to get out of hand...

I still need to add the actual flagging, for that I want to refactor `FlaggableElement`, but maybe in another PR because this will become too big.

Also, now we show the raw labels delivered from the backend, we need to define corresponding texts to display.


fozziethebeat:
I like the direction this is going.  I agree that the fetching is getting way out of hand.

website/src/hooks/useLabelingTask.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const { trigger } = useSWRMutation("/api/update_task", poster, {
+    onSuccess: async (reply) => {
+      const newTask: ConcreteTaskResponse = await reply.json();
+      setTasks((oldTasks) => [...oldTasks, newTask]);
+    },
+  });
+
+  const submit = (id: string, message_id: string, text: string, labels: Record<string, string>) =>
+    trigger({ id, update_type: "text_labels", content: { labels, text, message_id } });
```
AbdBarho, line 49: @fozziethebeat I had to override the `message_id` here otherwise the (real) backend would error.

`interaction.id` from prisma did not work.

fozziethebeat: Yes that looks correct.  `interaction.id` is just a web thing that can be ignored during backend interactions.

fozziethebeat (approved):

PR #480, opened 2023-01-07
#432 - Added basic Admin Dashboard layout

Klotske:
#432 
Should be done with the issue. For now decided against moving `GeneralSidebarLayout` and `AdminSidebarLayout` into separate components. But that could easily be fixed.

Fixed `/admin` page to use new `getAdminLayout` function.

Preview:
Preview

### DIFF
--- a/website/src/components/Layout.tsx
+++ b/website/src/components/Layout.tsx
@@ -1,7 +1,7 @@
 // https://nextjs.org/docs/basic-features/layouts
 
 import type { NextPage } from "next";
-import { FiLayout, FiMessageSquare } from "react-icons/fi";
+import { FiLayout, FiMessageSquare, FiUsers } from "react-icons/fi";
 import { Header } from "src/components/Header";
 
 import { Footer } from "./Footer";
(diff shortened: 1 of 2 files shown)


website/src/components/Layout.tsx
```diff
@@ -51,4 +51,22 @@ export const getDashboardLayout = (page: React.ReactElement) => (
...
+    <SideMenuLayout
+      menuButtonOptions={[
+        {
+          label: "Users",
+          pathname: "/admin",
+          desc: "Users Dashboard",
+          icon: FiUsers,
+        },
+      ]}
+    >
```
AbdBarho, line 66: I think the admin should also be able to see the dashboard & messages, at least for development purposes.

Klotske: I can add the links as in the original dashboard layout
image

But this would present another problem. Now we are not providing layout based on the user's role. It's basically a whole another layout component for `/admin` route. 
So when admin user interacts with these links to `/dashboard` and `/messages` sidebar won't include admin's menu options.

AbdBarho: ah, I see.. that is actually a problem.

would it make sense to add the items in the original dashboard if the user is admin?

Klotske: I think it would be reasonable. But need some comments from @fozziethebeat as the original issue mentioned separate layouts.

fozziethebeat: In my mind the admin page would be a separate view that's accessible from the user drop down in the header.  It would have the same look and feel of the contributing dashboard but with different side menu options.

That way, its *clearly* a different mode of interaction.  This is why I was thinking we'd use the same set of generic components but fill them independently.

This PR matches what I was requesting.

fozziethebeat (approved):
This is great! It's exactly what I had in mind.

website/src/components/Layout.tsx
fozziethebeat: I'm going to merge this since it matched the design I had in mind (but maybe didnt' explain super clearly).  Let's discuss this in the next live call to see how we might want to change the design.

PR #500, opened 2023-01-07
Conversation messages refactor add msg ids

danielpatrickhug:
- in oasst_shared/schemas/protocal.py I modified ConversationMessage adding the variable message_id and frontend_message_id
- in backend/oasst_backend/api/v1 I modified instances of ConversationMessage in tasks.py and utils.py to include the new variable when instantiating ConversationMessage
- in backend main.py I updated the instantiation of conversation message to include message_id and frontend_id
- Ran pre-commit to format


andreaskoepf (approved):
OK, nice .. looks good! :-)

PR #501, opened 2023-01-07
Using MessageTable in EvaluateTask

kostiak:
Implemeted the MessageTable from the "Messages" section in EvaluateTask. Also removed ContextMessages as that's not needed anymore.

This is how it looks now:
image
image


AbdBarho (approved):
Thanks!

Issue #506, opened 2023-01-07
Error when requesting new task


notmd:
![image](https://user-images.githubusercontent.com/33456881/211168044-e5704abc-bebe-465d-8de8-4a1684686ac7.png)

## Step to reproduce
- Open the OpenAPI UI
- Send a request to /tasks endpoint with the following payload
```json
{
  "type": "label_prompter_reply",
  "user": {
    "id": "string",
    "display_name": "string",
    "auth_method": "discord"
  },
  "collective": false
}
```
Maybe there is more `type` facing a similar error.

notmd:
I'm working on this, but it will take time cause I'm not very familiar with Python, also it's 3 AM now. It's quite high priority issue cause it blocks the frontend development. If anyone wants to take it, feel free.

andreaskoepf:
Should be fixed by https://github.com/LAION-AI/Open-Assistant/commit/d910c310c06ee314358a86ad950977984ec7e7cf

kostiak:
@andreaskoepf can you take a look at my PR? Got the fronend working and getting tasks. That's just a temporary fix to eliminate the bottleneck and someone else should probably do the final fix.

andreaskoepf:
It was temporary fixed before the message-ids created new problems. We need tests for the backend API.

Issue #503, opened 2023-01-07
Realistic Data Sample


Nil-Andreu:
I am wondering which are the source for the messages in order to be them realistic.
Should get it from an existing **database**? Any ideas?

Related issue [here](https://github.com/LAION-AI/Open-Assistant/issues/322).

andreaskoepf:
@danielpatrickhug is currently exploring if @Rallio67's synthetic dialog dataset could be used.

ontocord:
If you have ideas for existing datasources or sets with instruction->response paths, please do mention it ! thank you

yk:
closing this in favor of #315 and #322

PR #508, opened 2023-01-07
Temporary fix for #506 to get the frontend working

kostiak:
Got a workaround working with the frontend and getting tasks. Someone with more backend experience should probably look at this.


andreaskoepf:
#503 should be fixed, does the problem still occur?

andreaskoepf (approved):
@danielpatrickhug you said you tested your changes?!

PR #509, opened 2023-01-07
Initial revision to valid text labels

olliestanley:


andreaskoepf (approved):
Much better!

oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -272,36 +272,25 @@ def __new__(cls, label: str, display_text: str = "", help_text: str = None):
...
-    formal = "formal"
-    cringe = "cringe"
-    creative = "creative"
-    beautiful = "beautiful"
-    informative = "informative"
-    based = "based"
-    slang = "slang"
+    political_content = "political_content", "Expresses political views"
+    humor = "humor", "Contains humorous content including sarcasm"
+    hate_speech = "hate_speech", "Expresses sentiment which is discriminatory against a grouping of people"
```
yk, line 289: too general. in plain, "discriminate" just means making a difference, thus we discriminate against groups of people all the time justifiably.
Oxford has "abusive or threatening speech or writing that expresses prejudice on the basis of ethnicity, religion, sexual orientation, or similar grounds", I'd go with something like that or even more explicit

andreaskoepf (approved):
Still declared as draft, but I think we can merge it.

oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -272,36 +272,25 @@ def __new__(cls, label: str, display_text: str = "", help_text: str = None):
...
-    spam = "spam"
+    spam = "spam", "Seems to be intentionally low-quality or irrelevant"
     fails_task = "fails_task", "Fails to follow the correct instruction / task"
     not_appropriate = "not_appropriate", "Inappropriate for customer assistant"
     violence = "violence", "Encourages or fails to discourage violence/abuse/terrorism/self-harm"
     harmful = (
         "harmful",
         "Harmful content",
-        "The advice given in the output is harmful or counter-productive. This may be in addition to, but is distinct from the question about encouraging violence/abuse/terrorism/self-harm.",
+        "The advice given in the output is harmful or counter-productive. This may be in addition to, but is distinct from the label for encouraging violence/abuse/terrorism/self-harm.",
```
yk, line 282: the word "harmful" needs to be defined very tightly, because it's often used to justify pretty much any intervention. we can't just have "is harmful". 
MW has "physical or mental damage" and Britannica has "physical or mental damage or injury : something that causes someone or something to be hurt, broken, made less valuable or successful, etc."
the problem here is that this definition of harm is actually often, we are perfectly fine with such harm, i.e. cutting trees for lumber, breaking bones for surgery. We should probably go with something like "excessive", "unacceptable", ur "unnecessary" harm, change the label itself to something like "excessive_harm", and have a tight definition of harm be part of the description

PR #512, opened 2023-01-08
add example notebook

andrewm4894:
this adds

- an example notebook as `notebooks/example/`
- a readme in `notebooks/example/README.md`
- a code cell at the top of `notebooks/example/example.ipynb`, to be uncommented and ran if running in colab, that will set up the colab notebook to just work and be able to use and files or dirs that live with the notebook, for example reading from a /data/ folder or installing from a requirements.txt.

Main idea here is to give an example that's as easy as just pressing the "Open In Colab", uncommenting top set up cell (if needed) and then ability to just run all cells. This is useful as you can just get a gpu instance (if needed) in colab so should make it easier for users to play with notebooks without too crazy configurations/env's locally that otherwise might block them.

I had a look at mybinder but it did not really work well (was not able to get it to open in a subfolder ([thread](https://discourse.jupyter.org/t/mybinder-jupyterlab-link-to-a-repo-subfolder-works-but-not-to-a-specific-notebook-in-that-folder/9443) suggests it should), was much slower than colab also since needed to create images etc). 

So i think Colab + set up cell at top should probably be fine and does seem to work pretty well in terms of minimal time to just being able to press "run all" in a notebook and have it run.

Thoughts? I don't mind helping to make sure any notebooks that come in are "runnable in colab" as best we can.

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -6,8 +6,8 @@ explaining what they do.
 
 ## Contributing
 
-Contributing to both notebooks and making new notebooks is very welcome. If you
-do so, make sure to make a markdown (.md) file to go with your notebook, makes
+Contributing to notebooks and making new notebooks is very welcome. If you do
+so, make sure to make a markdown (.md) file to go with your notebook, it makes
 it easier for people to know what your notebook is about.
 
 Check out the [example notebook](example/) for a reference example you can use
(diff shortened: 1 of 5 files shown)


yk (approved):
very nice, thanks for this!

PR #513, opened 2023-01-08
[feature] Add GPTJ synthetic dataset, fix reference removal regex for…

theblackcat102:
Few changes:

1. Added GPT-J generated answers to the list of datasets for reward model training

2. Fix the regex for removing reference block and other changes proposed by @agoryuno. We needed to continue experiments on the latest reward trainer so can't wait for your merge


sanagno (approved):
Looks great to me

Issue #310, opened 2023-01-03
Add ability to tag tasks as `unanswerable`


fozziethebeat:
It's very likely we'll create a task that can't be answered because it's not sensible or complete garbage.

We should add a button and API piping to ensure a user can mark any task (not individual messages but the task itself) as *not answerable*.  This should ultimately  prevent the task being served to other users (perhaps after some threshold of users mark it as unanswerable) and provide negative feedback to whomever created the messages in the task.

othrayte:
(re: end goal) Might also be worth considering number of users that skip the task as well as mark it unanswerable.

fozziethebeat:
I like that as a signal and have used it before.

Depending on how the backend implements the support, that could be doable with this issue or separately.  We currently don't report anything to the backend when we skip but I could imagine the current endpoint we're using take in alternative fields like `skip` or `tag_as_garbage`.  Then a skip or unasnwerable button would operate the same as submitting an answer.

othrayte:
UI-wise how about something like this (I'm wary of too many buttons and clutter on mobile), the wording is just an example.
![image](https://user-images.githubusercontent.com/241372/210766760-fa0bb859-47c2-4398-b824-edfa56d3bdbf.png)

fozziethebeat:
I like this UI.

othrayte:
I'll take this on, it'll give me a change to learn some of the backend code.

fozziethebeat:
Issue #344 could be useful in handling this.  That's a strategy for applying a label to an entire task.  That label set could include `skip` and `unanswerable`.

othrayte:
During the front end teams call there was some discussion about `ack`-ing and `nack`-ing and it thought that maybe the `nack` could include a reason for rejecting/skipping the task. In order to implement this task (flagging tasks that can't be completed and should not be shown to other users) we need to send feedback from the frontend to the backend and from what I found we don't yet have a suitable API call.

I have the following questions that will help me with this task:
- Should we be sending `nack` events when skipping?
- Should we add a reason to the `nack` (no-reason, cant-complete, cant-read-language)?
- If we continue to auto `ack` in the website, or choose to wait for the user to accept and then `ack`, can we later `nack` if the user skips after accepting (better that they can skip at any time for any reason so we at least get a signal and can pass the task to another user).

fozziethebeat:
From what I understood in the meeting we can `ack` and then `nack` if needed.

So I think we can:
[...]

PR #520, opened 2023-01-08
Prompt for reason for skipping.

othrayte:
This is a draft for #310 (Add ability to tag tasks as unanswerable).

It tasks onboard the feedback that we want to guide people away from skipping a bit and that we want to know why they have skipped. The leading idea for that appeared to be allowing the user to provide a custom reason, so that is the direction this went.

The main reason that this is only a draft PR is that at the moment we are not able to NACK after ACK-ing and the website auto-ACKs on fetching tasks so there is no chance to NACK. In this PR I have commented out the auto-ACK-ing to allow for testing this.

### DIFF
--- a/website/src/pages/api/new_task/[task_type].ts
+++ b/website/src/pages/api/new_task/[task_type].ts
@@ -36,9 +36,6 @@ const handler = async (req, res) => {
     },
   });
 
-  // Update the backend with our Task ID
-  await oasstApiClient.ackTask(task.id, registeredTask.id);
-
   // Send the results to the client.
   res.status(200).json(registeredTask);
 };
--- a/website/src/pages/api/reject_task.ts
+++ b/website/src/pages/api/reject_task.ts
@@ -1,0 +1,29 @@
+import { Prisma } from "@prisma/client";
+import { getToken } from "next-auth/jwt";
+import { oasstApiClient } from "src/lib/oasst_api_client";
+
+const handler = async (req, res) => {
+  const token = await getToken({ req });
+
+  // Return nothing if the user isn't registered.
+  if (!token) {
+    res.status(401).end();
+    return;
+  }
+
+  // Parse out the local task ID and the interaction contents.
+  const { id: frontendId, reason } = await JSON.parse(req.body);
+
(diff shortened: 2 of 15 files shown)


website/src/pages/api/new_task/[task_type].ts
```diff
@@ -37,7 +37,8 @@ const handler = async (req, res) => {
   });
 
   // Update the backend with our Task ID
-  await oasstApiClient.ackTask(task.id, registeredTask.id);
```
Comment (fozziethebeat) [inline website/src/pages/api/new_task/[task_type].ts:40]: Controversial proposal: What if we really delete this here and move the auto-ack to when we submit the answer?

website/src/pages/api/reject_task.ts
```diff
@@ -0,0 +1,37 @@
+import { Prisma } from "@prisma/client";
+import { getToken } from "next-auth/jwt";
+import { oasstApiClient } from "src/lib/oasst_api_client";
+
+/**
+ * Returns a new task created from the Task Backend.  We do a few things here:
```
fozziethebeat, line 6: You'll want to delete these comments

fozziethebeat (commented):
Overall I think this makes sense if we move the auto-acking.

website/src/pages/api/new_task/[task_type].ts
Comment (AbdBarho) [reply website/src/pages/api/new_task/[task_type].ts:40]: in that case, shouldn't the auto ack happen in the (real) backend?

I don't see the advantage of acking & submitting immediately.

TBH, I still don't fully understand why we ack in the first place.

Comment (fozziethebeat) [reply website/src/pages/api/new_task/[task_type].ts:40]: I agree the acking is a bit weird.  Let's discuss in the next call about how needed it really is.  For now let's pretend its required and just work around the API as is.

othrayte:
I've moved the ACK to just before submitting interactions. Also I've fixed the summerisation and labeling tasks temporarily by making them skip without NACK-ing as they don't yet use the new Task.tsx component and I wasn't sure if someone was already working on doing that and figured that would be the better way to handle them.

othrayte:
Apologies for any review request spam, I'm still learning how that action works on github.

fozziethebeat:
Looks good, we can fix any small issues later.

fozziethebeat (approved):

PR #526, opened 2023-01-08
bugfix in alembic downgrade (frontend_type not frontend_id)

jojopirker:
I had an typo in the alembic upgrade script...


yk (approved):
thanks for noticing. you can also let alembic auto-generate these things

Issue #524, opened 2023-01-08
Refine text label "harmful"


yk:
https://github.com/LAION-AI/Open-Assistant/blob/710898903d545d1fec01461d5ba2df66a25a39a5/oasst-shared/oasst_shared/schemas/protocol.py#L282

the word "harmful" needs to be defined very tightly, because it's often used to justify pretty much any intervention. we can't just have "is harmful".
MW has "physical or mental damage" and Britannica has "physical or mental damage or injury : something that causes someone or something to be hurt, broken, made less valuable or successful, etc."
the problem here is that this definition of harm is actually often, we are perfectly fine with such harm, i.e. cutting trees for lumber, breaking bones for surgery. We should probably go with something like "excessive", "unacceptable", ur "unnecessary" harm, change the label itself to something like "excessive_harm", and have a tight definition of harm be part of the description

PR #534, opened 2023-01-08
524: Refine label for harmful -> excessive_harm

olliestanley:
Addresses #524


yk (approved):
very thoughtful, thank you!

PR #539, opened 2023-01-08
536: Add endpoint to resolve frontend user by auth method and username

### LINKED ISSUE #536: Endpoint to resolve frontend user by auth method and username
Untrusted (normal) api_clients must only resolve, view and edit users that were registered by themselves. Trusted api_clients can manage all users. Returned user model contains global OA user id (UUID) and the api_client's 'local' username & auth_method.

`GET` `/frontend_users/{auth_method}/{username}`: user by `auth_method` + `username` (+ api_key which is always required)

olliestanley:
Resolves #536

### DIFF
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -1,4 +1,5 @@
 import datetime
+from typing import Optional
 from uuid import UUID
 
 from fastapi import APIRouter, Depends, Query
@@ -18,7 +19,7 @@ router = APIRouter()
 def query_frontend_user(
     auth_method: str,
     username: str,
-    api_client_id: UUID = None,
+    api_client_id: Optional[UUID] = None,
     api_client: ApiClient = Depends(deps.get_api_client),
     db: Session = Depends(deps.get_db),
 ):
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -13,6 +13,22 @@ from starlette.status import HTTP_204_NO_CONTENT
 router = APIRouter()
 
 
+@router.get("/{auth_method}/{username}", response_model=protocol.User)
+def query_frontend_user(
+    auth_method: str,
+    username: str,
+    api_client_id: UUID = None,
+    api_client: ApiClient = Depends(deps.get_api_client),
+    db: Session = Depends(deps.get_db),
+):
+    """
+    Query frontend user.
+    """
+    pr = PromptRepository(db, api_client, user=None)
+    user = pr.query_frontend_user(auth_method, username, api_client_id)
+    return user
+
+
 @router.get("/{username}/messages", response_model=list[protocol.Message])
 def query_frontend_user_messages(
     username: str,
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -27,7 +27,7 @@ def query_frontend_user(
     """
     ur = UserRepository(db, api_client)
     user = ur.query_frontend_user(auth_method, username, api_client_id)
-    return user
+    return protocol.User(id=user.username, display_name=user.display_name, auth_method=user.auth_method)
 
 
 @router.get("/{username}/messages", response_model=list[protocol.Message])
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -6,6 +6,7 @@ from oasst_backend.api import deps
 from oasst_backend.api.v1 import utils
 from oasst_backend.models import ApiClient
 from oasst_backend.prompt_repository import PromptRepository
+from oasst_backend.user_repository import UserRepository
 from oasst_shared.schemas import protocol
 from sqlmodel import Session
 from starlette.status import HTTP_204_NO_CONTENT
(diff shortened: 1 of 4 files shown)


backend/oasst_backend/prompt_repository.py
```diff
@@ -26,6 +26,24 @@ def __init__(self, db: Session, api_client: ApiClient, user: Optional[protocol_s
...
+        if not self.api_client.trusted and api_client_id != self.api_client.id:
+            # Unprivileged API client asks for foreign user
+            raise OasstError("Forbidden", OasstErrorCode.API_CLIENT_NOT_AUTHORIZED, HTTP_403_FORBIDDEN)
+
+        user_query = self.db.query(User).filter(User.auth_method == auth_method, User.username == username)
+
+        if api_client_id:
+            user_query = user_query.filter(User.api_client_id == api_client_id)
+
+        user: User = user_query.first()
```
yk, line 44: is it ok that multiple users may match here? I'd guess this can happen if a trusted frontend supplies `api_client_id=None`. When would that happen? Why would it be interested in the first?

olliestanley: The spec for the issue is that trusted clients should be able to resolve any user. I sort of assumed usernames were unique per auth method but maybe that's not the case? If not I agree it isn't clear we want the first, maybe would be better to return all matches in that case

olliestanley: This now requires an `api_client_id` so avoiding this issue

backend/oasst_backend/api/v1/frontend_users.py
```diff
@@ -6,13 +6,30 @@
...
 from starlette.status import HTTP_204_NO_CONTENT
 
 router = APIRouter()
 
 
+@router.get("/{auth_method}/{username}", response_model=protocol.User)
+def query_frontend_user(
+    auth_method: str,
+    username: str,
+    api_client_id: UUID = None,
```
andreaskoepf, line 21: since it has `None` default it probably should be `Optional[UUID]` .. if it is optional there needs to be a fallback to the `api_client.id` somewhere (since the api-key can also be specified as header field and in this case the api_client_id param would be None but the api_client would be a correct ApiClient instance...)

backend/oasst_backend/user_repository.py
```diff
@@ -1,16 +1,32 @@
...
+            # Unprivileged API client asks for foreign user
+            raise OasstError("Forbidden", OasstErrorCode.API_CLIENT_NOT_AUTHORIZED, HTTP_403_FORBIDDEN)
+
+        user: User = (
+            self.db.query(User)
+            .filter(User.auth_method == auth_method, User.username == username, User.api_client_id == api_client_id)
+            .first()
+        )
+
+        return user
```
andreaskoepf, line 28: General question @yk: What is the correct return value here we return `null`, empty response, or 404 error when the requested user is not found in the db?

andreaskoepf (commented):
- fallback to `api_client.id` or removal of explicit `api_client_id` query param
- needs decision: correct server response for user not-found case

backend/oasst_backend/user_repository.py
olliestanley: Have made it 404 in this scenario for now but let me know if this needs changes

backend/oasst_backend/user_repository.py
```diff
@@ -1,16 +1,40 @@
...
+            api_client_id = self.api_client.id
+
+        if not self.api_client.trusted and api_client_id != self.api_client.id:
+            # Unprivileged API client asks for foreign user
+            raise OasstError("Forbidden", OasstErrorCode.API_CLIENT_NOT_AUTHORIZED, HTTP_403_FORBIDDEN)
+
+        user: User = (
+            self.db.query(User)
+            .filter(User.auth_method == auth_method, User.username == username, User.api_client_id == api_client_id)
+            .first()
```
Nil-Andreu, line 30: Do we have any type check to make sure that there is only one instance in the database?

backend/oasst_backend/user_repository.py
Nil-Andreu: Getting 404 response seems reasonable to me.

backend/oasst_backend/user_repository.py
```diff
@@ -1,16 +1,40 @@
...
         self.db = db
         self.api_client = api_client
 
+    def query_frontend_user(
+        self, auth_method: str, username: str, api_client_id: Optional[UUID] = None
+    ) -> Optional[User]:
+        if not api_client_id:
+            api_client_id = self.api_client.id
+
+        if not self.api_client.trusted and api_client_id != self.api_client.id:
```
Nil-Andreu, line 23: In which case can we have an api client id that is not the same as the self.api_client_id?

backend/oasst_backend/user_repository.py
andreaskoepf: We have `ix_user_username` in the DB which ensures "api_client_id", "username", "auth_method" are unique.
Theoretically the same user-id could be used with different auth_methods, these would be two different users with different backend user ID based on the current definition of what a user is.

backend/oasst_backend/user_repository.py
andreaskoepf: Trusted api_clients can see/edit all users. Normally they would probably not go through the frontend-user query but theoretically this is possible.

backend/oasst_backend/user_repository.py
yk: 404 sounds good

yk:
looks good to me, I'll leave approval to @andreaskoepf

backend/oasst_backend/user_repository.py
Nil-Andreu: Okay thanks!

backend/oasst_backend/user_repository.py
Nil-Andreu: Okay thanks a lot!

andreaskoepf (approved):
ok, let's merge!

Issue #507, opened 2023-01-07
Compute text-embeddings for incoming meassges via HF feature-extraction pipeline


andreaskoepf:
We want to store an embedding together with each message in the DB to measure similarity and diversity (e.g. to detect (near-)duplicates). 

1. Select a model for the e.g. see https://huggingface.co/models?pipeline_tag=feature-extraction&sort=downloads (MiniLM & LaBSE were mentioned in internal discussions), potentially discuss with ML-Team on discord
2. Find way to store the embedding vector as postgres array via SQLModel and add a new (nullable) <short_modelname>_emebbding column to store the embedding of message-text, create alembic update script
3. Use the [HuggingFaceAPI](https://github.com/LAION-AI/Open-Assistant/blob/main/backend/oasst_backend/utils/hugging_face.py) class to make an asycn web call for each incoming message and store the embedding in the db. in case of an exception store NULL in the embedding field (successfully store the message anyway).
4. Create a new debug-flag in the backend settings class (default False) that allows to disable the embedding-calculations. Se the env-variable to True in the `scripts/backend_development/run-local.sh` script.

We need something similar to the embedding calculated by sbert: https://www.sbert.net/docs/usage/semantic_textual_similarity.html
maybe: https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2 ?

jojopirker:
I would take a look at this!

Would it make sense to save the embeddings in a new table? My thinking is that with a new table with the columns `message_id`, `model_name` & `embedding` we could simply store multiple embeddings and experiment with different models.

Nil-Andreu:
I think I could also take a look at this one, as it is related to classification of the messages in HF.

olliestanley:
> I would take a look at this!
> 
> Would it make sense to save the embeddings in a new table? My thinking is that with a new table with the columns `message_id`, `model_name` & `embedding` we could simply store multiple embeddings and experiment with different models.

Having a new table would make sense to me to minimise schema changes on new models

ontocord:
@SummerSigh see this issue. Similar to embedders we are building for safety. Let's all keep in contact re this so we can cross use stuff.

jojopirker:
I'll ping you guys in the discord channel :) @ontocord

SummerSigh:
@ontocord Ok! Sounds good!

yk:
if I understand correctly, this was solved in #540

PR #540, opened 2023-01-08
Store Message embedding

Nil-Andreu:
Create message embeddings and store them in the DB.

Related to this [issue](https://github.com/LAION-AI/Open-Assistant/issues/507).


backend/oasst_backend/utils/hugging_face.py
```diff
@@ -41,6 +51,10 @@ async def post(self, input: str) -> Any:
             async with session.post(self.api_url, headers=self.headers, json=payload) as response:
                 # If we get a bad response
                 if response.status != 200:
+                    from loguru import logger
```
yk, line 54: import things at top level

docker-compose.yaml
```diff
@@ -95,6 +95,7 @@ services:
       - DEBUG_SKIP_API_KEY_CHECK=True
       - DEBUG_USE_SEED_DATA=True
       - MAX_WORKERS=1
+      - DEBUG_SKIP_EMBEDDING_COMPUTATION=True
```
yk, line 98: could you also add this to the ansible playbook?

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -41,6 +51,10 @@ async def post(self, input: str) -> Any:
             async with session.post(self.api_url, headers=self.headers, json=payload) as response:
```
yk, line 51: do we know what happens if the text is longer than the intended text length of the model?

backend/oasst_backend/models/message_embedding.py
```diff
@@ -0,0 +1,21 @@
...
+import sqlalchemy.dialects.postgresql as pg
+from sqlmodel import ARRAY, Field, Float, SQLModel
+
+
+class MessageEmbedding(SQLModel, table=True):
+    __tablename__ = "message_embedding"
+    __table_args__ = (sa.PrimaryKeyConstraint("message_id", "model"),)
+
+    message_id: UUID = Field(sa_column=sa.Column(pg.UUID(as_uuid=True), sa.ForeignKey("message.id"), nullable=False))
+    model: str = Field(max_length=256, nullable=False)
```
yk, line 15: the question is, is the model name really enough to specify fully how the embedding was done? how about tokenizer settings? or text cleaning settings? or what happens if the text is too long? etc.

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -1,10 +1,20 @@
...
 from oasst_shared.exceptions import OasstError, OasstErrorCode
 
 
+class HF_url(str, Enum):
+    HUGGINGFACE_TOXIC_ROBERTA = ("https://api-inference.huggingface.co/models/unitary/multilingual-toxic-xlm-roberta",)
+    HUGGINGFACE_FEATURE_EXTRACTION = "https://api-inference.huggingface.co/pipeline/feature-extraction"
+
+
+class HF_embeddingModel(str, Enum):
+    MINILM = "sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2"
```
yk, line 15: just "MINILM" is probably a bit too scarce, given that there are many different minilms. I guess it can stay for now, but we should keep this in mind for the past

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -1,10 +1,20 @@
+from enum import Enum
 from typing import Any, Dict
 
 import aiohttp
 from oasst_backend.config import settings
 from oasst_shared.exceptions import OasstError, OasstErrorCode
 
 
+class HF_url(str, Enum):
```
yk, line 9: this may have already been here, but the class names here mix camel- and snake case

yk (changes requested):
hey, thank you. I've left a few comments. my biggest worry here is about the question what happens if the text is too long. it might not "fail", i.e. we might get a result, but the model might not be trained for that length, so it might actually be garbage.

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -274,12 +276,26 @@ def tasks_interaction(
...
+                        hugging_face_api = HuggingFaceAPI(
+                            f"{HF_url.HUGGINGFACE_FEATURE_EXTRACTION.value}/{HF_embeddingModel.MINILM.value}"
+                        )
+                        embedding = await hugging_face_api.post(interaction.text)
+                        pr.insert_message_embedding(
+                            message_id=newMessage.id, model=HF_embeddingModel.MINILM.value, embedding=embedding
+                        )
+                    except OasstError:
+                        logger.error(
+                            f"Could not fetch embbeddings for  text reply to {interaction.message_id=} with {interaction.text=} by {interaction.user=}."
```
andreaskoepf, line 296: Typo: Two space after for `for  text` ..

backend/oasst_backend/models/message_embedding.py
andreaskoepf: Since all text-porcessing is done at the HF-endpoint it would IMO be sufficient to have a key that identifies the used HF feature-extraction pipeline. If something like `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v` is stored in `model` I think that would be sufficient.

backend/oasst_backend/utils/hugging_face.py
andreaskoepf: good, point Classes/Enums should always be UpperCamelCase. Unfortunately we are dealing here with two abbreviations, e.g. HF and URL ... depending on how we want to handle them it could be `HfUrl`, `HFUrl` or `HFURL`... since we have `OasstApiClient` . `HFURL` can be ruled out .. I would suggest to use `HfUrl` ...

andreaskoepf (commented):
- Added my opinion regarding model-id

recommended changes:
- remove 2x space
- upper camel case enum name `HfUrl`

Nil-Andreu:
> * Added my opinion regarding model-id

Yes, all of the text processing is done by the endpoint. And we are also storing which is the version that we are using of that model.

> 
> recommended changes:
> 
> * remove 2x space
> * upper camel case enum name `HfUrl`



> * Added my opinion regarding model-id
> 
> recommended changes:
> 
> * remove 2x space
> * upper camel case enum name `HfUrl`

Okay, have applied also those changes.

yk (approved):
good from my side 👍

backend/oasst_backend/utils/hugging_face.py
Nil-Andreu: I am not sure about this.

docker-compose.yaml
jojopirker: When deployed via ansible the fetching of the embeddings should _not_ be skipped?

Nil-Andreu: When should be skipped? Only in local-run.sh?

jojopirker: That's why I asked again... probably also won't be necessary for Frontend development..

andreaskoepf (approved):
Looks good, thanks a lot! :-)

Nil-Andreu:
> Looks good, thanks a lot! :-)

Thankss!

Issue #224, opened 2023-01-01
Website signin page needs to display the `OAuthAccountNotLinked` error correctly.


fozziethebeat:
If a user logs in first via email and then tries to login via discord, they will get a `OAuthAccountNotLinked` error.  We should present that error with sensible instructions (login with the original provider).  NextAuth's default pages present this but our custom page doesn't handle this case well.

callum-gander:
I'll take this if that's alright

fozziethebeat:
Great! I've assigned it to you~

fozziethebeat:
@callum-gander have you been able to look at this?

callum-gander:
Started and got interrupted, will finish it today

AbdBarho:
fixed in #542

PR #542, opened 2023-01-08
#224: display OAuthAccountNotLinked error message + other errors messages

callum-gander:
I checked how NextAuth does it https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/pages/signin.tsx. They seem to pass the error in props somehow. I didn't want to add too much code or change the overall structure too much so I just added a useEffect to check whether the router contains an error. If it does, use a type guard to check if it's a string or an array, and handle them appropriately. I don't think there are cases when you'll get two errors at once, so it's mainly to stop TypeScript errors. I also added all the other error code from the above NextAuth page mentioned, so if other errors pop up they should be correctly displayed to the user. I also changed the error message for the OAuthAccountNotLinked as it didnt' seem to correctly reflect what we'd want the user to do, which is to login with the account they initially logged in with. I added the error message below the terms and conditions, I hope that's alright, obviously it can easily be moved elsewhere. I'm not sure whether you want the SignInErrorTypes and the errors object moved to a separate location? Other than this, any additional thoughts or does this satisfy the requirements?

### DIFF
--- a/website/src/pages/auth/signin.tsx
+++ b/website/src/pages/auth/signin.tsx
@@ -4,7 +4,7 @@ import Head from "next/head";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import { getCsrfToken, getProviders, signIn } from "next-auth/react";
-import React, { useRef, useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { FaBug, FaDiscord, FaEnvelope, FaGithub } from "react-icons/fa";
 import { AuthLayout } from "src/components/AuthLayout";
 import { Footer } from "src/components/Footer";
@@ -23,7 +23,7 @@ export type SignInErrorTypes =
   | "SessionRequired"
   | "default";
 
-const errors: Record<SignInErrorTypes, string> = {
+const errorMessages: Record<SignInErrorTypes, string> = {
   Signin: "Try signing in with a different account.",
   OAuthSignin: "Try signing in with a different account.",
   OAuthCallback: "Try signing in with the same account you used originally.",
@@ -45,13 +45,12 @@ function Signin({ csrfToken, providers }) {
   const [error, setError] = useState("");
 
   useEffect(() => {
-    if (router?.query?.error) {
-      if (typeof router.query.error === "string") {
-        const errorType = errors[router.query.error];
-        setError(errorType);
+    const err = router?.query?.error;
+    if (err) {
+      if (typeof err === "string") {
+        setError(errorMessages[err]);
       } else {
-        const errorType = errors[router.query.error[0]];
-        setError(errorType);
+        setError(errorMessages[err[0]]);
       }
     }
   }, [router]);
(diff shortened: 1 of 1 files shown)


callum-gander:
As a sidenote, I've ran the pre-commit's and everything is passing but it's not showing it as having been done on the pull request?

Klotske:
> As a sidenote, I've ran the pre-commit's and everything is passing but it's not showing it as having been done on the pull request?

As I understand, running pre-commit locally is a way of testing your changes against different linters and format fixers. If it passed all is good.
But Github has pre-commit check too. It's not showing as "done" because first-time Contributors need a Collaborator's approval for running GH actions workflow.

callum-gander:
> 

Ahhh I thought it might have something to do with that, perfect, thanks!

callum-gander:
> As I understand, running pre-commit locally is a way of testing your changes against different linters and format fixers. If it passed all is good. But Github has pre-commit check too. It's not showing as "done" because first-time Contributors need a Collaborator's approval for running GH actions workflow.

Ahhh I thought it might have something to do with that, perfect, thanks!

website/src/pages/auth/signin.tsx
```diff
@@ -2,17 +2,60 @@ import { Button, Input, Stack } from "@chakra-ui/react";
...
+    if (router?.query?.error) {
+      if (typeof router.query.error === "string") {
+        const errorType = errors[router.query.error];
+        setError(errorType);
+      } else {
+        const errorType = errors[router.query.error[0]];
+        setError(errorType);
+      }
+    }
+  }, [router]);
```
AbdBarho, line 57: you could simplify this by extracting the error into its own variable.
```ts
useEffect(() => {
    const err = router?.query?.error;
    if (err) {
      if (typeof err === "string") {
        setError(errors[err]);
      } else {
        setError(errors[err[0]]);
      }
    }
  }, [router]);
```

website/src/pages/auth/signin.tsx
```diff
@@ -2,17 +2,60 @@ import { Button, Input, Stack } from "@chakra-ui/react";
...
+  | "OAuthCreateAccount"
+  | "EmailCreateAccount"
+  | "Callback"
+  | "OAuthAccountNotLinked"
+  | "EmailSignin"
+  | "CredentialsSignin"
+  | "SessionRequired"
+  | "default";
+
+const errors: Record<SignInErrorTypes, string> = {
```
AbdBarho, line 26: could you please rename to error messages or similar?

AbdBarho (commented):
Thanks!

website/src/pages/auth/signin.tsx
callum-gander: Sorry, which bit specifically were you referring to SignInErrorTypes or errors?

website/src/pages/auth/signin.tsx
callum-gander: True, that'd be much better, thank you. I'll change this when you respond to the previous comment I left

website/src/pages/auth/signin.tsx
AbdBarho: errors

AbdBarho (approved):
Thanks!

PR #546, opened 2023-01-08
CodeBugger added some disclaimer on the pip install and added a new notebook with big copypasta no need to install

furlat:


yk:
for the copypasta code, if I understand correctly, that's just the openbugger source code pasted into a notebook? wouldn't it make more sense to actually check in the library, for example at `openassistant/utils/` or under `scripts/` somewhere? your notebook could then import that, and you could retain an MIT header to indicate that it originally comes from your third-party library.

furlat:
I totally can do it that way! I just wanted to unlock #460, author there should be able to run so no urgency, I will add a few checks and tests and then commit where you thinks is more suitable ad the end of the week (Monday is here T_T )

53jk1 (approved):

Issue #514, opened 2023-01-08
Web Meeting 2023-01-10 (??)


fozziethebeat:
Agenda Items:
- Admin MVP
  * Any questions on what's needed before MVP deadline?
  * Discuss future changes to design
- Deployment
  * Getting the dev box working with secrets
- Misc high priority fixes
- Acking tasks needs discussion
- `/tasks/all`, hidden? only admins?
- Discussion
  * Sharing tasks
  * Communicating gamification nudges

othrayte:
As a lower priority meeting topic, the e2e tests of the website are currently limited because it is hard to check that the results successfully reached the backend, does anyone have any ideas how to improve this situation?

AbdBarho:
@othrayte by reached the backend you mean processed by the python backend or actually writting into the database?

for the former you can intercept requests and check that for the response's `200 <= status code < 300`

for the later, I would argue this is not part of the e2e test, the backend tests should cover this. Although I understand that "it does not hurt check".

othrayte:
@AbdBarho part of the goal of the e2e tests we added was to act as a smoke test that neither the website or the backend have changed in a way that causes the actual functionality to be broken, we have contract tests for the interaction but the goal of the e2e tests would be to show "if a user does {thing we expect} then {result we expect} happens", for some of our tasks like logging in that is easy to see that it worked, for something like the admin panel we could quite easily write a test that does an action and verifies that the result (which intentionally relies on the backend functioning) happens correctly. The issue is with the e2e tests for the tasks. From both the e2e test and users point of view there isn't any way to know that it worked, it kind of just goes into a black hole. It would be nice to show that it was recorded successfully and could be retrieved, but we don't have and API to check the result as we don't otherwise need one. All this said I just wanted to prompt for ideas, if the best we can do is show that the backend would receive it then I suppose we can find a way to do that. It will just be a shame as it doesn't prove that e2e (frontend to backend and back again) that the system still works which would be the most valuable smoke test.

fozziethebeat:
My Meeting Notes:
- Current Status:
  * Able to get all the tasks
  * Can label messages in tasks
  * Can view messages
  * Can login via Discord & Email 
  * Has minimal Admin page for web users
- Still need to do
  * Present `random` task
  * Given Andreas and Yannic admin access
  * Fetch valid flagging labels for message views
  * Handle peer review task: have user check `spam` label first and then reveal other labels.
[...]

PR #549, opened 2023-01-08
Add /tasks/all route to website

kostiak:
Added a route for all tasks in `/tasks/all` in preperation for adding `/tasks/random`.

Did not add a link to the page anywhere.

Note: We will need to later decide how to handle the page. Should we add it to a feature flag? Give access to "advanced users" or something?

In any case, this will be useful for dev once we move over to the random tasks.


AbdBarho (approved):
we can decide in #514 on the visibility of this page

PR #550, opened 2023-01-08
Use hooks for "create" tasks

AbdBarho:
clean up code.

The same should be done for evaluate tasks.

### DIFF
--- a/website/src/hooks/tasks/useGenericTaskAPI.tsx
+++ b/website/src/hooks/tasks/useGenericTaskAPI.tsx
@@ -23,6 +23,7 @@ export const useGenericTaskAPI = <TaskType,>(taskApiEndpoint: string) => {
     {
       onSuccess: (data) => setTasks([data]),
       revalidateOnMount: true,
+      dedupingInterval: 500,
     }
   );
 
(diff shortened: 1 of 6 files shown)


website/src/hooks/tasks/useGenericTaskAPI.tsx
```diff
@@ -26,10 +26,10 @@ export const useGenericTaskAPI = <TaskType,>(taskApiEndpoint: string) => {
   );
 
   useEffect(() => {
-    if (tasks.length === 0 && !isLoading && !error) {
+    if (data === undefined && !isLoading && !error) {
```
Klotske, line 29: Seems that change `undefined` broke fix #393. 
demo

AbdBarho: oh, how to reproduce?

AbdBarho: should be fixed now, could you please check again?

Klotske: Tested. Fix is working!

Klotske: Hmm. It's working 90% of the time. But when rapidly navigating it seems broken.
output3

Klotske: Maybe we should leave the check as it was? And wait for #551 to solve it?

AbdBarho: I reduced the dedupe time to half a second, which should be reasonable.

#551 might solve this entirely, I don't know if we should squeeze it before MVP.

fozziethebeat: Let's get the random task change done then we can switch to `getServerSideProps`.  That can be rolled out anytime.

fozziethebeat (approved):

PR #553, opened 2023-01-08
Store Message Toxicity in database

Nil-Andreu:
Implementing the calculation of the message toxicity in the workflow as well as storing its value in the database.


backend/oasst_backend/utils/hugging_face.py
```diff
@@ -1,10 +1,22 @@
...
 
 import aiohttp
+from loguru import logger
 from oasst_backend.config import settings
+from oasst_backend.models import Message
+from oasst_backend.prompt_repository import PromptRepository
 from oasst_shared.exceptions import OasstError, OasstErrorCode
 
 
+class HF_url(str, Enum):
```
Nil-Andreu, line 12: Follow the camelcase convention used for saving the embeddings.

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -49,3 +61,24 @@ async def post(self, input: str) -> Any:
...
+    pr: PromptRepository,
+    new_message: Message,
+):
+    try:
+        model_name = HF_model.TOXIC_ROBERTA.value
+        hugging_face_api = HuggingFaceAPI(f"{HF_url.HUGGINGFACE_TOXIC_ROBERTA.value}/{model_name}")
+
+        toxicity = await hugging_face_api.post(interaction.text)
+
+        toxicity_instance = pr.insert_toxicity(message_id=new_message.id, model=model_name, toxicity=toxicity)
```
Nil-Andreu, line 77: Return the toxicity_instance.

backend/oasst_backend/config.py
```diff
@@ -27,6 +27,7 @@ class Settings(BaseSettings):
     )
 
     HUGGING_FACE_API_KEY: str = ""
+    DEBUT_SKIP_TOXICITY_CALCULATION: bool = False
```
Nil-Andreu, line 30: DEBUG*

docker-compose.yaml
```diff
@@ -95,6 +95,7 @@ services:
       - DEBUG_SKIP_API_KEY_CHECK=True
       - DEBUG_USE_SEED_DATA=True
       - MAX_WORKERS=1
+      - DEBUT_SKIP_TOXICITY_CALCULATION=False
```
Nil-Andreu, line 98: Set to True.

andrewm4894:
Cool stuff! Am curious and have some questions:

- Do we have plans to use the toxicity or message embeddings within the app such that we need them right away?
- Do they add much in terms of resource overhead on the backend?
- Do they add any latency or complexity that could affect to user experience and flow?

Mainly I am wondering if/why this needs to happen within the app and not as some sort of regular batch job so we can have more separation of concerns.

I am not super familiar with the backend or anything so asking out of ignorance and curiosity and a little bit as devil's advocate but with the best intentions :)

backend/oasst_backend/api/v1/hugging_face.py
```diff
@@ -25,7 +25,8 @@ async def get_text_toxicity(
         ToxicityClassification: the score of toxicity of the message.
     """
 
-    api_url: str = HfUrl.HUGGINGFACE_TOXIC_ROBERTA.value
+    api_url: str = HfUrl.HUGGINGFACE_TOXIC_ROBERTA.value + "/" + HfModel.TOXIC_ROBERTA.value
```
yk, line 28: I don't think this is correct. You'll end up with the model name twice

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -276,27 +276,31 @@ async def tasks_interaction(
...
                 # here we store the text reply in the database
-                newMessage = pr.store_text_reply(
+                new_message = pr.store_text_reply(
                     text=interaction.text,
                     frontend_message_id=interaction.message_id,
                     user_frontend_message_id=interaction.user_message_id,
                 )
 
+                if not settings.DEBUG_SKIP_TOXICITY_CALCULATION:
+                    save_toxicity(interaction, pr, new_message)
```
yk, line 286: below for the embedding calculation, we implement this in a different way, i.e. we don't have a helper function in `hugging_face`, but delegate to `prompt_repository`. Is there a reason we cannot do it analogously here? I'm not saying one approach is better, but we should strive for consistency.

docker-compose.yaml
```diff
@@ -95,6 +95,7 @@ services:
       - DEBUG_SKIP_API_KEY_CHECK=True
       - DEBUG_USE_SEED_DATA=True
       - MAX_WORKERS=1
+      - DEBUG_SKIP_TOXICITY_CALCULATION=False
```
yk, line 98: set to true by default

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -63,3 +69,24 @@ async def post(self, input: str) -> Any:
...
+
+
+async def save_toxicity(
+    interaction,
+    pr: PromptRepository,
+    new_message: Message,
+):
+    try:
+        model_name = HfModel.TOXIC_ROBERTA.value
+        hugging_face_api = HuggingFaceAPI(f"{HfUrl.HUGGINGFACE_TOXIC_ROBERTA.value}/{model_name}")
```
yk, line 81: Again, I don't think that results in the correct endpoint (at least I get a "not found" when I go there)

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -4,14 +4,20 @@
...
 from oasst_shared.exceptions import OasstError, OasstErrorCode
 
 
 class HfUrl(str, Enum):
-    HUGGINGFACE_TOXIC_ROBERTA = ("https://api-inference.huggingface.co/models/unitary/multilingual-toxic-xlm-roberta",)
+    HUGGINGFACE_TOXIC_ROBERTA = "https://api-inference.huggingface.co/models/unitary/multilingual-toxic-xlm-roberta"
     HUGGINGFACE_FEATURE_EXTRACTION = "https://api-inference.huggingface.co/pipeline/feature-extraction"
 
 
+class HfModel(str, Enum):
```
yk, line 17: In analogy to below, this should probably be called something like `HfClassificationModel`

yk (commented):
hey thanks, I left a few comments

re: self-documenting expressions, hard to believe, 2019 is now 4 years ago :D not so new anymore... https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

backend/oasst_backend/api/v1/tasks.py
Nil-Andreu: Yes, well this helper function in huggingface is at the end using the `prompt_repository`. Just I divided the concerns to make the code more readable.

Nil-Andreu: For the creation of the rest api client of the embeddings calculation, I also refactored that code with this helper function (that used as well the pr underneath).

docker-compose.yaml
Nil-Andreu: Okay thanks! Done

backend/oasst_backend/api/v1/hugging_face.py
Nil-Andreu: Thanks! Yeah needed to separate the model name correctly.

backend/oasst_backend/utils/hugging_face.py
Nil-Andreu: Solved thanks!

backend/oasst_backend/utils/hugging_face.py
Nil-Andreu: Done thanks!

backend/oasst_backend/api/v1/tasks.py
yk: yes, I saw that. the issue is, a function that is called `save_X`, I would expect to take `X` as one of the arguments and then save that. But here, the function is actually mainly responsible for getting X, while the actual saving part is delegated to the prompt repository. so the prompt repository itself is perfectly capable of "saving X" in a single line of code, and I don't think we need additional redirection for that.

the bigger problem is this: the current change is not abstraction, it's just indirection & moving code around, making things more complicated. if anything should be abstracted, it's the actual computation part, but that is as of now simply placed into a different spot, it's not made easier. you can easily see how this design is suboptimal in the new PR you opened (#578 ) because there, you actually have to duplicate the code of building the URL and doing the http request into both your new helper function and the endpoint, because you obviously don't want to save in the endpoint. isn't the purpose of such a helper function to prevent you from having to do exactly that? on top of that, you're now missing error handling in the new PR.

I just don't see the current change as having positive impact. it moves control flow unnecessarily from one file to the next without gaining anything. I'd leave the calling of HF and storing as it is done below for the embedding, that's much clearer. If we want to refactor anything (in a separate PR), we can make a helper to make it easier to call HF endpoints and then use that from both branches

yk:
> * Do we have plans to use the toxicity or message embeddings within the app such that we need them right away?

Not concrete plans, but the idea is that a (trusted) frontend could check dynamically whether some input violates the classifier.

> * Do they add much in terms of resource overhead on the backend?

Not really, beyond an open socket.

> * Do they add any latency or complexity that could affect to user experience and flow?

Maybe, we'll have to see.

> * Any cost considerations with huggingface api in terms of scale and streaming vs batch usage of thier api?

This would only be for real-time inference. I think could still do batch computation for all stored things.

Nil-Andreu:
> > * Do we have plans to use the toxicity or message embeddings within the app such that we need them right away?
> 
> Not concrete plans, but the idea is that a (trusted) frontend could check dynamically whether some input violates the classifier.
> 
> > * Do they add much in terms of resource overhead on the backend?
> 
> Not really, beyond an open socket.
> 
> > * Do they add any latency or complexity that could affect to user experience and flow?
> 
> Maybe, we'll have to see.
> 
> > * Any cost considerations with huggingface api in terms of scale and streaming vs batch usage of thier api?
> 
> This would only be for real-time inference. I think could still do batch computation for all stored things.

Maybe batch processing for the messages that we were not able to obtain the toxicity score. This is something I could work on it after this PR.

Nil-Andreu:
Have made couple final changes, tomorrow will test them and make sure it works correctly.

backend/alembic/versions/2023_01_10_0827-e7f955d6ff1b_.py
```diff
@@ -0,0 +1,21 @@
+"""empty message
```
yk, line 1: this revision is empty and can be deleted

backend/alembic/versions/2023_01_10_0845-2d71d4cd5a1b_.py
```diff
@@ -0,0 +1,21 @@
+"""empty message
```
yk, line 1: this one too

backend/alembic/versions/2023_01_10_2207-33edba7076ff_.py
```diff
@@ -0,0 +1,21 @@
+"""empty message
```
yk, line 1: this one too

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -8,10 +8,14 @@
 
 
 class HfUrl(str, Enum):
-    HUGGINGFACE_TOXIC_ROBERTA = ("https://api-inference.huggingface.co/models/unitary/multilingual-toxic-xlm-roberta",)
+    HUGGINGFACE_TOXIC_ROBERTA = "https://api-inference.huggingface.co/models/unitary"
```
yk, line 11: this string value has nothing to do with "toxic" or "roberta", so it shouldn't be named that

backend/oasst_backend/utils/hugging_face.py
```diff
@@ -8,10 +8,14 @@
...
     HUGGINGFACE_FEATURE_EXTRACTION = "https://api-inference.huggingface.co/pipeline/feature-extraction"
 
 
+class HfClassificationModel(str, Enum):
+    TOXIC_ROBERTA = "multilingual-toxic-xlm-roberta"
```
yk, line 16: if you look at the hub, the model is called "unitary/multilingual-toxic-xlm-roberta", right now the "unitary" is in the URL part above and it should be here.

backend/oasst_backend/prompt_repository.py
```diff
@@ -229,6 +238,29 @@ def store_ranking(self, ranking: protocol_schema.MessageRanking) -> MessageReact
...
+            OasstError: if misses some of the before params
+        Returns:
+            MessageToxicity: the instance in the database of the score saved for that message
+        """
+
+        if None in (message_id, model, toxicity):
+            raise OasstError("Paramters missing to add toxicity", OasstErrorCode.GENERIC_ERROR)
+
+        message_toxicity = MessageToxicity(
+            message_id=message_id, model=model, score=toxicity.score, label=toxicity.label
```
yk, line 257: see now you added `label`, but it's missing in the `None` check above. I'm still quite convinced the `None` check is unnecessary and the pydantic validators will pull

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -276,27 +277,45 @@ async def tasks_interaction(
...
                 )
 
+                if not settings.DEBUG_SKIP_TOXICITY_CALCULATION:
+                    try:
+                        model_name: str = HfClassificationModel.TOXIC_ROBERTA.value
+                        hugging_face_api: HuggingFaceAPI = HuggingFaceAPI(
+                            f"{HfUrl.HUGGINGFACE_TOXIC_ROBERTA.value}/{model_name}"
+                        )
+
+                        toxicity: List[List[ToxicityClassification]] = await hugging_face_api.post(interaction.text)
```
yk, line 293: `post` returns `Any`. casting like this is really suboptimal, and especially to pydantic objects.
have you tested whether all of this works? if yes, we can leave it like this and fix later, but given that http calls return json as dicts, and later in your code you act properties of toxicity, it seems quite dangerous.

yk (commented):
thank you, I've left some more comments

also, I'd just delete all alembic revisions and recreate, because it's a big mess

Nil-Andreu:
Have changed the code based on the feedback. If there needs to be changed anything let me know!

backend/oasst_backend/prompt_repository.py
```diff
@@ -291,6 +292,26 @@ def store_ranking(self, ranking: protocol_schema.MessageRanking) -> Tuple[Messag
 
         return reaction, task
 
+    def insert_toxicity(self, message_id: UUID, model: str, toxicity) -> MessageToxicity:
```
andreaskoepf, line 295: please provide a type annotation for the toxicity parameter, e.g. `dict[str, Any]`

backend/oasst_backend/prompt_repository.py
```diff
@@ -291,6 +292,26 @@ def store_ranking(self, ranking: protocol_schema.MessageRanking) -> Tuple[Messag
...
+        """Save the toxicity score of a new message in the database.
+        Args:
+            message_id (UUID): the identifier of the message we want to save its toxicity score
+            model (str): the model used for creating the toxicity score
+            toxicity (float): the values obtained from the message & model
```
andreaskoepf, line 300: the doc string (float) type does not match how it is used below in line 308, e.g. it is probably a `dict[str, Any]`?

andreaskoepf (changes requested):
Looks good, I requested some minor changes (alembic revision & type annotation).

backend/oasst_backend/prompt_repository.py
andreaskoepf: Since the function seems to expect a single float in `toxicity["label"]` .. it would be better to extract that value outside fo the function and pass simply a float (then it would also match the docstring again ;-) ).

backend/oasst_backend/prompt_repository.py
```diff
@@ -306,9 +327,6 @@ def insert_message_embedding(self, message_id: UUID, model: str, embedding: List
             MessageEmbedding: the instance in the database of the embedding saved for that message
         """
 
-        if None in (message_id, model, embedding):
```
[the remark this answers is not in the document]
yk: this was me. pydantic should take care of this

andreaskoepf (approved):
ready to be merged!

PR #554, opened 2023-01-08
537: Endpoint to list frontend users

### LINKED ISSUE #537: Endpoint to list frontend users
- `GET` `/frontend_users/`: list users ordered by username, `max_count` (`int`, limit), `ge` (`str`, optional, `>=` comparison on username), `lt` (`str`, optional, `<` comparison on username), `auth_method` (`str`, optional)

Vechtomov:
Resolves #537


Vechtomov:
I've left a couple of questions here [https://github.com/LAION-AI/Open-Assistant/issues/537#issuecomment-1374931739](url), don't forget to consider them during review

backend/oasst_backend/api/v1/frontend_users.py
```diff
@@ -6,13 +6,28 @@
...
+    max_count: int = Query(10, gt=0, le=20),  # TODO: refine bounds
+    ge: str = None,
+    lt: str = None,
+    auth_method: str = None,
+    api_client: ApiClient = Depends(deps.get_api_client),
+    db: Session = Depends(deps.get_db),
+):
+    pr = UserRepository(db, api_client)
+    users = pr.query_users(limit=max_count, ge=ge, lt=lt, auth_method=auth_method)
+    return [utils.prepare_user(u) for u in users]
```
Nil-Andreu, line 28: I think that it is better to only import the function that you will use at the end, instead of the whole module.

backend/oasst_backend/user_repository.py
```diff
@@ -62,3 +62,32 @@ def get_user_leaderboard(self, role: str) -> LeaderboardStats:
...
+        users = self.db.query(User)
+
+        if not self.api_client.trusted:
+            users = users.filter(User.api_client_id == self.api_client.id)
+
+        if auth_method:
+            users = users.filter(User.auth_method == auth_method)
+
+        if ge:
+            users = users.filter(User.display_name >= ge)
```
Nil-Andreu, line 83: Which is the display name of a User? I think that is a string, so how is comparing a string greater than ge?

Nil-Andreu (changes requested):

backend/oasst_backend/api/v1/frontend_users.py
yk: importing modules is actually usually the preferred way, and importing individual members is discouraged.
- the entire module code gets executed either way, but
- with module imports, it's always clear where a function comes from

Vechtomov: I agree with @yk, besides `utils` was already used in such way in this module before my changes

backend/oasst_backend/user_repository.py
yk: that is indeed possible and intended here: https://www.educba.com/postgresql-compare-strings/

backend/oasst_backend/api/v1/utils.py
```diff
@@ -39,3 +39,7 @@ def prepare_tree(tree: list[Message], tree_id: UUID) -> protocol.MessageTree:
         tree_messages.append(prepare_message(message))
 
     return protocol.MessageTree(id=tree_id, messages=tree_messages)
+
+
+def prepare_user(u: User) -> protocol.User:
```
yk, line 44: I don't think this is well-named, I'd go with something descriptive like `db_user_to_protocol_user` or you could add a `@staticmethod` to the protocol user class like `from_db_user` to construct one from a db user. Or a `to_protocol_user` method on the db user class that transforms the db user into a protocol user. I think I'd go for the last option, since the second would create a back-dependency from shared to backend, and doesn't result in a separate extra lonely function like here.

backend/oasst_backend/user_repository.py
Vechtomov: Strings can be compared in alphabetical order and PostgreSQL supports string comparison operators. [https://stackoverflow.com/questions/40601324/comparing-strings-in-postgres-using-comparison-operators](url)

backend/oasst_backend/user_repository.py
```diff
@@ -62,3 +62,32 @@ def get_user_leaderboard(self, role: str) -> LeaderboardStats:
...
+
+        if lt:
+            users = users.filter(User.display_name < lt)
+
+        users = users.order_by(User.display_name)
```
yk, line 88: I know the query planner will probably take care of this, but could we do the sorting before the lt and gt comparisons?

backend/oasst_backend/user_repository.py
yk, line 83: this parameter should really be called gte in this case

yk (commented):
thanks for the PR, I left a few comments.

Regarding your questions:
- yes, for now I'd also accept an api_id like the other method. It's kinda ugly but only trusted frontends can submit one anyway
- the endpoints returning a list of users is fine
- I'm really not a fan of the username string comparisons here. as far as I can tell, we don't even have an index to support a prefix query here (although maybe the composite index would do the work). I'd be much more in favor of a classic `limit` and `offset` pagination, instead of the `lt`, `gt`, and `max_count`. @olliestanley is there a particular reason we're doing the prefix queries?

backend/oasst_backend/api/v1/frontend_users.py
Nil-Andreu: Okay nice! Thanks sorry

backend/oasst_backend/user_repository.py
Nil-Andreu: So we are comparing alphabetically? The display_name creation does follow an order when created?

olliestanley:
> * I'm really not a fan of the username string comparisons here. as far as I can tell, we don't even have an index to support a prefix query here (although maybe the composite index would do the work). I'd be much more in favor of a classic `limit` and `offset` pagination, instead of the `lt`, `gt`, and `max_count`. @olliestanley is there a particular reason we're doing the prefix queries?

@andreaskoepf would have to confirm on the reasoning for this, but I assumed it was for cases like wanting to lookup a user by username without knowing auth_method or api_client_id, so you could use these criteria to list all users with a certain username regardless of auth_method, api_client_id

backend/oasst_backend/user_repository.py
Vechtomov: Yes, we are comparing alphabetically. I'm not sure what you mean by following an order when created

yk:
> @andreaskoepf would have to confirm on the reasoning for this, but I assumed it was for cases like wanting to lookup a user by username without knowing auth_method or api_client_id, so you could use these criteria to list all users with a certain username regardless of auth_method, api_client_id

but then let's add a prefix filter (w/ corresponding index). probably the majority of use cases of the endpoint will be listing and paging, and it seems quite cumbersome to do that with the current implementation, especially when we impose some upper limit on the number of users returned (which we have to).

Vechtomov:
Hi, @yk. Thanks for the good review. I've made changes but I still don't understand why we need to pass `api_client_id` for untrusted clients. We pass `api_key` with the request and it uniquely defines `api_client`.

backend/oasst_backend/user_repository.py
Vechtomov: fixed

backend/oasst_backend/user_repository.py
Vechtomov: fixed

backend/oasst_backend/api/v1/utils.py
Vechtomov: fixed

yk:
> Hi, @yk. Thanks for the good review. I've made changes but I still don't understand why we need to pass `api_client_id` for untrusted clients. We pass `api_key` with the request and it uniquely defines `api_client`.

I think in your question, you referred to another endpoint where this is done. The reasoning there was the following: a trusted api key should be able to also see users of other keys, and would pass those keys along, i.e. saying "give me users of this other key". I honestly don't see a super good reason why we might do this, so I'd be comfortable leaving that away and just saying trusted frontends can just read globally.

backend/oasst_backend/user_repository.py
Nil-Andreu: I mean I do not understand which could be the use case in which we might want to compare alphabetically.

Vechtomov: We're also trying to figure it out in the main thread :)

andreaskoepf:
I am bit irritated by the paging discussion here.

I deliberately asked to select by interval-starting point, direction + limit. Naive paging with offset and limit is not scalable. Selecting result elements BY OFFSET (!) means in many cases the db-engine runs a linear search over the results until it reaches the indices of the index-window to return. For larger datasets and high offsets this has very bad performance characteristics.

Regarding indices:  BTREE indices store strings ordered (based on the configured collation) and are always used string comparisions == <= < > >= etc are used .. query engine will also use them for `LIKE "prefix%"` conditions.

andreaskoepf (approved):
It should probably be protocol.FrontEndUser instead of protocol.User .. but I can do the fix quickly. Will merge now.. I think the paging-strategy has to be discussed. To me naive paging is the noob option .. but you guys all seem to love it. We should just add naive index paging everywhere and we will see what happens.

PR #570, opened 2023-01-09
added data_augment.py

finitearth:
Second try of commiting for https://github.com/LAION-AI/Open-Assistant/issues/460


finitearth:
I was able to run pre-commit now, I hope the checks run through properly now

yk (approved):
thank you!

scripts/data_augment/data_augment.py
```diff
@@ -0,0 +1,464 @@
...
+from bs4 import BeautifulSoup as bs
+
+import requests
+import json
+import html
+import os
+import subprocess
+import argparse
+
+
```
andrewm4894, line 23: would you be able to maybe add a comment showing some example usage of the script? or some examples in a README.md in the folder if easier.

scripts/data_augment/data_augment.py
```diff
@@ -0,0 +1,464 @@
+import torch
```
andrewm4894, line 1:
```suggestion
'''What does this script do

example usage: 

args: 
'''

import torch
```

maybe add some small description and example usage and maybe args at top of file so easy for others to just get quick idea of what its doing.

finitearth:
Not sure why it keeps failing... On my command line it shows that all commits ran without fails

yk:
> Not sure why it keeps failing... On my command line it shows that all commits ran without fails

you have to commit and push again after letting it run, as it does auto-fixes and modifies your files

scripts/data_augment/data_augment.py
finitearth: Added the comment, thanks for your feedback

andrewm4894 (approved):

andrewm4894:
@finitearth you should be good to merge this.

PR #573, opened 2023-01-09
replace `andrewm4894` with `LAION-AI` in `example.ipynb`

andrewm4894:
now that example notebook is merged just clean up and replace `andrewm4894` with `LAION-AI` in the readme and config cell at top of example.ipynb.


yk (approved):
thank you

Issue #48, opened 2022-12-21
Try Supervised Fine-Tuning on pseudo-QA-data


yk:
The first step in InstructGPT (https://openai.com/blog/instruction-following/) is supervised fine-tuning on human instruction data. Our website and bot are being created to collect this data. Meanwhile, we can already try out whether and how it's possible to fine-tune LLMs on such data, by substituting the not-yet-collected data for pseudo-data. One idea is to take a QA dataset (like squad or natural questions) and convert this into instruction-response pairs, then running the fine-tuning on top of that to get a feel for the dynamics of training.

ekurtulus:
I can be of help for this.

yk:
Amazing! Maybe it's best if you write down an initial roadmap here, like a step-by-step plan to reach the goal, then we can see better where potential problems might be!

ekurtulus:
Here are the steps I am planning to take.

1. Find a good pretrained model which is not that large (I believe a mid-sized T5 would be a nice choice)
2. I will create two different variants of the SQUAD dataset, one with original labels and one with slightly corrupted labels (maybe back-translation ?)
3. I will train the reward model, basically a linear layer with output dims=2 on top of our pretrained model
4. Then, by using the reward model, I will try training the model with PPO

bth5032:
Are we planning to use an encoder-decoder model like t5 or a decoder only model (or is this irrelevant at this stage)? I'm interested in understanding the choice of T5 vs a huggingface GPT2 implementation or OPT which has large pretrained models already available.

yk:
> Are we planning to use an encoder-decoder model like t5 or a decoder only model (or is this irrelevant at this stage)? I'm interested in understanding the choice of T5 vs a huggingface GPT2 implementation or OPT which has large pretrained models already available.

architecturally, a decoder-only model is simpler, but an encoder-decoder system allows for explicit separation of input and output. ultimately, it will come down to the numbers which one we'd rather use. I feel a bit like decoder-only models shine in their simplicity and versatility, but I have no clue.

ekurtulus:
The reason why I thought starting with T5 would be a good idea is that Flan-T5 outperforms OPT-IML. Also, once we have the training codebase available, we will be able to experiment with whatever architecture we want.

bth5032:
> The reason why I thought starting with T5 would be a good idea is that Flan-T5 outperforms OPT-IML. Also, once we have the training codebase available, we will be able to experiment with whatever architecture we want.

[...]

PR #576, opened 2023-01-09
A more general support  for QA and summarization (review and feedback needed)

ekurtulus:
With this update, I am aiming for generalizing our training framework. I need feedback and review on the structure that I propose. Accordingly, I will make the codebase fully functional and extend it. Additionally, I added more papers to the docs side.

see [this issue](https://github.com/LAION-AI/Open-Assistant/issues/48)

I think we need support not only for QA but also summarization. In this PR, I wanted to extend the existing datasets, add support for seq2seq models like T5, PolyLoss and few other modifications. Normally, I was also planning to add support for [Sharpness Aware Training](https://github.com/LAION-AI/Open-Assistant/issues/48), but it requires overriding the Huggingface Trainer's __inner_training_loop_ function. I can do that but I am sure if it is something we want. If we decide to do so, I can submit another PR.

Furthermore, if the structure of this commit is approved, I can add more tasks, a script for synthetic data generation through data augmentation, and more.

### DIFF
--- a/website/src/lib/auth.ts
+++ b/website/src/lib/auth.ts
@@ -1,5 +1,20 @@
 import type { NextApiRequest, NextApiResponse } from "next";
-import { getToken } from "next-auth/jwt";
+import { getToken, JWT } from "next-auth/jwt";
+
+/**
+ * Wraps any API Route handler and verifies that the user does not have the
+ * specified role.  Returns a 403 if they do, otherwise runs the handler.
+ */
+const withoutRole = (role: string, handler: (arg0: NextApiRequest, arg1: NextApiResponse, arg2: JWT) => void) => {
+  return async (req: NextApiRequest, res: NextApiResponse) => {
+    const token = await getToken({ req });
+    if (!token || token.role === role) {
+      res.status(403).end();
+      return;
+    }
+    return handler(req, res, token);
+  };
+};
 
 /**
  * Wraps any API Route handler and verifies that the user has the appropriate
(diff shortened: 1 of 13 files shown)


yk:
you have some merge conflict markers in your files

sanagno:
Looks good, after the merge! Do you know which models use the SAM optimizer?

docs/docs/research/general.md
```diff
@@ -2,6 +2,43 @@
 
 This page lists research papers that are relevant to the project.
 
+<<<<<<< HEAD
```
andrewm4894, line 5: i think this means you might need to update from upstream/master

ekurtulus:
> Looks good, after the merge! Do you know which models use the SAM optimizer?

None. It is an optimizer you use in conjunction with other optimizers. Here is the [link](https://arxiv.org/pdf/2010.01412) to the paper. There are papers claiming it to be increasing generalization like [this one](https://arxiv.org/pdf/2110.08529.pdf).

sanagno:
Looks great, If you can merge the final conflicts and add a default accuracy metric as before would be great!

ekurtulus:
> @ekurtulus did you close this on purpose?

Yes, I realized that I was not using pre-commit files and not branching properly. While trying to set everything up properly, I had to delete and re-fork the main repository. So, it was closed automatically. I will submit a new PR.

ekurtulus:
Please see [PR 619](https://github.com/LAION-AI/Open-Assistant/pull/619).

PR #578, opened 2023-01-09
[NEW] Create message embedding API endpoint

Nil-Andreu:


yk:
Hey, thanks a lot. I think there's a bit of confusion what's to do here.
Generally, an "API client" is different from an "API endpoint". What you wrote here is an "API Endpoint", meaning we provide this as a service in our API to the front-end. We don't want to do this for embeddings, i.e. there is no reason a frontend should want direct text embedding from us. I see you've already also called this "API client" in your PR for the toxicity endpoint, and the naming was already wrong there, but for toxicity it makes sense that the frontend can dynamically check for toxicity. For embeddings it makes no sense, so an endpoint is not needed.

In general, an "API client" is just a python module (or function), whose job it is to make calling an external API (like huggingface) easier.
In general, an API client does the following:
- take inputs and bring them into the correct format for the external endpoint to be called
- figure out the correct URL to call, plus headers, etc.
- do the http call, optionally with retries, etc.
- take the http response and parse out the results and bring them into the correct format for intended use, and return them to the caller
- take care of error handling of the above and transforming errors into appropriate error classes / re-raising /etc.

So I don't think we'll merge this PR as is. What we *could* do is an API client that calls the HF API, but given how absolutely simple that API is, I also don't really see a need for this.

Nil-Andreu:
> Hey, thanks a lot. I think there's a bit of confusion what's to do here. Generally, an "API client" is different from an "API endpoint". What you wrote here is an "API Endpoint", meaning we provide this as a service in our API to the front-end. We don't want to do this for embeddings, i.e. there is no reason a frontend should want direct text embedding from us. I see you've already also called this "API client" in your PR for the toxicity endpoint, and the naming was already wrong there, but for toxicity it makes sense that the frontend can dynamically check for toxicity. For embeddings it makes no sense, so an endpoint is not needed.
> 
> In general, an "API client" is just a python module (or function), whose job it is to make calling an external API (like huggingface) easier. In general, an API client does the following:
> 
> * take inputs and bring them into the correct format for the external endpoint to be called
> * figure out the correct URL to call, plus headers, etc.
> * do the http call, optionally with retries, etc.
> * take the http response and parse out the results and bring them into the correct format for intended use, and return them to the caller
> * take care of error handling of the above and transforming errors into appropriate error classes / re-raising /etc.
> 
> So I don't think we'll merge this PR as is. What we _could_ do is an API client that calls the HF API, but given how absolutely simple that API is, I also don't really see a need for this.

Yep makes completely sense! Thanks a lot for the explanation!

53jk1 (approved):

PR #581, opened 2023-01-09
Documentation FAQ

Nil-Andreu:
Adding in documentation Frequently Asked Questions

### DIFF
--- a/docs/docs/faq/faq.md
+++ b/docs/docs/faq/faq.md
@@ -1,15 +1,13 @@
 ### Docker-Compose instead of Docker Compose
 
-If you have installed the version of `docker-compose` instead of
-`docker compose`, you might have to do some changes in the commands:
+If you are using `docker-compose` instead of `docker compose` (note the " "
+instead of the "-"), you should update your docker cli to the latest version.
+`docker compose` is the most recent version and should be used instead of
+`docker-compose`
 
-```bash
-    # docker compose up backend-dev --build --attach-dependencies
-    docker-compose up --build --attach-dependencies backend-dev     # starting only backend-dev
-
-    # docker compose up --build
-    docker-compose up --build                                       # starting the all services
-```
+For more details and information check out
+[this SO thread](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose)
+that explains it all in detail.
 
 ### Pre-commit
 
--- a/docs/docs/faq/faq.md
+++ b/docs/docs/faq/faq.md
@@ -42,7 +42,7 @@ Instead of running docker with the root command always, you could create a
     # Add the actual user to the group
     sudo usermod -aG docker $USER
 
-    # Log in the group
+    # Log in the group (apply the group changes to actual terminal session)
     newgrp docker
 ```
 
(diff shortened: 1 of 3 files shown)


andrewm4894:
Cool - i helped with the docs site so can help here. 

I think just would need to also be a small change to add a new category in here: https://github.com/Nil-Andreu/Open-Assistant/blob/documentation_faq/docs/sidebars.js

something like....oh i see you doing it as i type :)

andrewm4894:
(will wait until not draft :) - feel free to @ me if i can help at all)

Nil-Andreu:
> Cool - i helped with the docs site so can help here.
> 
> I think just would need to also be a small change to add a new category in here: https://github.com/Nil-Andreu/Open-Assistant/blob/documentation_faq/docs/sidebars.js
> 
> something like....oh i see you doing it as i type :)

Thanks for the feedback!

Nil-Andreu:
> (will wait until not draft :) - feel free to @ me if i can help at al
Thanks! I was wondering for the error of permission denied, if it could be due to the docker version. So to include it in the solution as well (not only restarting docker or deleting all the volumes).

Do you have any advice on that?

docs/docs/faq/faq.md
```diff
@@ -0,0 +1,67 @@
...
+If you have installed the version of `docker-compose` instead of
+`docker compose`, you might have to do some changes in the commands:
+
+```bash
+    # docker compose up backend-dev --build --attach-dependencies
+    docker-compose up --build --attach-dependencies backend-dev     # starting only backend-dev
+
+    # docker compose up --build
+    docker-compose up --build                                       # starting the all services
+```
```
andrewm4894, line 12:
```suggestion
If you are using `docker-compose` instead of `docker compose` (note the " " instead of the "-"), you should update your docker cli to the latest version. `docker compose` is the most recent version and should be used instead of `docker-compose`

For more details and information check out [this SO thread](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose) that explains it all in detail.
```

andrewm4894: @Nil-Andreu i think the advice here is to just update the docker version to latest. Some people have gotten `docker-compose` to work but just easier (hopefully) for everyone if users can just use `docker compose` instead.

andrewm4894:
@Nil-Andreu maybe let's get this merged and we can add and iterate on it bit by bit. Once you happy i can try it out locally myself and if all good i can approve.

docs/docs/faq/faq.md
Nil-Andreu: Okay completely fine for me thanks!

Nil-Andreu:
> @Nil-Andreu maybe let's get this merged and we can add and iterate on it bit by bit. Once you happy i can try it out locally myself and if all good i can approve.

Okay! I will test it locally tomorrow but this could be the first iteration.

Nil-Andreu:
Are we waiting to merge this one for something?

andrewm4894:
> Are we waiting to merge this one for something?

I was waiting for you to merge :)

andrewm4894 (approved):

Nil-Andreu:
> > Are we waiting to merge this one for something?
> 
> I was waiting for you to merge :)
> 
> Oh wait maybe I didn't approve, my bad

Ah okay no worries!!

PR #591, opened 2023-01-10
website: Refactor remaining task pages to use Task.tsx

othrayte:
This changes the labeling tasks to use Task.tsx to reduce duplication and differing implementations.

Note: I had some issues with submitting the labeling tasks as I we get an error and don't handle it very well. I'm still trying to verify if that is something I added or was already there.


fozziethebeat:
I see Yannic is fast on the labeling effort.  Will review once the checks complete.

fozziethebeat (approved):
I don't *see* anything that would introduce a new submission error.  I can try to replicate said error later.

fozziethebeat:
Verified submission failure is not from these changes.

PR #593, opened 2023-01-10
Encapsulate valid label fetching to `FlaggableElement`

### LINKED ISSUE #584: Web Cleanup: Ensure `FlaggableElement` always has a list of valid labels
`FlaggableElement` now wraps any message.  These messages get used throughout the frontend in both tasks and when displaying messages or conversations.

To make things work I've piped an array of valid labels where possible for all task types and an empty array for the message views.

A more scalable approach would be fore `FlaggableElement` to automatically fetch the valid labels itself (and thus simplify all places that use `FlaggableElement`) but using smart caching so that the call only happens the first time `FlaggableElement` is used somewhere in the app.

fozziethebeat:
Closes #584 

This creates a new api route `/api/valid_labels` that returns the list of valid labels that can apply to any message.

`FlaggableElement` is updated to dynamically fetch labels from the new route and cache them.  This allows `FlaggableElement` to be used anywhere without needing to pipe in the valid labels from call sites.  This still calls the api route once per page load.


fozziethebeat:
Blocking on #590 before marking ready for review so I can merge.

othrayte:
I see code 422 (unprocessable entity) responses when submitting the label report on the rank pages, is this expected? Also I get the same on the labeling pages (but I'm not sure is we are planning to have the flags there).

fozziethebeat:
> 

Good catch, I didn't catch that.  Fixed now

fozziethebeat:
Posted #614 as a future cleanup around messages.  The types are pretty confusing and different depending on use case.

website/src/pages/api/valid_labels.ts
```diff
@@ -0,0 +1,23 @@
+import { getToken } from "next-auth/jwt";
+import { oasstApiClient } from "src/lib/oasst_api_client";
+
+/**
+ * TODO
```
AbdBarho, line 5: todo what?

website/src/pages/api/valid_labels.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+const handler = async (req, res) => {
```
AbdBarho, line 7: can we use one of the middlewares here?

website/src/pages/api/valid_labels.ts
fozziethebeat: I see you did the review carefully.  TODO was to write the handler comments.  done :)

website/src/pages/api/valid_labels.ts
fozziethebeat: Done

AbdBarho (approved):

Issue #569, opened 2023-01-09
Refactor remaining task pages to use Task.tsx


othrayte:
The main create and evaluate task pages use a central Task.tsx component but don't of the others don't and this leads to some duplications and inconsistencies. This task aims to unify theremaining task pages to also use the common component. (As discussed in discord)

In addition, as all tasks need the task controls that component should probably be added and setup in Task.tsx rather than in each task type.

othrayte:
I plan on starting on this tomorrow.

AbdBarho:
I have did some clean up of the types in #575

othrayte:
Awesome, thanks for letting me know.

AbdBarho:
can this be closed?

https://github.com/LAION-AI/Open-Assistant/blob/5d1b03b02dce0f87d9ae6f9322c249be5905dfd6/website/src/components/Tasks/Task.tsx#L77-L86

PR #598, opened 2023-01-10
website: Unify task controls

othrayte:
Unify the TaskControls implementation into one place.

Also happens to fix some issues with the no-changes are-you-sure dialog only appearing once and other form values not being reset correctly for new tasks etc.

This one ended up touching more code that I was originally hoping for, but I think it at least leaves everything in a better state.

This is the second PR for #569, if there is a third that would be to change the summarisation task pages to use Task.tsx.


fozziethebeat (approved):
🚢

website/src/components/Tasks/Task.tsx
```diff
@@ -18,51 +37,85 @@ export const Task = ({ tasks, trigger, mutate }) => {
...
+  const submitResponse = () => {
+    switch (taskStatus) {
+      case "NOT_SUBMITTABLE":
+        return;
+      case "DEFAULT":
+        setShowUnchangedWarning(true);
+        break;
+      case "SUBMITABLE": {
+        trigger({
+          id: task.id,
```
fozziethebeat, line 69: This breaks things.  task.id isn't the frontend ID it's the backend ID.  Right now we can't do a guaranteed lookup to find the proper backend ID. 

Once the backend changes remove the frontend ID this will be correct but as of right now submissions are broken.

othrayte: Ah, right. I can look into fixing it in about 5hrs, or do you want to fix it before then?

fozziethebeat: I've got my fix up and ready :)

PR #601, opened 2023-01-10
538: Add endpoints to manage users

### LINKED ISSUE #538: Endpoints to resolve/modify user by global user_id
Untrusted (normal) api_clients must only resolve, view and edit users that were registered by themselves. Trusted api_clients can manage all users. Returned user model contains global OA user id (UUID) and the api_client's 'local' username & auth_method.

- `GET` `/users/{id}`: user by global OA user_id (UUID)
- `PUT` `/users/{id}`: user by user-id (UUID), send json body with attributes to modify (currently only `enabled` bool is supported, setting to false bans user)
- `DELETE` `/users/{id}`: user by user-id (UUID)
- Add `notes` string to the `user` table for admin to note why user was banned

olliestanley:
Resolves #538


andreaskoepf (approved):
Nice, thanks!

PR #604, opened 2023-01-10
Add note on Python version to backend README.md

olliestanley:
We have had a couple of GitHub issues created by people reporting issues with running the backend which seem to be a result of running with Python versions older than 3.10. This PR modifies the backend README.md to explicitly state the Python 3.10 requirement to hopefully avoid anyone else running into this issue.


yk (approved):
thank you!

Issue #84, opened 2022-12-27
Write instructions and examples for unit-testing python code (backend and bot)


yk:
use pytest for this

note: this issue concerns unit tests, meaning if e.g. a database is required, it should be mocked

devforfu:
Hi @yk, that sounds like something I could look into. I worked with `pytest` and did some testing that required DB/API mocking. If no one else is interested, I might've taken a look. (Though I would need some time to start, like fork/set up the project, read contribution guidelines, etc.)

Also, I guess that would help me learn more about the codebase, as I'm pretty interested in building an open AI assistant, and wish to contribute something.

NickDiSanto:
@devforfu Hi Ilia, I'm in the same boat as you regarding this project. This seems like a good task to get started with. If you're interested, let's connect by email or Discord to work together on setting up the project and getting started.
Email: nick.c.disanto@gmail.com
Discord: Nick DiSanto#4045

yk:
Hey @devforfu @NickDiSanto thanks to both of you for speaking up :) very cool! I'll assign both of you and leave it up to you how to organize it. Remember, this is not about writing all the tests, but about giving people a bit of a framework, some instructions, and some examples on how tests are done.

devforfu:
@yk Sounds good! Yes, that makes sense. I guess that would be helpful to establish some testing foundation.

@NickDiSanto That's great! Yes, let's connect via Discord. My tag there is `Ilia#9789`.

jack-michaud:
@devforfu & @NickDiSanto , I started working on #205, we should connect -- lomz#2555

danielpatrickhug:
Langchain has a pretty thorough set of unit and integration tests for a very similar problem.
https://github.com/hwchase17/langchain/tree/master/tests/unit_tests
there are some pretty good examples for database testing too.

jack-michaud:
I'm starting to look at adding tests for the discord bot. The critical flow I'm looking to test is _handle_task (found [here](https://github.com/jack-michaud/Open-Assistant/blob/4e6da8a6f66c6e3860a73d7139c3420d2368ee6d/discord-bot/bot/extensions/work.py#L46)) 

To  test _handle_task , I'm finding we'd either need to **A**) do a lot of mocking of Discord methods or **B**) encapsulate interactions with Discord inside of a TaskCoordinator class, something that would handle sending messages, receiving replies, etc. 

[...]

PR #605, opened 2023-01-10
[WIP] Setting up scaffolding for backend tests

devforfu:
Establishes a setup for testing described in #84.

The purpose of this PR is to enable users for writing unit tests for the backend and, probably, discord-bot components of the Open-Assistant repo.

So far, it only has a trivial test that runs against a single API endpoint. (The app is instantiated in debug mode via the `.env` file.) But I decided to open a draft PR in order to keep the process transparent and communicate to other contributors. Also, to see if I'm on the right track with this issue.

cc @jack-michaud @andreaskoepf


backend/oasst_backend/api/v1/tasks.py
```diff
@@ -25,22 +50,17 @@
...
-    api_client = deps.api_auth(api_key, db)
-
     try:
-        pr = PromptRepository(db, api_client, client_user=request.user)
-        pr.ensure_user_is_enabled()
-
-        tm = TreeManager(db, pr)
-        task, message_tree_id, parent_message_id = tm.next_task(desired_task_type=request.type, lang=request.lang)
-        pr.task_repository.store_task(task, message_tree_id, parent_message_id, request.collective)
+        ctx: TaskContext = task_context_factory(request.user)
```
jack-michaud, line 60: It's a shame `.user` is not common between all requests. This `task_context_factory` is workable but a little more difficult than if it were to just be injected automatically and be user "aware". Just musing

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -9,12 +10,36 @@
...
+from oasst_shared.schemas.protocol import User
 from sqlmodel import Session
 from starlette.status import HTTP_204_NO_CONTENT
 
 router = APIRouter()
 
 
+@dataclass
+class TaskContext:  # todo: come up with a better name?
+    pr: PromptRepository
```
jack-michaud, line 22: Can you spell this out completely?
```suggestion
    prompt_repository: PromptRepository
```

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -25,22 +50,17 @@
...
-        pr = PromptRepository(db, api_client, client_user=request.user)
-        pr.ensure_user_is_enabled()
-
-        tm = TreeManager(db, pr)
-        task, message_tree_id, parent_message_id = tm.next_task(desired_task_type=request.type, lang=request.lang)
-        pr.task_repository.store_task(task, message_tree_id, parent_message_id, request.collective)
+        ctx: TaskContext = task_context_factory(request.user)
+        ctx.pr.ensure_user_is_enabled()
+        task, message_tree_id, parent_message_id = ctx.tm.next_task(desired_task_type=request.type, lang=request.lang)
+        ctx.pr.task_repository.store_task(task, message_tree_id, parent_message_id, request.collective)
```
jack-michaud, line 63: Should `task_repository` be a member on TaskContext?

backend/oasst_backend/api/v1/text_labels.py
```diff
@@ -11,23 +13,30 @@
 router = APIRouter()
 
 
+def prompt_repository_factory(
```
jack-michaud, line 16: I'd recommend moving these factories into a shared place; `api/deps.py` would work

backend/tests/unit/oasst_backend/api/v1/test_tasks.py
```diff
@@ -0,0 +1,69 @@
...
+
+class FakeApp:
+    def __init__(self, app: FastAPI):
+        self.app = app
+
+        fake = mock.Mock(spec=TaskContext)
+        fake.pr = mock.Mock(spec=PromptRepository)
+        fake.tm = mock.Mock(spec=TreeManager)
+        fake.tm.next_task.return_value = TaskDone(), "fake_tree_id", "fake_parent_message_id"
+        fake.pr.task_repository = mock.Mock(spec=TaskRepository)
```
jack-michaud, line 24: Be careful about using `spec` with `return_value`... even though you're using spec, you can still set `return_value` to whatever you want. We should also have integration tests, so we can catch this sort of issue

jack-michaud (commented):
It looks like there are two approaches for injecting dependencies in this PR: building a context object and injecting that (`TaskContext`), and injecting one dependency (`PromptRepository`). 

I'd recommend picking one before moving forward

53jk1 (approved):

backend/oasst_backend/api/v1/tasks.py
devforfu: Yeah, I also wanted to inject the object but had to go with this indirect factory because of the user.

devforfu:
I guess it was somehow automatically closed because of rebasing/merging the fork 🤔 

I made one more attempt to restructure the dependencies a bit and try injecting them without a single "context" class including all possible repositories at once. Also, I had to move `FrontendUserId` into a separate file to avoid circular imports. Plus, I deleted one test as I am not sure how to deal with a nested `@managed_tx_function`.

backend/tests/unit/oasst_backend/api/v1/test_tasks.py
devforfu: Removed specs, and restructured the test a bit.

backend/oasst_backend/api/v1/text_labels.py
devforfu: Done!

backend/oasst_backend/api/v1/tasks.py
devforfu: Removed the context class, but `TreeManager` still uses a short `pr` property. Do you think it makes sense to update that one instead?

backend/oasst_backend/api/v1/tasks.py
devforfu: Updated `deps.py` a bit, so should be somewhat decoupled from each other.

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -9,12 +10,36 @@
...
 from sqlmodel import Session
 from starlette.status import HTTP_204_NO_CONTENT
 
 router = APIRouter()
 
 
+@dataclass
+class TaskContext:  # todo: come up with a better name?
+    pr: PromptRepository
+    tm: TreeManager
```
ArgiesDario: same as "pr", as defined in [PEP 8: Function and Variable names](https://peps.python.org/pep-0008/#function-and-variable-names), it should contain the whole name with underscores

pr should be prompt_repository
tm should be tree_manager

backend/tests/unit/oasst_backend/README.md
```diff
@@ -0,0 +1,5 @@
+# Backend Testing Guidelines
+
+## Testing Endpoints
+
+TBD
```
ArgiesDario, line 5: If possible could you provide some basic doc so others ones have a starting point? ty :)

PR #606, opened 2023-01-10
Pin python package versions

HyperGH:
Pinned packages to current latest versions to ensure stability. Also removed a duplicate dependency declaration for hikari that was possibly erroneous. It may be a good idea to set up dependabot or a similar utility to monitor and update these dependencies, but that is outside the scope of this PR.


AlexanderHOtt (approved):
looks good. Adding dependabot would be a good idea after we hit MVP

yk (approved):
thank you!

PR #609, opened 2023-01-11
85: Unit-testing nextjs & react code

ericv105:


fozziethebeat:
Will review after one of my big PRs that fixes things goes in.

website/cypress/contract/oasst_api_contract_tests.cy.ts
```diff
@@ -1,4 +1,4 @@
-import { OasstApiClient, OasstError } from "src/lib/oasst_api_client";
+import { OasstApiClient, OasstError } from "../../src/lib/oasst_api_client";
```
fozziethebeat, line 1: I'd really like to avoid relative imports.  What requires this relative import here?

website/cypress/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "types": ["cypress", "node"],
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "jsx": "react-jsx"
```
fozziethebeat, line 8: The regular `tsconfig.json` has `baseUrl: "."` could that work here too?

fozziethebeat (changes requested):

website/cypress/tsconfig.json
ericv105: Adding `baseUrl: ".."` worked to avoid relative imports for the contract and the component test.

PR #613, opened 2023-01-11
Add realistic seed data json and update seed_data to properly label task payloads

danielpatrickhug:
1. Adds a json of about 450 messages from anthropic's helpful test dataset.
2. Update seed_data to properly label task payloads as assistant and prompter.


yk (approved):
awesome, thank you!

PR #619, opened 2023-01-11
Continuing #576

ekurtulus:
Please [see](https://github.com/LAION-AI/Open-Assistant/pull/576).

### DIFF
--- a/model/supervised_finetuning/trainer.py
+++ b/model/supervised_finetuning/trainer.py
@@ -3,21 +3,22 @@ import os
 from distutils.util import strtobool
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-import bitsandbytes
 import torch
 from torch import nn
 from transformers import PreTrainedModel, Trainer, TrainingArguments
-from transformers.training_args import OptimizerNames
-from utils import get_dataset, get_loss, get_model, get_tokenizer, read_yamls
+from utils import get_dataset, get_loss, get_metrics, get_model, get_tokenizer, read_yamls
 
 os.environ["WANDB_PROJECT"] = "supervised-finetuning"
 
 
-def compute_metrics(eval_pred):
-    pred_ids = eval_pred.predictions
-    labels = eval_pred.label_ids
+def compute_metrics(eval_pred, preprocess_fn, metrics):
+    preds, labels = preprocess_fn(eval_pred)
 
-    return {"accuracy": (pred_ids[labels > 0] == labels[labels > 0]).mean()}
+    out = {}
+    for metric in metrics:
+        out = dict(**out, **metric.compute(predictions=preds, references=labels))
+
+    return out
 
 
 def preprocess_logits_for_metrics(logits, labels):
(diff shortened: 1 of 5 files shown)


model/supervised_finetuning/trainer.py
```diff
@@ -92,6 +94,8 @@ def argument_parsing(notebook=False, notebook_args=None):
     parser.add_argument("--local_rank", type=int, default=-1)
     parser.add_argument("--deepspeed", action="store_true")
     parser.add_argument("--no-deepspeed", dest="deepspeed", action="store_false")
+    parser.add_argument("--poly_eps", type=float, default=1.0)
+    parser.add_argument("--seq2seq_model", action="store_true")
```
sanagno, line 98: I would add these to the config files

model/supervised_finetuning/utils.py
```diff
@@ -32,8 +36,51 @@ def get_tokenizer(conf):
...
+    qa_datasets = ["squad_v2", "adversarial_qa", "trivia_qa_context", "trivia_qa_noconext"]
+    summarization_datasets = ["xsum", "cnn_dailymail", "samsum", "multi_news"]
+
+    # the reason behind using a list is that we might want to extend the list of our
+    # metrics in the future for more thorough evaluation
+    if any(dataset in qa_datasets for dataset in conf.datasets):
+        metrics, preprocess_fn = [evaluate.load("squad_v2")], preprocess_qa
+    elif any(dataset in summarization_datasets for dataset in conf.datasets):
+        metrics, preprocess_fn = [evaluate.load("rouge")], partial(
+            preprocess_summarization, tokenizer, ignore_pad_token_for_loss=conf.ignore_pad_token_for_loss
```
sanagno, line 75: Is ignore_pad_token_for_loss defined in the config?

model/supervised_finetuning/trainer.py
ekurtulus: > I would add these to the config files

Yes, normally it was but because of the file problems I had, it seems like that part was not updated.

model/supervised_finetuning/utils.py
ekurtulus: > Is ignore_pad_token_for_loss defined in the config?

It is one of the default arguments of huggingface trainer, so I thought it would be in the config.

model/supervised_finetuning/trainer.py
```diff
@@ -131,23 +135,14 @@ def argument_parsing(notebook=False, notebook_args=None):
...
-
-    optimizer = OptimizerNames.ADAMW_BNB if training_conf.quantization else None
-
-    if training_conf.quantization:
-        for module in model.modules():
-            if isinstance(module, torch.nn.Embedding):
-                bitsandbytes.optim.GlobalOptimManager.get_instance().register_module_override(
-                    module, "weight", {"optim_bits": 32}
-                )
+    metrics, preprocess_fn = get_metrics(training_conf)
```
sanagno, line 138: I am not sure these are used.

Issue #126, opened 2022-12-28
Training data from Twitter


dhruv2601:
Twitter is a good source for gathering multi-turn conversation training data. One approach is to convert any thread into 
`A:  B:   A: ` format by linking user_id to the tweets. It has worked well in my experience.

I can prepare a plan, notebooks and share sample results(a few 1000 for a start) on what the data would look like, and further access quality over toxicity metrics on random samples.
The data can be taken freely from [twitterstream archives](https://archive.org/details/twitterstream).

@yk what do you think?

degenai:
I would love to see this plan as someone trying to educate themselves.

tusharagarwal25:
Would love to collaborate here. Lmk if there is task that can be split.

yk:
this looks like a neat plan. I think this would be really supercharged if we had some sort of "instruction/request detector", i.e. a model (or heuristic) that detects when the initial tweet is an instruction. I think this would reduce data noise by a huge amount. do you think that's a good plan?

I've created #143 for this. Do you want to work on that or do you already want to start here?

dhruv2601:
@yk I think that's a good idea, a [deberta](https://huggingface.co/microsoft/deberta-v3-base)-like model would be great for this task. It adapts to varied styles quickly helping us have a singular model to run on all different social media(style of conversation differs)/normal web-styled texts. Further, a task like `instruction classifier` it would need something like 10-15k high-quality samples and there is enough instruction-based data to get good positive and negative samples.

To be fair, heuristic-based would be much quicker to infer and therefore data collection would greatly increase given the huge amount of data we would need, but firstly I will keep it as an offline experiment to see how a few of those would work on a task like this and if there are certain patterns we could use.

I would like to start working on #143 first, you could assign it to me. I will propose a more detailed plan for that task here or on discord DM tonight.

yk:
amazing, thank you!

yk:
you need to comment over there, I can't assign you otherwise 😄

dhruv2601:
Done 👯 
I'll circle back to this current issue of Twitter collection when done with 143.

Jmete:
[...]

PR #620, opened 2023-01-11
126 twitter data

Jmete:
Hello @yk  , @andreaskoepf  , and others, I added code under scripts/data-collection/twitter for converting compressed jsonl twitter archive files into jsonl conversation threads based on the tree and node architecture.  I have ran pre-commit. If there are any issues, please let me know and I can help fix it. This is for issue #126 

Further work will need to be done to filter them into usable instruction -> fulfilment since a lot of twitter threads are spam which we can use the instruction-detector issue 143 on.  We can also use a .env file to make managing paths easier probably.

Note: User should download the large .tar archive files from archive.org and decompress them. We could automate this also in the future if needed but using a downloader like jdownloader is better for archive.org or else it is painfully slow.


scripts/data-collection/twitter/README.md
```diff
@@ -0,0 +1,80 @@
...
+  using the tree and node architecture. This results in aroun 17K conversation
+  threads bassed on a dump of 90M tweets.
+- Script can output the conversation threads into a jsonl file for further
+  filtering or use in models.
+
+## Main Issue
+
+- The issue is that we can easily scrape replies, but there is no guarantee the
+  original tweet is in the archive file. Furthermore, the archives are large so
+  they would need to be kept completely in-memory or in a db to reference. We
```
andreaskoepf, line 50: What is roughly the total size of data we are dealing with (e.g. all compressed files together)?

andreaskoepf (approved):
Clean code and a best-practice readme & overall code documentation. Thanks a lot.

scripts/data-collection/twitter/README.md
Jmete: Thank you @andreaskoepf , much appreciated! 

The interesting part of the twitter archives is the potential size is massive in totality. There is ~128 pages and possibly more of them on archive.org. Each one of those pages has tar files inside with thousands of compressed json files which makes the IO part a bit of a hassle (thus my initial consolidation into parquet files with some pre-processing). I would say **each dump is around 50 - 100GB in tar format**. **_Rough napkin calculation means the total might be around 6 - 12TB of data if we wanted to use all of it._** 

My script doesn't handle the downloading or decompression of the initial tar files since archive.org is slow (it expects folders of compressed json files), but this could be added in the future. I found using external tools specialized for downloading and decompression of large files is easier for initial setup and more efficient.

The initial script for parquet files can handle that but most likely would run into issues in the conversation thread script if we tried to ingest all of it at once to do the conversation thread matching so would need more intelligent code or db usage to handle the full scale in my eyes unless we process only a few dumps at a time and accept any missed connections of tweets between dumps.

I downloaded a few of them for variety for about 65GB of data, and it turned out to be around **90M tweets** after some minimal preprocessing to remove truncated tweets or deletion rows which exist in the json. That 90M tweets then became around **17K english conversation threads**, **but most are spammy / noisy or not useful instructions** by brief manual observations due to the nature of twitter. Filtering and most likely post-processing will be needed to elevate the quality.

Jmete: @andreaskoepf @yk Hi gents, can we merge this? If any adjustments are needed, feel free to let me know.

pruksmhc: @Jmete is the 17k english conversation threads shared anywhere? I'm doing cleanup/aggregation of the data we currently have.

Jmete: @pruksmhc Hey, It is scraped data so currently just the scripts are uploaded publically, but I can share a link on discord. I have 17k that are very messy (most are spam). I am working on an instruction-detector model to help filter that down to around 800 to remove a lot of the gibberish. Those remaining 800 though would still need to go through safety filters, and other filters which will be part of the data pipeline.

PR #622, opened 2023-01-11
Update the web to match the backends new API

### LINKED ISSUE #616: Forward backend task IDs from web to backend during interactions
The backend will soon have a new field `task_id` that should contain the backend's task id.  Instead of sending the frontend's unique task ID, we should start sending the backend's task ID when sending a task interaction.

fozziethebeat:
Closes #616 

The backend now expects a `task_id` field when calling `/api/v1/interaction` which matches the backend's task ID.  The backend currently only requires this for labeling tasks but soon will require it for all tasks.  This forwards that for all interaction types.

Also, this fixes the labeling tasks by ensuring that `sliderValues` is always set properly.

### DIFF
--- a/website/src/pages/tasks/random.tsx
+++ b/website/src/pages/tasks/random.tsx
@@ -9,16 +9,16 @@ const RandomTask = () => {
   const { tasks, isLoading, trigger, reset } = useGenericTaskAPI("random");
 
   useEffect(() => {
-    if (tasks.length == 0) {
+    if (tasks.length === 0) {
       reset();
     }
-  }, [tasks]);
+  }, [tasks, reset]);
 
   if (isLoading) {
     return <LoadingScreen text="Loading..." />;
   }
 
-  if (tasks.length == 0) {
+  if (tasks.length === 0) {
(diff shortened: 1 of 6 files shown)


website/src/pages/tasks/random.tsx
```diff
@@ -9,16 +9,16 @@ const RandomTask = () => {
   const { tasks, isLoading, trigger, reset } = useGenericTaskAPI("random");
 
   useEffect(() => {
-    if (tasks.length == 0) {
+    if (tasks.length === 0) {
       reset();
     }
-  }, [tasks]);
+  }, [tasks, reset]);
```
AbdBarho, line 15: is this useEffect necessary?

fozziethebeat: purged!

fozziethebeat:
Note: the failing end2end tests is expected with the random update.  #623 requests a fix for this.

AbdBarho (approved):

AbdBarho:
I will merge with the hope of having a fix for #623 soon.

PR #624, opened 2023-01-11
Added CodeInstructions notebook

GravermanDev:
CodeInstructions is a notebook which takes a code as an input and converts it into an Instruction


andrewm4894:
@GravermanDev cool - cant wait to play with this :) 

can you add it into the `notebooks/` folder and try follow similar example like this one: https://github.com/GravermanDev/Open-Assistant/tree/main/notebooks/example

pip install looks good at top of notebook maybe just try and:

- add open in colab button cell to the top.
- submit the notebook as ran with output so people browsing the repo can easily just see outputs from each cell.
- maybe change `CodeInstructions.md` to `README.md` so will render the readme automatically in github.

Issue #627, opened 2023-01-11
Does `test-api-contract` need to run on everything?


andrewm4894:
https://github.com/LAION-AI/Open-Assistant/blob/0117159da0651915119c319c86c91b3afaba1544/.github/workflows/test-api-contract.yaml#L3

Wondering if might be a clear subset of folders and or filetypes to only trigger it on?

PR #630, opened 2023-01-11
add some "paths" to `test-api-contract.yaml`

andrewm4894:
It seems to maybe be running a lot of the time. 

A starter PR for this issue: https://github.com/LAION-AI/Open-Assistant/issues/627

### DIFF
--- a/.github/workflows/test-api-contract.yaml
+++ b/.github/workflows/test-api-contract.yaml
@@ -4,11 +4,15 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - "docs/**"
-      - "notebooks/**"
-      - "assets/**"
+    paths:
+      - "oasst-shared/**"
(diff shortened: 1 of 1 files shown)


yk:
I'd rather include than exclude.
also, the same paths need to go into the pull request even

andrewm4894:
> I'd rather include than exclude. also, the same paths need to go into the pull request even

yeah - was just a bit unsure what paths we should use for include - any idea? i can rejig as needed if someone can give me a quick list.

andrewm4894:
Actually looking at the workflow itself it looks like maybe its just 

- oasst-shared/**
- backend/**
- website/**

but was just hesitant in case can be some other dependencies im not aware of

yk (commented):
yes, this looks good, thank you :)

yk (approved):

PR #642, opened 2023-01-12
Updated docker compose command

aasisdhakal:


README.md
```diff
@@ -45,7 +45,7 @@ website, backend, and associated dependent services.
 ##### To start the demo, run this in the root directory of the repository:
 
 ```sh
-docker compose up --build
```
fozziethebeat, line 48: Generally the team has been using `docker compose` without a problem.  What is the motivation to changing the documentation towards `docker-compose`?

aasisdhakal: This will not work and will result in an error as (docker: 'compose' is not a docker command)

The correct syntax should be docker-compose up.

fozziethebeat: You might need to update your docker setup.  `docker compose` is now the preferred command.  We've been consistently using it.  You might want to check [docker's install instructions](https://docs.docker.com/compose/install/linux/) for how to update.

aasisdhakal: i'm sorry thank you so much

PR #643, opened 2023-01-12
Style Privacy Policy & TOS

rsandb:


website/src/components/TOS/PolicyContents.tsx
```diff
@@ -0,0 +1,380 @@
...
+}
+
+export interface SectionInfo {
+  id: number;
+  number: string;
+  title: string;
+  desc: string;
+}
+
+export const TermsData: ChapterInfo[] = [
```
fozziethebeat, line 16: It feels a little weird to hide the contents here.  Can we define these in the actual pages? that would make it easy to look just at `website/src/pages/privacy-policy` and see the policy contents.  Alternatively, these could be embedded in a JSON file that we import.  

Also by moving these data declarations outside of here, you can move the interface declarations into the actual component definitions (since they're basically tied to the component definition).

website/src/components/TOS/SectionCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </Stack>
+        {chapter.sections && chapter.sections.length
+          ? chapter.sections.map((section, sectionIndex) => <SectionCard key={sectionIndex} section={section} />)
+          : ""}
+      </Stack>
+    </Box>
+  );
+};
+
+export const SectionCard = ({ section }: SectionProps) => {
```
fozziethebeat, line 38: Rather than putting both of these components in the same file, can we split them up into two separate files:
 - `website/src/components/PolicySectionCard/PolicySectionCard.tsx`
 - `website/src/components/PolicyChapterCard/PolicyChapterCard.tsx`

Then, later on it'll be easy to add the Storybook stories associated with these and make UI revisions easier.  Once the MVP is settled, I'll be pushing us to break up all the components into simpler sub directories like these.

fozziethebeat: Also by splitting these up you can do `export default const ...` and import via `import PolicySectionCard from ...`

fozziethebeat (changes requested):
Thank you for breaking things up.  I made some small requests that nudge us towards a simpler file directory structure.

fozziethebeat:
Also, you can ignore the failing e2e tests.  They're broken until my fix goes in.

fozziethebeat (approved):

fozziethebeat:
🚢

PR #649, opened 2023-01-12
Semi-comprehensive fix to the web end to end tests

fozziethebeat:
Ensures that the web end to end tests test something meaningful and will fail if the backend doesn't provide any task type.  This isn't perfect or fully comprehensive but at least makes the cypress tests more informative than they currently are.


othrayte (approved):
LGTM

AbdBarho (approved):

PR #656, opened 2023-01-12
#614 - Consistent message ID & FlaggableElement fix

### LINKED ISSUE #614: Web Cleanup: Use a consistent ID for messages
When presenting messages within a task, messages have the field `message_id` that uniquely identifies each message.

When presenting messages within a message timeline or conversation, messages have the field `id` that unique identifies each message.

This is *really* confusing and makes it easy for leave out the correct ID.

We should go through and unify the fields in all the flows so that there's no duplicate fields serving the same purpose.

Klotske:
Might close #614.

Introduced consistent ID's; Changed typings to include `frontend_message_id`.
Included message type in props, separated props from function definitions.

Changed `FlaggableElement` component & `/api/set_label` api route to fix report sending.

Fixed Chakra error in `MessageTableEntry` component (`LinkOverlay` is redundant).

### DIFF
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -25,6 +25,7 @@ import { FlagIcon, QuestionMarkCircleIcon } from "@heroicons/react/20/solid";
 import { useEffect, useState } from "react";
 import fetcher from "src/lib/fetcher";
 import poster from "src/lib/poster";
+import { Message } from "src/types/Conversation";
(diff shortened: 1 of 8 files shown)


fozziethebeat (approved):
There's probably a little more cleanup to the message issue left but this is 90% of what we needed

fozziethebeat:
Merging, this hasn't incorporated the end2end test fixes so the failure is expected

PR #662, opened 2023-01-12
Make example prompt responses more empathatic and direct

mashdragon:
I tried to make the example responses more empathetic, direct, and natural. For instance, telling an angry person to go for a walk might not elicit the best human response, so this sentence was changed to ask the user to be more specific.


yk (approved):
thank you, great eye for detail!

ChristopherKing42:
I like this edit (I struggled with how to write the example empathetically).

However, the electrical example is no longer technically accurate (the human body is a conductor, not an insulator). Also, I notice you omitted the remark about the circuit and fuse box, despite that being one of the more dangerous aspects. What do you think about something like this instead? Is it still emphatic?

> Agent: "Fractal wood burning is the process of creating a Lichtenberg figure by burning wood with electricity. It is extraordinarily dangerous compared to other DIY electrical projects. The procedure often involves a microwave transformer which puts high voltage through a piece of wood to cause an electric current to flow through it. The microwave transformer also usually prevents your power system's fuse box or circuit breaker from tripping, even if the electricity flows through a conductor, such as your body. The voltage is high enough for electricity to flow through things that are ordinarily insulators, such as rubber safety gloves and the air, so a slight mistake or bad luck could deliver a fatal shock or could burn a Lichtenberg figure into your skin.
> 
> There are alternative methods for creating Lichtenberg figures that are safer, such as using static electricity generated by rubbing against a piece of fur or carpet to generate the figure out of ink toner. I recommend asking me about these methods before attempting fractal wood burning."

mashdragon:
> However, the electrical example is no longer technically accurate (the human body is a conductor, not an insulator). Also, I notice you omitted the remark about the circuit and fuse box, despite that being one of the more dangerous aspects. What do you think about something like this instead? Is it still emphatic?

I'm glad you noticed my mistake about the insulator. I am no expert in fractal wood burning or electricity!

I would recommend keeping the language direct and concise. For instance, if you would like to include the part about the circuit breaker, we should be explicit about why that is important:

"The microwave transformer's isolation of the electrical circuit bypasses important safety features in the electrical system, such as circuit breakers, that help protect against electrocution."

I had to look up what the difference between a fuse box and a circuit breaker was, and it appears that fuses aren't really used anymore: https://en.wikipedia.org/wiki/Distribution_board. Therefore, we'll only refer to circuit breakers. In this new version, it explicitly tells the reader why bypassing the circuit breaker is bad: the circuit breaker is a safety device that could prevent electrocution. Now that we have the reason front and center, the casual reader is much more likely to understand.

The part about burning a Lictenberg figure into your skin, even if that's possible, is not necessary. In addition, "a slight mistake or bad luck" seems a little repetitive. Here's how I would revise that sentence:

"The voltage is high enough for electricity to flow through things that are ordinarily insulators, including the air around you—even just a single mistake could deliver a fatal shock."

I'm glad you added the part about the ink toner, since that adds a better description.

I'm not sure about the call to action to ask the assistant for something (in the last sentence). I prefer the one I wrote, since it doesn't portray the assistant as the only source of information.

<hr>

I don't think we have a formal style guide yet, so most of these are my own personal style preferences. That said, I try to be a good editor! It would be great if we could reproduce the professional style of writing that ChatGPT achieves.

ChristopherKing42:
@mashdragon I quite like that wording, yeah! That also sounds technically accurate as far as I can tell. You're right that fuses are very rare now-a-days, so probably don't need to be mentioned.

(At first I was afraid that "I recommend looking into these methods before attempting fractal wood
burning." sounded like telling the user to "get lost", but your explanation makes sense; checking with online sources is much better than just relying on the assistant. For a related discussion see https://github.com/LAION-AI/Open-Assistant/pull/290.)

PR #663, opened 2023-01-12
website: Combine fetcher and poster files, update all references to use axios for both GET and POST

rjmacarthy:
Hi there,

Firstly, thank you for the amazing work you are doing it's truly inspirational and I'm happy to help if I can. After browsing the front-end code I propose a change. I've made some updates to the application in this PR and I wanted to share them with you. 

I've combined the `fetcher` and `poster` files into one file, `api.ts`, and standardised the use of `axios` for both GET and POST requests. This should simplify the code and make it easier to maintain going forward. I've also updated all references to these methods throughout the codebase.  I noticed that data is run through `JSON.stringify(arg)` I've removed the need for it.  By creating an axios api instance we can setup an interceptor which controls the response and reject in one place.

Please let me know if you have any questions or concerns, and I'll be happy to address them.

Thank you

### DIFF
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -12,7 +12,6 @@ interface MessageTableEntryProps {
 export function MessageTableEntry(props: MessageTableEntryProps) {
   const { item } = props;
   const backgroundColor = useColorModeValue("gray.50", "gray.800");
-
   return (
     <div>
       <FlaggableElement message={item} key={`flag_${item.id || item.frontend_message_id}`}>
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -23,8 +23,7 @@ import {
 } from "@chakra-ui/react";
 import { FlagIcon, QuestionMarkCircleIcon } from "@heroicons/react/20/solid";
 import { useEffect, useReducer } from "react";
-import fetcher from "src/lib/fetcher";
-import poster from "src/lib/poster";
+import { get, post } from "src/lib/api";
 import { Message } from "src/types/Conversation";
 import { colors } from "styles/Theme/colors";
 import useSWR from "swr";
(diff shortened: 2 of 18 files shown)


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -6,7 +6,7 @@ import { FlaggableElement } from "src/components/FlaggableElement";
 interface Message {
   text: string;
   id: string;
-  message_id: string;
+  frontend_message_id: string;
```
fozziethebeat, line 9: You'll need to merge, I just merged in a change that I think fixes this.

fozziethebeat (approved):
Thank you! I like this cleanup.

You'll need to merge against main as I think you fixed a bug someone else just fixed.

fozziethebeat:
Also, from the build failure it looks like there's still an import site for `src/lib/fetcher`.  That'll need to be fixed too.

AbdBarho (approved):

website/src/components/Messages/MessageTableEntry.tsx
rjmacarthy: Thank you very much for your review.

rjmacarthy:
> Also, from the build failure it looks like there's still an import site for `src/lib/fetcher`. That'll need to be fixed too.

Thank you for your review, I have addressed this and updated my pull request.

fozziethebeat (changes requested):
The cypress contract tests break due to something changed in `oassist_api_client.ts` Can you investigate?

fozziethebeat:
If possible, maybe revert the changes to the `oassist_api_client.ts`.  That's used only on the API routes and not on the client side so it's actually not using axios, it's using `node-fetch`.  That library is working well enough and doesn't need refactoring.  But our `useSWR` callsites _definitely_ need cleaning up.

rjmacarthy:
> If possible, maybe revert the changes to the `oassist_api_client.ts`. That's used only on the API routes and not on the client side so it's actually not using axios, it's using `node-fetch`. That library is working well enough and doesn't need refactoring. But our `useSWR` callsites _definitely_ need cleaning up.

Thank you. I have reverted the change to this file and updated the code to remove the dependency of environment variables for non proxied api calls.

fozziethebeat (approved):
Thank you for the changes! everything passes!

PR #678, opened 2023-01-13
installed sharp for next standalone deployment

yk:
Solves the following issue:
https://nextjs.org/docs/messages/sharp-missing-in-production


fozziethebeat:
I want to hold off on this.  I installed this before and it breaks docker builds on mac M1 and M2 machines.  I spent quite a bit of time trying to figure out why and removed Sharp for the time being.

fozziethebeat (approved):
Checked on my machine and it apparently now works.  Docker and mac is now a mystery to me.

Issue #675, opened 2023-01-13
Add `Report bug` button


AbdBarho:
This should allow users to quickly create an issue in github.

This button should be shown on the 404 page.

Arguably it should also be available everywhere, maybe from the dropdown on the user avatar in the top right?

fozziethebeat:
Updating the 404 page and adding a 500 page

sir-braz:
![Screenshot from 2023-01-13 22-04-28](https://user-images.githubusercontent.com/89554544/212443256-95f5d2e2-e2b9-47c4-9fc3-cdd0445b52a2.png)

Hi, I'm trying to make a pull request, but I don't have permission to push the file to the branch I created, 675-create-a-button-report

fozziethebeat:
Oh, i have a PR for this already.  I didn't notice you were also working on this.

We've updated the contributing doc on how to best get started: https://github.com/LAION-AI/Open-Assistant/blob/main/CONTRIBUTING.md#submitting-work

The brief recommendation is to fork the repo, make a new branch in your forked version, and then push your changes there.  Then when ready, make a pull request to merge in your forked branch.

PR #681, opened 2023-01-13
#675 Report Bug Button

Klotske:
Partial fix to #675.

Added new entry under `UserMenu` to create new issues.
For now - opens the link in the current tab.
It's possible to open a link in new tab, but requires some tweaking.

**Demo:**
demo


fozziethebeat (approved):
I'm loving these small targeted PRs

Issue #671, opened 2023-01-13
Optimize answer submission flow for review


fozziethebeat:
Right now when a user hits the `submit` button the answer is submitted immediately.  It's possible for the user to change their answer and trigger a new submit that will fail.  They then have to click a `next` button.

We should improve this in two steps:
1) First, remove the `next` button and fetch a new task immediately after submission.  This removes the ability to edit a task that's already been answered and enter into a failure state.
2) Add a `review` button.  When a user writes their answer, they first hit `review` to lock their answer.  They can either revert to an edit mode (with an `edit` button somewhere) or the can `submit` to send the answer and fetch a new task.

othrayte:
I'll take this on if available.

othrayte:
I think we have completed this and can close it.

PR #683, opened 2023-01-13
website: Remove next task button and load new task immediately

othrayte:
Pressing "submit" now also loads the next task.

Also disable the task inputs while the task is submitted and before the new task is ready in case the users network is slow or something.

Most of the work was in disabling the inputs.

This is the first step towards #671.

### DIFF
--- a/website/src/components/Sortable/Sortable.stories.tsx
+++ b/website/src/components/Sortable/Sortable.stories.tsx
@@ -1,0 +1,33 @@
+import React from "react";
+
+import { Sortable } from "./Sortable";
+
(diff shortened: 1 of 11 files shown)


website/src/components/Sortable/Sortable.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
```
fozziethebeat, line 1: Stories! Rock on!

fozziethebeat (approved):

fozziethebeat:
🔥

PR #687, opened 2023-01-13
#668 - Add simple label

### LINKED ISSUE #668: Simplify labeling task when there's only one label (spam)
When the user gets a labeling task asking for just the spam label we should present a clearer and easier to answer question (does this have spam: yes/no).  This should make the task faster to accomplish and be more obviously differentiated from the normal labeling task.

Klotske:
Closes #668.

Added new `LabelRadioGroup` component which reuses some logic from `LabelSliderGroup`.
Thought it would be better if component supported multiple labels if we want to simplify UX.
`LabelRadioItem` contains `states` prop to configure _(specify the text, value or color)_ presented buttons.
**Here's a demo:**
image
image

Blocked by #686.

### DIFF
--- a/website/cypress/e2e/tasks/random.cy.ts
+++ b/website/cypress/e2e/tasks/random.cy.ts
@@ -44,12 +44,44 @@ describe("handles random tasks", () => {
                 break;
               }
               case "label-task": {
-                // Clicking on the slider will set the value to about the middle where it clicks
-                cy.get('[aria-roledescription="slider"]').first().click();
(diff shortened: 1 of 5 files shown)


andreaskoepf:
Just a first-impression usability opinion: YES -> red, No -> green feels strange .. even though it may make sense for a "Spam?" question. Maybe we can ask a slightly different question for which Yes becomes the positive answer that could be shown on a green button?

fozziethebeat (approved):

fozziethebeat (approved):

PR #688, opened 2023-01-13
website: Fix Header.stories.js rendering add FlagsProvider

rjmacarthy:
Hi!

The story at `src/components/Header/Header.stories.jsx` is not loading correctly in when running storybook.  

It needs to be wrapped in the `FlagsProvider` as is using the Flags component. If the `Flags` component is going to be used in more places it would be advisable to write a storybook decorator for it.

Please let me know if you have any questions or concerns.

Thank you,


fozziethebeat (approved):

fozziethebeat:
I think in a followup PR we should figure out where to put this in storybook's main wrapper code.  I would have to investigate where that is.

PR #692, opened 2023-01-14
Udpating the leaderboard components to use real data

### LINKED ISSUE #672: Fetch leaderboard data from the real backend endpoint or hide the features
The backend currently supplys information for two different leaderboards depending on the answer type.  Later this will be replaced with a unified leaderboard that accounts for all activities.

We should call just one of these endpoints to supply to leaderboard data and later update it to use the unified backend endpoint.  This will make the leaderboard data partrially real and useful.

fozziethebeat:
Closes #672 

This does a few things:
  - Refactors the `RankItem` component to encapsulate all of the leaderboard grid functionality (with a rename to `LeaderboardGridCell`)
  - Moves the leaderboard page to the top level of pages and out of a directory
  - Updates the `LeaderboardTable` and `LeaderboardGridCell` to fetch data from an API route
  - Create a new API route that fetches data from the backend.  The URL it's using will change later.
  - We also comment out a few leaderboard features that aren't supported yet.


website/src/components/LeaderboardGridCell/LeaderboardGridCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+                  {title}
+                </Text>
+
+                <FiChevronDown size="16" />
+              </Box>
+            </GridItem>
+          ))}
+        </GridItem>
+      </Grid>
+      <Grid templateColumns={columns} bg={backgroundColor} borderRadius="xl" shadow="base" p="4" gap="6">
```
AbdBarho, line 40: is there a  reason why these need to be 2 separate grids?

the columns won't be aligned between the headers & the rows.

fozziethebeat: Right now they line up well enough with the column fields. 

But I think the answer is no, they can be the same grid or a simple table.  Ultimately I think this component and `LeaderboardTable` should be the same.  But I want to assign that as a UI cleanup to Riley as he's better at that than I am.  I mostly wanted to do a little re-organizational cleanup and fold in the real data.

AbdBarho (approved):

PR #693, opened 2023-01-14
Message Styling For Mobile

rsandb:
Decreased padding for mobile and changed some styling to give better contrast

### DIFF
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -1,4 +1,5 @@
 import {
+  Box,
   Button,
   Checkbox,
   Flex,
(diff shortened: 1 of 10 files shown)


AbdBarho (changes requested):
I like the modified design!

Could please add the tooltips back to the report button.

Also, maybe it makes sense to have the messages take full width of parent container like before.
image

fozziethebeat:
No comments from me.  We can merge this after addressing @AbdBarho's requests.

rsandb:
> I like the modified design!
> 
> Could please add the tooltips back to the report button.
> 
> Also, maybe it makes sense to have the messages take full width of parent container like before. image

Sure thing!

The messages take up full width at the medium breakpoint. The reason for this is to avoid strain on the users when reading long blocks of text :)

AbdBarho (approved):

PR #694, opened 2023-01-14
Update UserMenu

rsandb:
now uses chakra components and includes score. Score needs to be implemented, currently is just a placeholder

### DIFF
--- a/website/src/components/Header/UserMenu.tsx
+++ b/website/src/components/Header/UserMenu.tsx
@@ -13,7 +13,7 @@ import {
 } from "@chakra-ui/react";
 import { signOut, useSession } from "next-auth/react";
 import React from "react";
-import { FiAlertTriangle, FiHelpCircle, FiLayout, FiLogOut, FiSettings, FiShield } from "react-icons/fi";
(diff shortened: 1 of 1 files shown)


website/src/components/Header/UserMenu.tsx
```diff
@@ -29,12 +37,20 @@ export function UserMenu() {
...
         href: "https://github.com/LAION-AI/Open-Assistant/issues/new/choose",
         desc: "Report a Bug",
         icon: FiAlertTriangle,
       },
+      {
+        name: "FAQ",
+        href: "/faq",
+        desc: "Report a Bug",
+        icon: FiHelpCircle,
+      },
```
AbdBarho, line 53: do we have a FAQ page?

fozziethebeat: We don't.  We can drop this entry for now.

fozziethebeat:
Approve after addressing @AbdBarho's request for changes.

fozziethebeat (approved):
Confirm that the requested change is done! merging.

PR #695, opened 2023-01-14
Style Empty State and 404

rsandb:
Update styling to match rest of app

### DIFF
--- a/website/src/pages/404.tsx
+++ b/website/src/pages/404.tsx
@@ -1,22 +1,31 @@
-import { Button, Link, Stack } from "@chakra-ui/react";
+import { Box, Button, Center, Link, Text, useColorModeValue } from "@chakra-ui/react";
 import Head from "next/head";
-import NextLink from "next/link";
+import { useRouter } from "next/router";
 import { FiAlertTriangle } from "react-icons/fi";
+import { PageEmptyState } from "src/components/EmptyState";
+import { getTransparentHeaderLayout } from "src/components/Layout";
 
-export default function Error() {
+function Error() {
+  const router = useRouter();
+  const backgroundColor = useColorModeValue("white", "gray.800");
+
   return (
     <>
       <Head>
         <title>404 - Open Assistant</title>
         <meta name="404" content="Sorry, this page doesn't exist." />
       </Head>
-      <main className="flex h-3/4 items-center justify-center overflow-hidden subpixel-antialiased text-xl">
-        <Stack>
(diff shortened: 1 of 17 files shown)


website/src/pages/404.tsx
```diff
@@ -1,15 +1,37 @@
...
         <meta name="404" content="Sorry, this page doesn't exist." />
       </Head>
-      <main className="flex h-3/4 items-center justify-center overflow-hidden subpixel-antialiased text-xl">
-        <p>Sorry, the page you are looking for does not exist.</p>
-      </main>
+      <Center bg={backgroundColor2} flexDirection="column" gap="4" fontSize="lg" className="subpixel-antialiased">
+        <Box bg={backgroundColor} p="10" borderRadius="xl" shadow="base">
+          <Box display="flex" flexDirection="column" alignItems="center" gap="8">
+            <FiAlertTriangle size="30" color="DarkOrange" />
+            <Box display="flex" flexDirection="column" alignItems="center" gap="3">
```
AbdBarho, line 22: would make sense to have `EmptyState` accept children so you can re-use it here?

website/src/pages/404.tsx
```diff
@@ -1,15 +1,37 @@
...
+              <Text>Sorry, the page you are looking for does not exist.</Text>
```
fozziethebeat, line 23: Can you merge in my changes? I added in a button to file a bug here.

website/src/pages/404.tsx
fozziethebeat: I agree, `EmptyState` should be a more generic component that takes in children.

fozziethebeat (changes requested):

fozziethebeat:
Also, if you're feeling motivated, I added a `500.tsx` page.  That could use the same styling update as I copied it from `404`.

rsandb:
> Also, if you're feeling motivated, I added a `500.tsx` page. That could use the same styling update as I copied it from `404`.

Is it just in the pages dir? I can't find it

website/src/pages/404.tsx
rsandb: working on it now

website/src/pages/404.tsx
fozziethebeat: I think this is sufficient for now and a fairly large PR.  my request is to make a follow up PR to make `EmptyState` a re-usable component and update only other components using `EmptyState`.

fozziethebeat (approved):

Issue #685, opened 2023-01-13
Fix expand button for long texts in ranking tasks


AbdBarho:
The three dots here should be a button that expands to show the entire content.

![image](https://user-images.githubusercontent.com/24505302/212347332-463dfc20-a651-4f06-a4da-baf0eaa6d675.png)

Currently it is not functioning

othrayte:
I think it does have an implementation that shows the full text but at the moment I find I have to click and lot (and quickly) on the "..." to open the "Full Text" popup, and again on the popup's "x" to close it.

![dot-dot-dot-fulltext](https://user-images.githubusercontent.com/241372/212448810-f53cec61-5931-47b3-92a2-6015b56d3c58.gif)

othrayte:
Also, I didn't realize what was happening at the time but you can see in that GIF that if you drag (on) the modal dialog the items move.

othrayte:
The fix for this is simple, turns out it was something that I had looked into previously, I'll raise a PR in a moment.

AbdBarho:
I believe we can close this now!

PR #696, opened 2023-01-14
website: Fix responsiveness of '...' in sortable item

othrayte:
This fixes the core issue with #685 and makes the '...' button work reliably in the sortable items and subsequent dialog easily closable.

It doesn't fix the issue where the modal dialog can be used to drag the sortable item or the fact that the '...' doesn't look like it can be clicked on.

Note: the change from PointerSensor to MouseSensor was required to make this work and dnd-kit says that PointerSensor shouldn't be used with touch sensor, but MouseSensor can be.


AbdBarho (approved):
Great!

PR #697, opened 2023-01-14
website: Make the more text button clearer

othrayte:
Before (yes it's a button, see #685):
Screenshot from 2023-01-14 15-40-37

After:
Screenshot from 2023-01-14 15-40-01

This is a suggestion, it was not explicitly mentioned that this was an issue in #685, but I figured it could be a compounding factor.

If you liked this PR you might also like these PRs:
* [website: Fix responsiveness of '...' in sortable item](https://github.com/LAION-AI/Open-Assistant/pull/696)


AbdBarho (approved):
Thanks!

PR #698, opened 2023-01-14
Small typo fix + Consistent Docstrings

kayjay-is-here:
Also removed a redundant comment.

### DIFF
--- a/backend/oasst_backend/config.py
+++ b/backend/oasst_backend/config.py
@@ -77,7 +77,7 @@ class Settings(BaseSettings):
     DEBUG_USE_SEED_DATA_PATH: Optional[FilePath] = (
         Path(__file__).parent.parent / "test_data/realistic/realistic_seed_data.json"
     )
-    DEBUG_ALLOW_SELF_LABELING: bool = False # allow users to label their own messages
+    DEBUG_ALLOW_SELF_LABELING: bool = False  # allow users to label their own messages
     DEBUG_SKIP_EMBEDDING_COMPUTATION: bool = False
 
     HUGGING_FACE_API_KEY: str = ""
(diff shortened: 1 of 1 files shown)


kayjay-is-here:
There's really no point to this, but I mean it doesn't really hurt to make it look nicer.

backend/oasst_backend/config.py
```diff
@@ -77,7 +77,7 @@ class Settings(BaseSettings):
     DEBUG_USE_SEED_DATA_PATH: Optional[FilePath] = (
         Path(__file__).parent.parent / "test_data/realistic/realistic_seed_data.json"
     )
-    DEBUG_ALLOW_SELF_LABELING: bool = False  # allow users to label their own messages
```
andreaskoepf, line 80: Thanks for the replies typo fix, but do you think self-labeling is really self-explanatory and it is better to remove the comment?

backend/oasst_backend/config.py
```diff
@@ -20,28 +20,28 @@ class TreeManagerConfiguration(BaseModel):
...
     goal_tree_size: int = 15
-    """Total number of messages to gather per tree"""
+    """Total number of messages to gather per tree."""
 
     num_reviews_initial_prompt: int = 3
     """Number of peer review checks to collect in INITIAL_PROMPT_REVIEW state."""
 
     num_reviews_reply: int = 3
-    """Number of peer review checks to collect per reply (other than initial_prompt)"""
+    """Number of peer review checks to collect per reply (other than initial_prompt)."""
```
andreaskoepf, line 29: (I omitted the full-stop at the end of these comments because they are not full sentences (a verb is missing))

andreaskoepf (changes requested):
I'd be happy to merge the docstring changes but I don't agree that the self-labeling comment is redundant and should be removed.

kayjay-is-here:
@andreaskoepf any idea why pre-commit suddenly fails? This doesn't make any sense

andreaskoepf (approved):
Thanks!

Issue #673, opened 2023-01-13
Support saving notes on admin decisions for users


fozziethebeat:
The backend supports saving notes on why a user's status has been changed.  We should forward those notes to the backend and store the admin decisions.

fozziethebeat:
Further, we should update the admin page to call from the real backend instead of the web's database:
- `/api/v1/frontend_users/` To fetch all users according to their frontend status.

Additionally, when we mark a user as banned, we should call `/api/v1/users/users/{user_id}` and set them as not enabled.  This will require fetching their backend uuid.

PR #701, opened 2023-01-14
673 enhanced admin management

fozziethebeat:
Applies to #673 

Updates the admin views to user the backend's User API routes:
- Fetch all users from the backend and merge them with what the web knows
- Update web API routes to use the more updated backend routes
- Set the user's enabled and notes fields when managing a user

What's not supported yet:
- Pagination.  This is broken but will be fixed in a follow up PR.  How pagination works has changed when using the backend strategy.


website/src/pages/api/admin/user_messages.ts
```diff
@@ -1,15 +1,12 @@
...
   const { user } = req.query;
-
-  const messagesRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/frontend_users/local/${user}/messages`, {
-    method: "GET",
-    headers: {
-      "X-API-Key": process.env.FASTAPI_KEY,
-    },
-  });
-  const messages = await messagesRes.json();
+  const messages = await oasstApiClient.fetch_user_messages(user as string);
```
AbdBarho, line 9: string array? or `Message[]`?

website/src/pages/api/admin/users.ts
```diff
@@ -1,31 +1,44 @@
...
+    result.set(user.id, user.role);
+    return result;
+  }, new Map());
+
+  const users = all_users.map((user) => {
+    const role = local_user_map.get(user.id) || "general";
+    return {
+      ...user,
+      role,
+    };
```
AbdBarho, line 41: will this happen every time you open the users dashboard?

website/src/pages/api/admin/user_messages.ts
fozziethebeat: Added the explicit type.

website/src/pages/api/admin/users.ts
fozziethebeat: Yes, it's not ideal and not something I want to keep for long.  I think the real fix is to push the role information to the backend but that would require a variety of new routes to make that fetching and setting easier.

AbdBarho (approved):

andreaskoepf (approved):
Front-end change approved!

yk (approved):
lgtm, thank you

fozziethebeat:
Merging now that I see approval from all required people.

PR #703, opened 2023-01-14
fix task card background color

rsandb:
colors got changed through merges


rsandb:
before
image

after
image

AbdBarho (approved):

Issue #670, opened 2023-01-13
Display welcome information for brand new web users


fozziethebeat:
When a new user signs into the website for the first time, they should see some informative widget or popup when they reach the dashboard.  We can leverage NextAuthJS's `isNewUser` field during the signin event and include some query param when redirecting the user.

AbdBarho:
@fozziethebeat would it make sense to show this pop-up per device instead of per user? by using `localStorage`?

fozziethebeat:
I think just per user right now is sufficient.  Let's keep it simple and support per device if we see people really getting confused.

rsandb:
working on this now

AbdBarho:
@rsandb @fozziethebeat 

it seems that the different auth providers have different behaviours, not all have the `isNewUser` attribute (or maybe I am using them incorrectly).

rsandb:
@fozziethebeat @AbdBarho 

How do you guys feel about putting a Help button where the color mode toggle is? Then we can move the color mode toggle to the usermenu. It would make more sense to have colormode in a menu as it's not very important to have on the dashboard all the time

rsandb:
The help button would launch the quick start modal

fozziethebeat:
I checked huggingface and they have the color model toggle in the left side.  I think keeping it there is cleaner and a little simple.  The user drop down is for navigation rather than UI settings.

The Help Icon could go in the header role.  But I think we're planning on something simpler.  When a user lands on the dashboard there would be a query param (`show_help_stuff`).  When that's set, we render some help widget.  I don't think we need an explicit button that's present all the time.

Let me poke at the NextAuth stuff to see when `isNewUser` works.

fozziethebeat:
the `signin` event in NextAuth doesn't let us change the query params to the redirected path.

I tried inspecting when the token and sessions are created and there's nothing consistent we can do.

So instead I think we can do the following:
- Create a new db field: `isNew`.  Set that to true by default.
- Include that in the token & session
- Set `isNew` to false when the user submits their first answer (they've done something meaningful)

I will make a PR to do this and then let others handle the presentation when `session.user.isNew` is true.

rsandb:
Sounds good. I'll finish that up tonight then

AbdBarho:
Finished

PR #706, opened 2023-01-14
Tag in the user table when they are new and include it in use session

fozziethebeat:
Applies to #670 

When a new user signs up we tag them as `isNew` in the User table.  When they contribute data we update the field to be `false`.

The `isNew` field is forwarded during JWT and session creation so that we can present onboarding information.  The one flaw with this approach is that they will see the onboarding information until they create a new session.


rsandb:
finished with modal and will PR after this gets merged

AbdBarho (approved):

PR #710, opened 2023-01-14
599 add row versioning to backend tables

melvinebenezer:
### Solution - Draft 1 for Review

- Use isolation level to Repeatable read
- Replace all db.commit() with wrapper (decorators) to retry on concurrent update error **sqlalchemy.exc.OperationalError**
- Current Max retries set to 3
- Some log messages and commented code left until Review of solution
- Suggestions required on Errors to be raised regarding exhaustion of max retires


backend/oasst_backend/utils/database_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                        result = f(self, *args, **kwargs)
+                        if flag:
+                            db.commit()
+                        else:
+                            db.flush()
+                        return result
+                    except OperationalError:
+                        db.rollback()
+                        pass
+                raise Warning("DATABASE_MAX_RETIRES_EXHAUSTED")
```
andreaskoepf, line 73: should become `OasstError()`

backend/oasst_backend/utils/database_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                logger.exception("Seed data insertion failed")
+            except Exception as e:
+                raise e
+
+        return wrapped_f
+
+    return decorator
+
+
+def manage_method_db_transaction(db, flag=True, num_retries=3):
```
andreaskoepf, line 57: flag should be `auto_commit: bool = True`

andreaskoepf: or maybe define an enum with `flush`, `commit` and `none` as options?

backend/oasst_backend/utils/database_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                        result = f(self, *args, **kwargs)
+                        if flag:
+                            self.db.commit()
+                        else:
+                            self.db.flush()
+                        return result
+                    except OperationalError:
+                        self.db.rollback()
+                        pass
+                logger.exception("DATABASE_MAX_RETIRES_EXHAUSTED")
```
andreaskoepf, line 23: I think logger.exception should be used in except handlers to log the exception .. to signal an error `logger.error()` is better

backend/oasst_backend/task_repository.py
```diff
@@ -129,8 +131,9 @@ def bind_frontend_message_id(self, task_id: UUID, frontend_message_id: str):
         task.ack = True
         # ToDo: check race-condition, transaction
```
andreaskoepf, line 132: yay ..my old comment can finally be removed when this gets merged. :-)

backend/oasst_backend/utils/database_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                for i in range(num_retries):
+                    try:
+                        result = await f(self, *args, **kwargs)
+                        if flag:
+                            self.db.commit()
+                        else:
+                            self.db.flush()
+                        return result
+                    except OperationalError:
+                        self.db.rollback()
```
andreaskoepf, line 46: Is the rollback necessary here? And we should probably do at least minimal logging if we retry.

backend/oasst_backend/utils/database_utils.py
melvinebenezer: Sure, Done

backend/oasst_backend/utils/database_utils.py
melvinebenezer: Made the changes

backend/oasst_backend/utils/database_utils.py
melvinebenezer: changed this

backend/oasst_backend/task_repository.py
melvinebenezer: :) removed it

backend/oasst_backend/utils/database_utils.py
melvinebenezer: Adding minimal logging here

backend/oasst_backend/prompt_repository.py
```diff
@@ -104,8 +106,8 @@ def insert_message(
             review_result=review_result,
         )
         self.db.add(message)
-        self.db.commit()
-        self.db.refresh(message)
+
+        # self.db.refresh(message)
```
andreaskoepf, line 110: could we reactivate this (e.g. sever-side `created_date`)

backend/oasst_backend/prompt_repository.py
```diff
@@ -312,10 +317,9 @@ def insert_toxicity(self, message_id: UUID, model: str, score: float, label: str
 
         message_toxicity = MessageToxicity(message_id=message_id, model=model, score=score, label=label)
         self.db.add(message_toxicity)
-        self.db.commit()
-        self.db.refresh(message_toxicity)
```
andreaskoepf, line 316: we could first keep all existing refresh() code .. just replacing `db.commit()` by `db.flush()` ...

backend/oasst_backend/task_repository.py
```diff
@@ -128,6 +129,7 @@ def store_task(
         assert task_model.id == task.id
         return task_model
 
+    @managed_tx_method(CommitMode.COMMIT)
```
andreaskoepf, line 132: many of those methods are now still Commit .. shouldn't they all become FLUSH and we only commit in the outer transaction, e.g. in the controller where the web call is handled?

oasst-shared/oasst_shared/exceptions/oasst_api_error.py
```diff
@@ -17,6 +17,7 @@ class OasstErrorCode(IntEnum):
     GENERIC_ERROR = 0
     DATABASE_URI_NOT_SET = 1
     API_CLIENT_NOT_AUTHORIZED = 2
+    DATABASE_MAX_RETRIES_EXHAUSTED = 3
```
andreaskoepf, line 20: I think 3 was now already taken by Yannic.. ;-). ..

andreaskoepf (changes requested):
The main file database_utils.py looks good. But we could now remove the commit calls in many of the functions and simply use flush there .. i.e. they don't need to be `managed_tx_methods()` . only the most outer calls should open the transaction, e.g. one transaction for the whole request.

backend/oasst_backend/task_repository.py
melvinebenezer: I have done the same. Added to FLUSH to more atomic functions(i.e functions used by others) and COMMIT to outer functions.

This looked to be in the outer transaction. that way this logic of managed transaction is encapsulated inside the pr (PromptRepository) and the web call handler need not know of the db tx details
e.g 
pr.task_repository.bind_frontend_message_id(task_id=task_id, frontend_message_id=ack_request.message_id)

@router.post("/{task_id}/ack", response_model=None, status_code=HTTP_204_NO_CONTENT)

backend/oasst_backend/prompt_repository.py
melvinebenezer: if the intention is to get the database-side value for all the attributes of Message then this is achieved by the FLUSH in the decorator. It turns the lazy loading to eager loading. Message will contain all the values before it is returned to the caller

backend/oasst_backend/prompt_repository.py
melvinebenezer: if the intention is to get the database-side value for all the attributes of Message then this is achieved by the FLUSH in the decorator. It turns the lazy loading to eager loading. Message will contain all the values before it is returned to the caller. The refresh is not required here. let me know your opnion

andreaskoepf (approved):
ok, ready to merge for MVP. We will have to move the transactions to outer scopes (e.g. to rest-call level) later.

PR #719, opened 2023-01-14
export and auto-commit `docs/docs/api/openapi.json` as part of `test-api-contract` workflow

andrewm4894:
It seems that an `openapi.json` is generated and used as part of the `test-api-contract` workflow. This PR just 

- copies that `openapi.json` file to `docs/docs/api/openapi.json` as part of `start-mock-server.sh`.
- adds a step to the end of the `test-api-contract` workflow that just auto-commits changes to the `docs/docs/api/openapi.json` file.
- excludes `docs/docs/api/openapi.json` from `pre-commit` since its already been formatted by `jq`.

[Here](https://github.com/andrewm4894/Open-Assistant/commit/b9cfd6f0a70bac099761dd1ac53117f496517054) is an example of the commit made by the workflow if the file `docs/docs/api/openapi.json` changes.


.github/workflows/test-api-contract.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - add-api-docs-workflow
```
andrewm4894, line 7: if/once we happy to merge i will make a final commit to this PR to remove this branch - just needed as part of testing this PR.

.github/workflows/test-api-contract.yaml
```diff
@@ -44,3 +45,8 @@ jobs:
         run: ./scripts/frontend-development/run-contract-test.sh
 
       - run: ./scripts/backend-development/stop-mock-server.sh
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          file_pattern: "docs/docs/api/openapi.json"
```
andrewm4894, line 51: this ensures will only ever auto-commit for the `docs/docs/api/openapi.json` file to reduce risk of any other files accidentally getting autocommited for whatever reason.

yk:
what happens if the file remains unchanged? no commit?

andrewm4894:
> what happens if the file remains unchanged? no commit?

yep no commit. 

just need to make the `openapi.json` file pretty so `pre-commit` wont freak out and fail next time anyone makes a PR. last commit hopefully sorts that - just doing last few checks.

andrewm4894:
draft while if slay pre-commit

andrewm4894:
pre-commit just complaining about this:

yk (approved):
we can also include openapi.json from pre-commit. it's auto-generated, so the formatting is fixed

andrewm4894:
> we can also include openapi.json from pre-commit. it's auto-generated, so the formatting is fixed

going to see if can just run pre-commit on the file as part of the script and be done with it that way. If fails then yep - i might add it to the `exclude` in the pre-commit yaml.

Klotske:
Seems like this PR broke contract testing...

All the recent merges are failing on `stefanzweifel/git-auto-commit-action@v4`.

Here's a sample of log with error:
image

andrewm4894:
ah i think it's hitting some sort of github error relating to the branch protection stuff - i'll make a quick PR to just turn this step off for now while we try figure it out.

andrewm4894:
seems like we need to do this: https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches

PR #722, opened 2023-01-15
Nicer docusaurus openapi docs

andrewm4894:
- uses [this plugin](https://github.com/cloud-annotations/docusaurus-openapi) to make a nice static site docusaurus based api section of docs site.
- adds some nicer postman code snippets etc too as part of api docs.
- makes api docs more "self hosted" as opposed to the dynamic swagger url approach.

image


yk (approved):
thank you!

PR #723, opened 2023-01-15
turn off auto-commit on `docs/docs/api/openapi.json` as clashing with branch protection

andrewm4894:
this PR should [fix this](https://github.com/LAION-AI/Open-Assistant/pull/719#issuecomment-1382966107) by disabling last step to auto-commit docs/docs/api/openapi.json. 

the auto-commit seems to be clashing with GH branch protection rules - will need to read up to see how to resolve. 

looks like to resolve we would need to do this: seems like we need to do this: https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches

i think would need to be a repo admin who can make a access token the action can use.

### DIFF
--- a/.github/workflows/test-api-contract.yaml
+++ b/.github/workflows/test-api-contract.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - add-api-docs-workflow
     paths:
       - "oasst-shared/**"
       - "backend/**"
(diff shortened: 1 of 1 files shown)


jack-michaud:
Just a thought - instead of committing the file to the repository, have you considered generating the file only in the gh-pages branch? I'm not familiar with the docusaurus flow, but just brainstorming ways to get around the branch protection.

fozziethebeat (approved):
plz submit soon.  This is adding unexpected and undesired behavior to web pull requests.

AbdBarho (approved):

andrewm4894:
Auto merge is on, it just needs approval from @yk or @andreaskoepf and should merge straight away then

yk (approved):

PR #729, opened 2023-01-15
Fix FlaggableElement grouping issue for desktop

rsandb:
Before (Items not grouped)
image

After (Items nice n' tidy)
image

Mobile
image

The reason for keeping these items grouped is to make it easier and faster to identify which message the tooltip belongs to. It also just looks awkward for a bunch of tooltips to sit in empty space

Also replaced grid with box because it wasn't using any grid props and was just calling an extra element

### DIFF
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -3,7 +3,6 @@ import {
   Button,
   Checkbox,
   Flex,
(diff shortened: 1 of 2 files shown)


fozziethebeat (approved):

fozziethebeat:
The failed test contract seems unrelated to your changes.  Merging.

PR #730, opened 2023-01-15
Ensuring the user can find the privacy and terms of service 

### LINKED ISSUE #394: Hard to find privacy policy and terms of service when on Website Dashboard page
With the  new Dashboard layout on the web, it's very difficult to find the privacy policy or terms of service.  I think one minimal fix is to add the `Footer` component to the page.  But can we design something visually more appealing?

Screenshot 2023-01-05 at 9 45 32 AM

fozziethebeat:
Closes #394 

ensures the privacy and terms of service links are always findable by including the footer in the dashboard layout.

### DIFF
--- a/docs/docs/api/openapi.json
+++ b/docs/docs/api/openapi.json
@@ -1320,7 +1320,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/User"
+                  "$ref": "#/components/schemas/FrontEndUser"
                 }
               }
             }
--- a/docs/docs/api/openapi.json
+++ b/docs/docs/api/openapi.json
@@ -1320,7 +1320,7 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/FrontEndUser"
+                  "$ref": "#/components/schemas/User"
                 }
               }
             }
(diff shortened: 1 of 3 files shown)


fozziethebeat:
What is this workflow that's auto updating the docs? Can that be turned off?

rsandb:
I'll find a better spot to put the TOS and Privacy Policy if you don't mind. the footer on the dashboard just doesn't look right :)

rsandb:
I think allowing the user to return to the homepage would be a good option now that dashboard is located in the UserMenu

fozziethebeat:
I'd rather avoid having the user have to guess as to where they can find these links.  [Common Voice](https://commonvoice.mozilla.org/en) includes their footer on the dashboard page and it looks pretty nice.  The task page maybe doesn't need a footer (Common Voice doesn't include it) but the primary dashboard, messages, and leaderboard pages should.

I think we can fix our layouts by making sure the whole page scrolls rather than just the inner component which feels kind of weird.

fozziethebeat:
I am open to seeing the footer be more compact and simpler for these views.  But that should be a later PR.

AbdBarho (approved):

AbdBarho:
not sure why pre-commit is stuck?

fozziethebeat:
Pre-commit is stuck due to a doc workflow that injects changes to the documentation on pull requests.  PR #723 is fixing this.

yk (approved):
#723 is in, sorry for the inconvenience

PR #732, opened 2023-01-15
Supporting user pagination on the web admin view

### LINKED ISSUE #673: Support saving notes on admin decisions for users

fozziethebeat:
Closes #673 

This:
- Updates the `frontend_users` endpoint on the backend to use strict greater than and less than ordering when paginating.  This is to avoid receiving the user the client already has access to (the user used as the parameter's value).
- Updating the `UsersCell` component to paginate forward and backward using the display name cursor
- Adding a toast to inform the admin that no more users can be found when this occurs (and retaining the set of found users).
- Updating the web API routes to support paginating in both directions.

### DIFF
--- a/website/src/components/UsersCell.tsx
+++ b/website/src/components/UsersCell.tsx
@@ -62,17 +62,35 @@ const UsersCell = () => {
   });
 
   const toPreviousPage = () => {
-    setPagination({
-      cursor: users[0].display_name,
-      direction: "back",
-    });
+    if (users.length >= 0) {
+      setPagination({
+        cursor: users[0].display_name,
+        direction: "back",
+      });
+    } else {
+      toast({
+        title: "Can not paginate when no users are found",
+        status: "warning",
+        duration: 1000,
+        isClosable: true,
+      });
+    }
   };
 
   const toNextPage = () => {
-    setPagination({
-      cursor: users[users.length - 1].display_name,
-      direction: "forward",
-    });
+    if (users.length >= 0) {
+      setPagination({
+        cursor: users[users.length - 1].display_name,
+        direction: "forward",
+      });
+    } else {
+      toast({
+        title: "Can not paginate when no users are found",
+        status: "warning",
+        duration: 1000,
+        isClosable: true,
+      });
+    }
   };
 
   // Present users in a naive table.
--- a/website/src/components/UsersCell.tsx
+++ b/website/src/components/UsersCell.tsx
@@ -11,6 +11,7 @@ import {
   Th,
   Thead,
   Tr,
+  useToast,
 } from "@chakra-ui/react";
 import Link from "next/link";
 import { useState } from "react";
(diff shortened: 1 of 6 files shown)


website/src/components/UsersCell.tsx
```diff
@@ -11,33 +11,68 @@ import {
...
 import { useState } from "react";
 import { get } from "src/lib/api";
 import type { User } from "src/types/Users";
 import useSWR from "swr";
 
+interface Pagination {
+  /**
+   * The user's `display_name` used for pagination.
+   */
+  cursor: string;
```
AbdBarho, line 26: Now all users are coming from the real backend and the data is combined with our prisma db, correct? why are we using `display_name`? Is it a unique key for the user?

fozziethebeat: This is due to the backend API.  `/api/v1/frontend_users/` takes in a `gte` or a `le` field for pagination and it sorts (filters actually) based on `display_name`.  I had to read `backend/oasst_backend/user_repository.py` to verify this and hence why I also updated it to make the ordering non-inclusive.

AbdBarho: I see...

AbdBarho (approved):

yk:
@andreaskoepf what was the conclusion of the pagination discussion? I can live with how it is now, but would also be a good time to "do it right"

fozziethebeat:
> @andreaskoepf what was the conclusion of the pagination discussion? I can live with how it is now, but would also be a good time to "do it right"

The ideal way to paginate is to use a unique sequential field that's indexed or can be used for sorting.  In our case it could be the created at timestamp.  I've also seen write-ups about people doing it with a combined index of a UUID & a created at timestamp.  I think its also maybe possible to just do it with the UUID.  [this Prisma discussion](https://github.com/prisma/prisma/issues/2855) talks about using UUIDs + timestamps.

AbdBarho:
to add to what @fozziethebeat said,
where I work we use UUID v6/v7 which is both unique for a db primary key and lexicographical sorting == time sorting for queries

fozziethebeat:
> to add to what @fozziethebeat said, where I work we use UUID v6/v7 which is both unique for a db primary key and lexicographical sorting == time sorting for queries

Ah yes, the many versions of UUID.  I've also seen UUID v6/v7 recommended for these reasons.  What UUID version is the backend using right now?

website/src/components/UsersCell.tsx
```diff
@@ -11,33 +11,68 @@ import {
...
+      if (data.length === 0) {
+        toast({
+          title: "No more users",
+          status: "warning",
+          duration: 1000,
+          isClosable: true,
+        });
+        return;
+      }
+      setUsers(data);
```
notmd, line 60: I think we should use the `data` return by `useSWR` and avoid calling setState on success which may cause unnecessary rerender. I already saw we are calling setState in the `onSuccess` callback in many places.

website/src/components/UsersCell.tsx
```diff
@@ -11,33 +11,68 @@ import {
...
+    },
   });
 
   const toPreviousPage = () => {
-    setPageIndex(Math.max(0, pageIndex - 1));
+    setPagination({
+      cursor: users[0].display_name,
+      direction: "back",
+    });
```
notmd, line 68: Do we need to check if the `users` is empty?

website/src/components/UsersCell.tsx
```diff
@@ -11,33 +11,68 @@ import {
...
   };
 
   const toNextPage = () => {
-    setPageIndex(pageIndex + 1);
+    setPagination({
+      cursor: users[users.length - 1].display_name,
+      direction: "forward",
+    });
```
notmd, line 75: Same here.

website/src/components/UsersCell.tsx
fozziethebeat: In other places I agree we can probably not use `onSuccess` and should move away from it.  Here I want to specifically because I want to avoid updating the user state when data is empty.  Do you  happen to know if there's an alternative way to accomplish the same result?

website/src/components/UsersCell.tsx
fozziethebeat: Done.  I trigger a warning toast if we hit that condition.

website/src/components/UsersCell.tsx
fozziethebeat: Here as well.

yk:
> Ah yes, the many versions of UUID. I've also seen UUID v6/v7 recommended for these reasons. What UUID version is the backend using right now?

I think it's uuid4

yk (approved):
good from my side

website/src/components/UsersCell.tsx
notmd: This may be implemented using custom swr middleware or something. I think it's ok to merge this now. I can send a PR to refactor this later.

andreaskoepf:
> @andreaskoepf what was the conclusion of the pagination discussion? I can live with how it is now, but would also be a good time to "do it right"

`username` alone is (theoretically) not unique, (e.g. when `auth_method` wasn't specified and the unlikely case happens that the same username is reused by another auth-provider). The correct way would be to include a unique value like thePK (user uuid in this case) as secondary sort-key and let the api-clients send the uuid of the last seen row of the previous page-window with the next request - the backend then would have to return elements following in defined sort order.

website/src/components/UsersCell.tsx
fozziethebeat: that would be very appreciated!

PR #742, opened 2023-01-15
added root tokens and endpoint for adding api keys

yk:
- added root tokens
- added an admin api router for creating api clients
- changed the dummy api key mechanism to only support the fixed key 1234


yk:
Note: I did not provide a server default for the root token entry in the api client table, thus, we'll have to wipe DBs upon merging

backend/oasst_backend/api/deps.py
```diff
@@ -33,16 +35,44 @@ async def get_api_key(
...
-        logger.info(f"ANY_API_KEY missing, inserting api_key: {token}")
-        api_client = ApiClient(
-            id=ANY_API_KEY_ID,
-            api_key=token,
+        logger.info(f"ANY_API_KEY missing, inserting api_key: {DUMMY_API_KEY}")
+        api_client = create_api_client(
+            db=db,
+            added_by_root_token="1234",
+            api_key=DUMMY_API_KEY,
             description="ANY_API_KEY, random token",
```
andreaskoepf, line 76: "random token" could now be removed from the description

andreaskoepf (approved):
ok, now we need to clarify how the create_api key endpoint is called, e.g. manually via /docs ui, by a pythin-script (cmdline tool) or a special mode of the frontend when an api-key is missing in the config.

yk:
- I removed the tracked_by_root_token, we have no use for it currently
- the endpoint is just for super-admins to create new api keys. we can call it via curl or /docs

PR #748, opened 2023-01-15
Footer styling

rsandb:
Now within the dashboard item group and fits on mobile

### DIFF
--- a/website/src/components/Footer.tsx
+++ b/website/src/components/Footer.tsx
@@ -1,40 +1,27 @@
-import { useColorMode } from "@chakra-ui/react";
+import { Box, Divider } from "@chakra-ui/react";
 import Image from "next/image";
 import Link from "next/link";
 import { useMemo } from "react";
 
 export function Footer() {
-  const { colorMode } = useColorMode();
-  const bgColorClass = colorMode === "light" ? "bg-transparent" : "bg-gray-800";
-  const borderClass = colorMode === "light" ? "border-slate-200" : "border-transparent";
-
   return (
-    <footer className={bgColorClass}>
-      <div className={`flex mx-auto max-w-7xl justify-between border-t p-10 ${borderClass}`}>
-        <div className="flex items-center pr-8">
+    <footer>
+      <Divider />
+      <Box display="flex" gap="4" flexDir="column" alignItems="center" mt="10">
+        <Box display="flex" alignItems="center">
           <Link href="/" aria-label="Home" className="flex items-center">
-            <Image src="/images/logos/logo.svg" className="mx-auto object-fill" width="52" height="52" alt="logo" />
+            <Image src="/images/logos/logo.svg" className="mx-auto object-fill" width="48" height="48" alt="logo" />
           </Link>
-
-          <div className="ml-2">
-            <p className="text-base font-bold">Open Assistant</p>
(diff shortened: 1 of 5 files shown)


rsandb:
image

rsandb:
Before
image

After
image

Placed after the last element in the dashboard item container

website/src/components/Footer.tsx
```diff
@@ -1,40 +1,27 @@
...
-  const bgColorClass = colorMode === "light" ? "bg-transparent" : "bg-gray-800";
-  const borderClass = colorMode === "light" ? "border-slate-200" : "border-transparent";
-
   return (
-    <footer className={bgColorClass}>
-      <div className={`flex mx-auto max-w-7xl justify-between border-t p-10 ${borderClass}`}>
-        <div className="flex items-center pr-8">
-          <Link href="/" aria-label="Home" className="flex items-center">
-            <Image src="/images/logos/logo.svg" className="mx-auto object-fill" width="52" height="52" alt="logo" />
+    <footer>
```
fozziethebeat, line 8: This looks like it changes the footer for all views.  I actually like the footer for the landing page and for most other views.  If we're changing this, can we make a new component `SlimFooter`or somehow structure this so it renders differently depending on props?

website/src/components/Layout.tsx
```diff
@@ -53,8 +53,8 @@ export const getDashboardLayout = (page: React.ReactElement) => (
       ]}
     >
       {page}
+      <Footer />
```
fozziethebeat, line 56: This leaves the footer in a weird floating state as in the screenshot.  Can we get the footer to always be at the bottom?  Embedding it inside of the dashboard panel feels really weird.  Footers typically always live at the bottom below everything.

fozziethebeat (changes requested):
Visually this doesn't feel much cleaner to me.  I still prefer the footer being fixed at the bottom of every page.   Rather than changing all the footers and making the footer floating. could we just add a prop to make the footer shorter in some situations?

website/src/components/Layout.tsx
rsandb: The dashboard has two sections: a sticky sidebar and the content. If the footer was under both it would shrink the view of the content and sidebar which results in a less than ideal user experience. The reason it "floats" under the content is because it's positioned after the last object in the container. Once the container fills up with more widgets and the leaderboard is filled with the correct data it won't be floating anymore. It's just a consequence of the dashboard not being finished.

Think of how Chat GPT or instagram has their footer in the content container rather than under both their sidebar and content container. The footer doesn't necessarily have to be under every container and in a lot of cases isn't even shown in some parts of some apps.

website/src/components/Footer.tsx
rsandb: Sure thing!

rsandb:
> Visually this doesn't feel much cleaner to me.  I still prefer the footer being fixed at the bottom of every page.   Rather than changing all the footers and making the footer floating. could we just add a prop to make the footer shorter in some situations?

I can justify-between so that when there isn't enough content to push the footer to the bottom the container will push it

rsandb:
after
image

fozziethebeat (approved):

Issue #728, opened 2023-01-15
Hippocorpus dataset for Open Assistant


MightyAlex200:
As per @christophschuhmann's request, I'll be cleaning up and formatting the [Hippocorpus](https://huggingface.co/datasets/hippocorpus) dataset for training with the assistant.

Hippocorpus is a dataset of 6,854 English diary-like short stories about recalled and imagined events, collected through a crowdsourcing framework and paired with author demographics and other variables.

The plan is as follows: (correct me if I get any of this wrong)
1. Create a python script which
    1. Converts the Hippocorpus dataset to a Parquet file
    1. The Parquet file has columns "INSTRUCTION", "RESPONSE" and "SOURCE" where
        1. "INSTRUCTION" is a natural language instruction to produce the story given the `mainEvent`/`summary`, and potentially a random sentence from `story` to include
        1. "RESPONSE" is `story` prepended with a "Sure! Here's a story about `mainEvent`" or similar prefix.
        1. "SOURCE" is "Hippocorpus" and `AssignmentId`
    1. Exports the modified dataset with `row_group_size=100`
1. Check at least 100 samples of the exported dataset for quality and report findings
1. Once deemed high enough quality upload python script to repository and modified dataset to HF

ontocord:
Thank you! Looking forward to your contributions!

ontocord:
If the hippocampus dataset isn't already in the form of qa or instruction->response, you would need to convert the text to do this.

MightyAlex200:
I created a draft pull request with my code. #750 
Here is a sample of the current output of this script. It's worth nothing that there are still quite a few mistakes, but almost all of them come from the unreliability of the original data. Are the mistakes that are present of an acceptably low frequency or should I continue to whittle down the dataset and apply modifications to the data?
[output.csv](https://github.com/LAION-AI/Open-Assistant/files/10421697/output.csv)

MightyAlex200:
I've modified the script per Huu Nguyen's request and specifications.
Here is a sample of the new output: [output.csv](https://github.com/LAION-AI/Open-Assistant/files/10422259/output.csv)

ontocord:
The results look really good!

ontocord:
@MightyAlex200 would you be interested in doing this dataset? https://www.kaggle.com/datasets/elvis23/mental-health-conversational-data ?

christophschuhmann:
[...]

PR #750, opened 2023-01-16
Add Hippocorpus dataset script

MightyAlex200:
This adds the Hippocorpus dataset script from #728 
It's a notebook file which converts an existing Hippocorpus dataset into a Parquet file.
This PR should stay a draft until the code quality and output quality is confirmed.


yk (approved):
looks good to me, thank you. maybe will need some extra work in the future to make it into a handle-able input, but it's good we already have it like this

53jk1 (approved):

PR #751, opened 2023-01-16
use `useSWR` data instead calling `setState`

notmd:
I also added some global SWR config.
I didn't refactor the change in #732 cause using custom middleware is more complex and doesn't have any improvement in this case.

### DIFF
--- a/website/src/pages/api/auth/[...nextauth].ts
+++ b/website/src/pages/api/auth/[...nextauth].ts
@@ -44,6 +44,7 @@ if (boolean(process.env.DEBUG_LOGIN) || process.env.NODE_ENV === "development") 
         const user = {
           id: credentials.username,
           name: credentials.username,
+          role: "admin",
         };
         // save the user to the database
         await prisma.user.upsert({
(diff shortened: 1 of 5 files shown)


website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -44,6 +44,7 @@ if (boolean(process.env.DEBUG_LOGIN) || process.env.NODE_ENV === "development")
         const user = {
           id: credentials.username,
           name: credentials.username,
+          role: "admin",
```
Comment (notmd) [inline website/src/pages/api/auth/[...nextauth].ts:47]: This is temporal. I will refactor the debug login to allow select roles when login in later PR.

Comment (fozziethebeat) [inline website/src/pages/api/auth/[...nextauth].ts:47]: This seems unrelated.  Is it needed?  the `.env` is configured to set the `admin` debug user with the `admin` role.  We didn't document that anywhere but it should work.

fozziethebeat (changes requested):
Just one small change that I think is handled already by env variables.

fozziethebeat (approved):
Just saw comment.  Cleanup in a follow up PR is fine.

PR #756, opened 2023-01-16
Allow to select role when sigin in debug mode

notmd:


website/src/lib/auth.ts
```diff
@@ -1,11 +1,15 @@
 import type { NextApiRequest, NextApiResponse } from "next";
 import { getToken, JWT } from "next-auth/jwt";
+import { ElementOf } from "src/types/utils";
 
+export const roles = ["general", "admin", "banned"] as const;
+
+export type Role = ElementOf<typeof roles>;
```
fozziethebeat, line 7: Could this be put somewhere under `src/types`? That way the react components don't need to import stuff from `src/lib/auth.ts`

fozziethebeat (changes requested):

website/src/lib/auth.ts
notmd: How about put it in the `RoleSelect` file? The `roles` const is not being used anywhere except the `RoleSelect`. About the `Role` type, Typescript will remove the import when we use the `import type`  syntax, so we don't have to worry about tree shaking here.

fozziethebeat: Sure that works.

notmd: Done!

fozziethebeat (approved):

PR #758, opened 2023-01-16
Switch back to useEffect in FlaggableElement 

### LINKED ISSUE #753: Web on Dev: Can't flag second messages in `/messages`
When going to `https://web.dev.open-assistant.io/messages` and looking at the list of messages, clicking the report button works on the first message but the second message shows no options.  Each report buttons should display the same set of labels.

fozziethebeat:
Fixes #753 

We have to save the data from `useSWR` and then update via `useEffect`.  Otherwise the valid label set gets thrown away when different versions of `FlaggableElement` is rendered and the label list is empty.

### DIFF
--- a/website/src/components/FlaggableElement.tsx
+++ b/website/src/components/FlaggableElement.tsx
@@ -22,7 +22,7 @@ import {
 } from "@chakra-ui/react";
 import { QuestionMarkCircleIcon } from "@heroicons/react/20/solid";
 import clsx from "clsx";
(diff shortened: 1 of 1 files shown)


notmd:
Sorry for the regression. Didn't notice that the URL is static.

fozziethebeat:
No problem.  I keep forgetting how `useSWR` works myself.

AbdBarho (approved):

PR #768, opened 2023-01-16
Ensure labeling task requires answers before review stage

### LINKED ISSUE #763: Labeling Task: Not providing labels then reviewing and submitting enters error state
When getting a full labeling task it's possible to review something without labeling anything then submitting.  The backend is requiring some labels to be answered and then returns a 500.  At that poin the UI is broken and there's no way to resolve the issue.

For tasks with mandatory answers, we shouldn't switch to review mode until the answer is given.

fozziethebeat:
Fixes #763 

Fixing the labeling task is rather hard with how rare they are, so this does a few things to make future changes easier:
- Makes a minimal set of changes to get storybook working again (some typescript changes were required due to storybook issues).
- Moves a few components into folders where stories can go alongside the component.
- Create a story for `Task` to more easily test changes to it with fake data
- Require an answer for the multi-labeling task before going to review stage.  The initial state has an empty set of labels which triggers a failure in the backend.


website/src/lib/oasst_api_client.ts
```diff
@@ -16,7 +16,13 @@ export class OasstError {
...
 
 export class OasstApiClient {
-  constructor(private readonly oasstApiUrl: string, private readonly oasstApiKey: string) {}
+  oasstApiUrl: string;
+  oasstApiKey: string;
+
+  constructor(oasstApiUrl: string, oasstApiKey: string) {
+    this.oasstApiUrl = oasstApiUrl;
+    this.oasstApiKey = oasstApiKey;
+  }
```
notmd, line 25: Why do we need to change this? Both codes are the same right?

fozziethebeat: For reasons I haven't quite figured out, storybook breaks with `private readonly`.  With the previous code storybook stories can't present.

AbdBarho (approved):

website/src/lib/oasst_api_client.ts
AbdBarho: maybe because some of our storybook stories are `.jsx` instead of `.tsx`, and therefore typescript features are not supported?

PR #778, opened 2023-01-16
Scroll to top on submit

### LINKED ISSUE #762: Improve scrolling during task view
With long tasks it's possible to enter into strange scroll states:
 - When scrolling to the bottom and submitting, scroll view stays at the bottom
 - Can be hard to reach the top of the task instructions when scrolling down and updating to new task with long content

Not clear what a better scroll experience could be but some exploration would be appreciated.

AbdBarho:
~~The dashboard layout now scrolls through the entire page, not only in the small box for the layout.~~

~~This required changes to the sidebar, the mode toggle has been moved to the top, although it would make sense to have it somewhere else.~~

Scrolls back to top after submit, a bit jarring I admit, but solves the problem I hope.

Fixes #762

### DIFF
--- a/website/src/components/Tasks/Task/Task.tsx
+++ b/website/src/components/Tasks/Task/Task.tsx
@@ -27,6 +27,8 @@ export const Task = ({ frontendId, task, trigger, mutate }) => {
   const replyContent = useRef<TaskContent>(null);
   const [showUnchangedWarning, setShowUnchangedWarning] = useState(false);
 
+  const rootEl = useRef<HTMLDivElement>(null);
+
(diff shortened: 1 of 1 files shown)


rsandb:
I think this creates a larger problem to the user experience. The changes have been made to my footer PR which will fix the footer issue

fozziethebeat (commented):
It's probably worth checking what Riley has submitted to main. I think that also fixes most of the scroll problems (except for scrolling to the top).  I like the auto scroll to the top as it gets instructions front and center without having to search for them.  So this probably could be simplified to just the scroll to top change.

AbdBarho:
I changed it to only scroll to top, btw. I like the new footer, much less enterprise-y.

fozziethebeat (approved):

PR #779, opened 2023-01-16
website: Enable i18n internationalization and refactor index page

rjmacarthy:
Hey there, I noticed that this project is not using any form of internationalization.

This pull request adds the most basic implementation of i18n localization support to the React project and refactors the index page to use it. The following packages have been added: `next-i18next` and `react-i18next` to package.json.

This change improves maintainability as strings can be kept in one place.  It also offers the opportunity to introduce further locales in the future.

Please let me know if you have any questions or concerns and I'll be happy to help.

Thanks!


website/src/components/CallToAction.tsx
```diff
@@ -1,17 +1,31 @@
-import { useColorMode } from "@chakra-ui/react";
+import { Box, Link, Text, useColorMode } from "@chakra-ui/react";
+import Image from "next/image";
+import { useTranslation } from "next-i18next";
 import { useId } from "react";
 
+import discord from "./assets/discord.svg";
```
fozziethebeat, line 6: If we're updating this, I'd rather we use an icon from `react-icons`.  Can you use one of those and get rid of these custom local svgs?

website/src/components/AnimatedCircles/AnimatedCircles.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <stop stopColor={gradStopColor} />
+            <stop offset="1" stopColor={gradStopColor} stopOpacity="0" />
+          </linearGradient>
+        </defs>
+      </svg>
+    </Box>
+  );
+};
+
+export default AnimatedCircles;
```
fozziethebeat, line 54: With the way index works i think it's cleaner to do `export AnimatedCircles` here and then index is just
`export * from './AnimatedCircles'`

fozziethebeat (changes requested):
This is really great! To go further on some cleanup and simplification I have a few small requests.

In the future tho, could you break a PR like this up into two separate PRs:
1.  Adding i18n support
2. Pure refactoring and cleanup

i18n is a rather big improvement so I'd like to make sure it's as clean and simple and isolated from any other changes going on.

website/src/components/AnimatedCircles/AnimatedCircles.tsx
notmd: I think we should stay away from default export and barrel files completely. NextJS has some issues with tree-shaking barrel files when using with Typescript [https://github.com/vercel/next.js/issues/12557](https://github.com/vercel/next.js/issues/12557). I also feel barrel files make it harder to navigate code.

fozziethebeat: Does this NextJS problem occur for simple index.tsx files that export just one thing?  This looks like it causes a problem when we're barreling together multiple components from multiple files, but not a problem when it's a simple index.tsx that exports a single file's components.

notmd: It seems doesn't happen with barrel files that just export only one component. But having a barrel file will encourage contributors to use this pattern.

fozziethebeat: For the sake of this PR, let's leave this as is.  Let me make a discussion issue where we can hash out a good strategy.

rjmacarthy: Thank you for the review, I have removed the default export in favour of a named export.

website/src/components/CallToAction.tsx
rjmacarthy: Thank you for the suggestion I have updated the PR to use `react-icons` and have deleted the custom `svg` assets.

rjmacarthy:
> This is really great! To go further on some cleanup and simplification I have a few small requests.
> 
> In the future tho, could you break a PR like this up into two separate PRs:
> 
> 1. Adding i18n support
> 2. Pure refactoring and cleanup
> 
> i18n is a rather big improvement so I'd like to make sure it's as clean and simple and isolated from any other changes going on.

Sure thing, I will make sure to create smaller and more concise pull requests which are more suited towards the changes and leave refactoring for another day,.

fozziethebeat (approved):

PR #781, opened 2023-01-16
Auto text_frontend and adds Consensus rank to messages upon entering READY_FOR_SCORING

danielpatrickhug:
This pull request adds the following
- Updates message rank on consensus(given by ranked_pairs algorithm) ranking upon entering READY_FOR_SCORING state.
- adds auto main which automates 10 random tasks(labeling, ranking, reply generation) each time it's run. This is good for bringing the seed data into ready for scoring state. 
- A single message_tree json file for testing.
- moves ranking.py from scripts/postprocessing to backend/oasst_backend.utils.


backend/oasst_backend/tree_manager.py
```diff
@@ -587,6 +588,7 @@ def check_condition_for_ranking_state(self, message_tree_id: UUID) -> bool:
         self._enter_state(mts, message_tree_state.State.RANKING)
         return True
 
+    @managed_tx_method(CommitMode.COMMIT)
```
andreaskoepf, line 591: can this be removed here?

backend/oasst_backend/tree_manager.py
```diff
@@ -603,8 +605,34 @@ def check_condition_for_scoring_state(self, message_tree_id: UUID) -> bool:
...
+        self.update_message_ranks(rankings_by_message)
         return True
 
+    @managed_tx_method(CommitMode.COMMIT)
+    def update_message_ranks(self, rankings_by_message: Dict[int, int]) -> None:
+        for parent_msg_id, ranking in rankings_by_message.items():
+            unordered_ranking = []
+            for msg_reaction in ranking:
+                unordered_ranking.append(
+                    [msg_reaction.payload.payload.ranking, msg_reaction.payload.payload.ranked_message_ids]
```
andreaskoepf, line 617: Why do we use the `ranking` field here, IMO it should be completely ignored and not be read/processed? Everything we need should be in `ranked_message_ids` .. i.e. the ids of the messages as they were ordered by the user.

backend/oasst_backend/tree_manager.py
```diff
@@ -603,8 +605,34 @@ def check_condition_for_scoring_state(self, message_tree_id: UUID) -> bool:
...
+            for rank, uuid in zip(consensus, sorted_uuids[0]):
+                # set rank for each message_id for Message rows
+                msg = self.db.query(Message).filter(Message.id == uuid).one()
+                msg.rank = rank
+                self.db.add(msg)
+
+    def sort_uuid_rankings(self, data: List[Tuple[List[float], List[str]]]) -> List[Tuple[List[float], List[str]]]:
+        sorted_data = []
+        for rankings, uuids in data:
+            sorted_uuids = sorted(uuids)
```
andreaskoepf, line 631: This looks a bit strange, e.g. the order of the uuids specifies the preference of the user .. if we sort it we destroy this information.

andreaskoepf (changes requested):
The use of sorted_rankings is a bit unclear to me...

PR #785, opened 2023-01-17
Exclude certain users from leaderboard stats

akhil-datla:


akhil-datla:
When I try generating the alembic migration script, I get this error `ModuleNotFoundError: No module named 'oasst_shared'.`

backend/oasst_backend/user_stats_repository.py
```diff
@@ -39,6 +39,7 @@ def get_leaderboard(self, time_frame: UserStatsTimeFrame, limit: int = 100) -> L
             self.session.query(User.id.label("user_id"), User.username, User.auth_method, User.display_name, UserStats)
             .join(UserStats, User.id == UserStats.user_id)
             .filter(UserStats.time_frame == time_frame.value)
+            .filter(User.stats_enabled)
```
yk, line 42: does this really do the correct thing, or should it be `== True`?

backend/oasst_backend/models/user.py
```diff
@@ -23,6 +23,10 @@ class User(SQLModel, table=True):
     created_date: Optional[datetime] = Field(
         sa_column=sa.Column(sa.DateTime(), nullable=False, server_default=sa.func.current_timestamp())
     )
+    stats_enabled: bool = Field(
+        nullable=False, sa_column=sa.Column(sa.Boolean, nullable=False, server_default=sa.true())
```
andreaskoepf, line 27: probably the `nullable` is ignored when `sa_column` is specified, but no problem.

backend/oasst_backend/user_stats_repository.py
andreaskoepf: yes, I think single boolean var works. Problem is always that linter things `abc == True` is evil ... but `abc is True` doesn't work with sqlalchemy (!) .. so you could use `sa.true()` for example to make linter happy.
Multiple filter conditions (AND) can be specified on one `filter(conditionA, conditionB)` call with comma separated...

andreaskoepf (commented):
OK so far but but it currently does still compute RANK etc for the user, even when `stast_enabled == False`. That means the leaderboard would show non-continuous RANK numbers (e.g. with gaps for the hidden users).

akhil-datla:
@andreaskoepf I'll work on the task this week.

fozziethebeat (changes requested):
Why is the website `.env` file deleted?

AlexanderHOtt:
> Why is the website `.env` file deleted?

isnt `.env` in the .gitignore

@akhil-datla could rename it to `.env.example`

AlexanderHOtt:
or we could add
```
!website/.env
``` 
to the `.gitignore` file

fozziethebeat:
> or we could add
> 
> ```
> !website/.env
> ```
> 
> to the `.gitignore` file

We include a `.env` in the website directory with working values so devs don't have to go create them.  They aren't secret in any way.  I'd like to keep that file.

akhil-datla:
@fozziethebeat, the website CI is failing. However, all the other CI workflows pass.

fozziethebeat:
> @fozziethebeat, the website CI is failing. However, all the other CI workflows pass.

Please ignore that failing for now.  Its a problem we need to investigate but is unrelated to your changes.

fozziethebeat (approved):
Removing my blocking review.  I'll leave the rest to others.

backend/oasst_backend/user_stats_repository.py
```diff
@@ -244,14 +245,18 @@ def update_ranks(self, time_frame: UserStatsTimeFrame = None):
...
-        ) AS "rank", user_id, time_frame
+FROM (
+    SELECT ROW_NUMBER () OVER(
+        PARTITION BY time_frame
+        ORDER BY leader_score DESC, user_id
+    ) AS "rank", user_id, time_frame
     FROM user_stats
-    WHERE (:time_frame IS NULL OR time_frame = :time_frame)) AS r
+    WHERE (
+        SELECT "visible_in_public_leaderboards" FROM "user"
```
andreaskoepf, line 255: this should be a join

andreaskoepf:
replaced by #961

PR #771, opened 2023-01-16
migrate `formik` to `react-hook-form`

notmd:
close #738 

This PR only changes forms that currently use `formik`. I will migrate all other forms that we are manually handling to use `react-hook-form` in the next PR.

PR #787, opened 2023-01-17
dashboard shadow fix

rsandb:
There was a problem where the shadow of components would get cut off by the container on large screens

before
image

after
image


notmd:
I think we should start using the `Card` component from `CharkaUI` for these kinds of components. I already do some initial theme setup in #771

fozziethebeat (approved):
This is good as is.

But I agree with @notmd 

Before we start making a ton of small styling fixes I'd prefer we do the following:
* Add storybook stories for all of our visual components so its easy to visually test them in isolation
* Replace our custom uses of `Box` with a Chakra `Card` that uses the new custom styling that @notmd put together (with some extensions as needed).
* Factor out any custom visual components we need and make them much  more re-usable

Then once we've cleaned up and refactored things, we can go do more visual touch ups.   Right now things are visually functional and don't need too much more tweaking.

PR #788, opened 2023-01-17
convert header to chakra components

rsandb:
also added z-index changes


rsandb:
sorry for the rapid commits, realized the mistake after

fozziethebeat:
Can you look into what breaks the build?

rsandb:
> Can you look into what breaks the build?

just did. should be good now. forgot to remove a z-index on the usermenu that shouldnt have been there

fozziethebeat (approved):

PR #789, opened 2023-01-17
style 500 page

rsandb:


website/src/components/EmptyState.tsx
```diff
@@ -32,3 +32,9 @@ export const TaskEmptyState = () => {
 export const PageEmptyState = () => {
   return <EmptyState text="Sorry, the page you are looking for does not exist." icon={FiAlertTriangle} />;
 };
+
+export const ServerEmptyState = () => {
+  return (
+    <EmptyState text="Sorry, we encountered a server error. We're not sure what went wrong." icon={FiAlertTriangle} />
```
fozziethebeat, line 38: Does this need to be declared here? It's only used in `500.tsx`? I think it'd be cleaner to skip making a new component in this file and just use `EmptyState` inside of `500.tsx`

fozziethebeat (changes requested):

website/src/components/EmptyState.tsx
rsandb: I was just following how the file was set up, each page that uses the EmptyState component declares the text in the EmptyState file. I can change that though if you'd like

fozziethebeat: I'd have to check if these other definitions are used in multiple places.  If it's used in multiple places it's fine but while it's just in one location and specific to the 500 page, we should just use it in 500 rather than create a one off component.

rsandb: I believe they're all page-specific. The 404 is only used on the 404 and the empty state is used on tasks

fozziethebeat: I will later clean those usages up later.  For this PR, please just embed the content in the page itself :)

fozziethebeat (approved):
I see the end2end test fails but that is probably due to a change I made.  Approved and will merge.

PR #791, opened 2023-01-17
use card component in dashboard + text theme

rsandb:
per @notmd 's recommendation

### DIFF
--- a/website/src/components/Dashboard/LeaderboardTable.tsx
+++ b/website/src/components/Dashboard/LeaderboardTable.tsx
@@ -1,26 +1,32 @@
-import { Card, CardBody, Flex, Link, Text } from "@chakra-ui/react";
+import { Box, Link, Text, useColorModeValue } from "@chakra-ui/react";
 import NextLink from "next/link";
 import { LeaderboardGridCell } from "src/components/LeaderboardGridCell";
 import { LeaderboardTimeFrame } from "src/types/Leaderboard";
 
 export function LeaderboardTable() {
+  const backgroundColor = useColorModeValue("white", "gray.700");
+  const accentColor = useColorModeValue("gray.200", "gray.900");
   return (
     <main className="h-fit col-span-3">
-      <Flex direction="column" gap="4">
-        <Flex alignItems="end" justifyContent="space-between">
-          <Text variant="h1">Top 5 Contributors Today</Text>
+      <div className="flex flex-col gap-4">
+        <div className="flex items-end justify-between">
+          <Text className="text-2xl font-bold">Top 5 Contributors Today</Text>
           <Link as={NextLink} href="/leaderboard" _hover={{ textDecoration: "none" }}>
             <Text color="blue.400" className="text-sm font-bold">
               View All -&gt;
             </Text>
           </Link>
-        </Flex>
-        <Card>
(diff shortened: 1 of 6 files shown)


website/src/styles/Theme/components/Card.ts
```diff
@@ -9,11 +9,10 @@ export const cardTheme = defineMultiStyleConfig({
     return {
       container: {
         backgroundColor: isLightMode ? "white" : "gray.700",
+        p: "6",
```
notmd, line 12: Do we really want this behavior? Personally, I feel using the extra `CardBody` is quite verbose. But I'm seeing many other frameworks use it as well

AbdBarho: maybe to be more responsive we can use the values `[2, 3, 4, 5, 6]` for all break points?

fozziethebeat: @rsandb can you comment on these suggestions?

notmd: I prefer the response padding values on the `CardBody` component option. If we want padding on the `Card` component we may need to check if there are any usage of `CardBody` to avoid duplicate padding or set the padding to 0 cause Charka already has default padding in the `CardBody` component.

notmd: @rsandb So let's add the padding Card component and reset the padding of CardHeader, Body, and Footer

fozziethebeat: Will approve after @notmd's request has been implemented.

rsandb:
Sorry for the holdup! I've been busy with work and personal matters

website/src/styles/Theme/components/Card.ts
rsandb: > @rsandb So let's add the padding Card component and reset the padding of CardHeader, Body, and Footer

I went with your first suggestion so Divider elements don't get shrunk inside the card element

fozziethebeat (approved):
Approve! But looks like there's a new small merge conflict.

website/src/components/Dashboard/LeaderboardTable.tsx
```diff
@@ -1,32 +1,26 @@
-import { Box, Link, Text, useColorModeValue } from "@chakra-ui/react";
+import { Card, CardBody, Flex, Link, Text } from "@chakra-ui/react";
```
fozziethebeat, line 1: I'd suggest dropping the changes to this file.  I think it's changed a lot and has the relevant changes you're working on already.

fozziethebeat (approved):

website/src/components/Dashboard/LeaderboardTable.tsx
rsandb: Is that in a different PR? The changes build off the current version and this just swaps native elements to Chakra components

fozziethebeat: Yeah that just went in with some changes to various table components.

fozziethebeat:
Unfortunately i'm going to close this.  We did quite a bit of refactoring to these components and I think this change has gotten a bit too out of date to refurbish.

PR #800, opened 2023-01-17
Guarantee all users get a name

### LINKED ISSUE #795: E-mail address of users who signed up via E-mail shown in leaderboad
When users sign up via E-mail and don't change their username via account settings their E-mail address becomes their display name and is shown in the leaderboard.

Possible mitigations:
- require users to set their username when they first sign-in
- create a synthetic name similar to how [docker automatically names things](https://github.com/moby/moby/blob/master/pkg/namesgenerator/names-generator.go)
- hide parts of the e-mail address in the leaderboard `a*@p*.com` (unesthetic)
- use UUID as username (unesthetic)
- requires backend changes: set display name to empty string or null (not sure if possible) and let backend choose a synthetic username or display the UUID

fozziethebeat:
Fixes #795 

When a new user is created via email, we give them a random generated username.

When communicating with the backend we always use the user name instead of the email to retrain privacy.

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -45,6 +45,7 @@
         "sharp": "^0.31.3",
         "swr": "^2.0.0",
         "tailwindcss": "^3.2.4",
+        "unique-username-generator": "^1.1.3",
         "use-debounce": "^9.0.2"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


andreaskoepf (approved):
ok, looks good!

PR #803, opened 2023-01-17
check my component feedback button

sir-braz:


fozziethebeat:
Thank you for offering a PR, before we review it, can you add more description regarding what this component is intending on doing and how to use the data?  If possible, it'd be even better to first file an issue and discuss it first if this doesn't address any open issues.

fozziethebeat:
Closing this as there's been no response for a week.

PR #805, opened 2023-01-17
Add api auth to user stat repo

danielpatrickhug:
Small pull request, to remove old get_dummy_api and replace it with new api_auth function.
it was breaking before. I'm working on user_stats_repo so just adding this quick fix.

### DIFF
--- a/backend/oasst_backend/user_stats_repository.py
+++ b/backend/oasst_backend/user_stats_repository.py
@@ -300,5 +300,3 @@ if __name__ == "__main__":
         usr = UserStatsRepository(db)
         usr.update_all_time_frames()
         db.commit()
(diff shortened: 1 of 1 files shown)


yk (commented):
are you sure it's always this exact uuid?

This entire `main` block is probably way more suited as a test, no?

andreaskoepf:
> This entire main block is probably way more suited as a test, no?

Yes, I added it for local testing with minimal deps during development, would normally go into a unit test. The UUID is a local id that makes only sense on the dev machine .. it can be removed. (I haven't done this since it is very helpful to have these mini test codes in case something breaks or for debugging.)

andreaskoepf (approved):
ok

Issue #794, opened 2023-01-17
message_id not correct in /api/set_label


melvinebenezer:
{"message_id":2,"label_map":{"spam":1,"fails_task":1},"text":"Do you know what damages the blood vessels that supply the retina?"}

the message_id should be a UUID like ba2f887a-c50d-4998-b172-1f4e43b8499f

PR #818, opened 2023-01-18
519_updating existing entry in text_labels table on message flag

melvinebenezer:
For every report from the frontend , a new entry was created in text_label table.

**Solution**
- Update existing record based on message_id in text_label table
- create new row if none exists

**Dependency**
- Blocked by https://github.com/LAION-AI/Open-Assistant/issues/794

**Test Evidence**
- SInce https://github.com/LAION-AI/Open-Assistant/issues/794#event-8242335186 is still open. Used Postman to recreate the scenario. 

**First Update**
Screenshot 2023-01-18 at 11 01 17 AM

**Second Update**
Screenshot 2023-01-18 at 11 00 47 AM

### DIFF
--- a/backend/oasst_backend/prompt_repository.py
+++ b/backend/oasst_backend/prompt_repository.py
@@ -368,7 +368,7 @@ class PromptRepository:
         self.db.add(reaction)
         return reaction
 
-    @managed_tx_method(CommitMode.FLUSH)
+    @managed_tx_method(CommitMode.COMMIT)
     def store_text_labels(self, text_labels: protocol_schema.TextLabels) -> Tuple[TextLabels, Task, Message]:
 
         valid_labels: Optional[list[str]] = None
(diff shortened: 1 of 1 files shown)


andreaskoepf:
Labels should not be replaced by random users (e.g. trolls). If updating labels should be supported then a user-reference needs to be stored in the text_labels rows and we can allow users to update their own labeling-submissions. Of course this has to be distinguished from the task-based label-responses by the user which went into the message-tree calculation and must not be updated.

melvinebenezer:
The user_id exits in the test_labels. Currently the labels are not linked to the task-based label-responses from the frontend. i.e it is possible to simply flag a message (the API is called to update the table) and skip the task. 
shall I update based on user_id and message_id ?

melvinebenezer:
@andreaskoepf As discussed with the new design , will make a new PR . Will close this one

PR #819, opened 2023-01-18
Cleaning up the username api route and ensuring the users name is fresh

### LINKED ISSUE #797: Username is reset when page is refreshed for E-mail users
Username is reset to E-Mail address on page refresh.

Reproduction: 
1. Set a username (DisplayName) via account settings 
2. Navigate to Dashboard (now correctly shows selected user name)
3. Refresh page by pressing F5
4. Username disappeared and E-Mail address is shown. When navigating to Account settings it says "No username".

fozziethebeat:
Fixes #797 

This:
- Restructures the `api/username` route to use the new role wrappers
- Update users based on their token ID instead of their email
- Ensure the session uses the latest username from the token
- Delete a check in the Cypress auth test that depends on showing the user's email (we now show an unpredictable username)


website/src/pages/api/username.tsx
```diff
@@ -1,20 +1,20 @@
...
     where: {
-      email: session.user.email,
+      id: token.sub,
     },
     data: {
       name: username,
     },
   });
-  res.json({ name: result.name });
-}
```
notmd, line 20: This file should be rename to `.ts` too

fozziethebeat: Very good point.  Autocompletion obscured that to me.  Done.

website/cypress/e2e/auth/signin.cy.ts
```diff
@@ -27,13 +27,6 @@ describe("signin flow", () => {
         });
       });
   });
-  it("shows the logged in users email address if logged in with email", () => {
-    const emailAddress = "user@example.com";
-    cy.signInWithEmail(emailAddress);
-    // The user will only see the email address if the window is wide enough, not technically required as even when hidden this will find it in the page.
-    cy.viewport(1920, 1000);
-    cy.contains('[data-cy="username"]', emailAddress);
```
notmd, line 35: Instead of deleting this test case, should we check that the `username` element exists?

fozziethebeat: That's done in the normal test on line 26.  With that, i figured this test doesn't do much that's useful.

AbdBarho (approved):

PR #820, opened 2023-01-18
Moving the page tests into pages directory to simplify tab completion

fozziethebeat:
Rather than putting tests in a `folder.test` folder, pair them with their pages.


fozziethebeat:
e2e test is broken since I am not including my other PR that fixes the test.

AbdBarho (approved):

PR #821, opened 2023-01-18
Allow to filter `user` by `display_name`

### LINKED ISSUE #766: Admin Enhancement: Make it easier to find specific users
The admin view can be improved in two ways:
- More efficient pagination using UUID based cursors
- Add a search bar to search for a user based on their display name (or prefix).  This will reduce search time when there's over 100 users registered.

notmd:
close #766
~~DO NOT MERGE, depend on #851~~ It's ready now.
- Implement a new `DataTable` component. I implement our own `filter` and `paginate` logic cause the built-in doesn't really useful when fetching data from server side and is inflexible.
- Use `react-table` v8. I plan to migrate other components currently using v7 to v8 when I complete #766.
- Allow searching for the user by `display_name`.
- ~~Use UUID based cursor.~~ It now provided by the backend.
- ~~Fix a bug when querying the user in the backward direction.~~

~~Note that this PR doesn't support filtering users by prefix and the `api/v1/frontend_users/by_display_name` endpoint doesn't support pagination yet. I'm going to address this in the next PR.~~
Here is the new UI

image


website/src/pages/api/admin/users.ts
```diff
@@ -1,11 +1,12 @@
 import { withRole } from "src/lib/auth";
 import { oasstApiClient } from "src/lib/oasst_api_client";
 import prisma from "src/lib/prismadb";
+import { BackendUser } from "src/types/Users";
 
 /**
  * The number of users to fetch in a single request.  Could later be a query parameter.
  */
-const PAGE_SIZE = 20;
+const PAGE_SIZE = 1;
```
fozziethebeat, line 9: Is this part intentional? Or should we switch it back to 20?

fozziethebeat (changes requested):

fozziethebeat:
I like the look of the new table!

website/src/pages/api/admin/users.ts
notmd: Sorry, I was debugging the pagination. Fixed with latest commit

backend/oasst_backend/user_repository.py
```diff
@@ -161,10 +161,10 @@ def query_users(
         users = users.order_by(User.display_name)
 
         if gt:
-            users = users.filter(User.display_name > gt)
+            users = users.filter(User.id > gt)
```
andreaskoepf, line 164: Not sure what the intention here is, but it doesn't look right.

notmd: I thought we want to cursor based on the user id?

notmd: Is the user id column monotonically increasing?

fozziethebeat: UUID V4 isn't guaranteed to be monotonically increasing.  I think only UUID V6 or V7 have that property.  V4 is random in it's ordering.

fozziethebeat (approved):

website/src/components/DataTable.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <Thead>
+              {getHeaderGroups().map((headerGroup) => (
+                <Tr key={headerGroup.id}>
+                  {headerGroup.headers.map((header) => (
+                    <Th key={header.id}>
+                      <Box display="flex" alignItems="center">
+                        {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
+                        {(header.column.columnDef as DataTableColumnDef<T>).filterable && (
+                          <FilterModal
+                            value={filterValues.find((value) => value.id === header.id)?.value ?? ""}
```
rjmacarthy, line 97: If `filterValues` was an object and not an array could we avoid using `find` and `findIndex`?

website/src/components/UserTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+  useSWR(
+    `/api/admin/users?direction=${pagination.direction}&cursor=${pagination.cursor}&display_name=${display_name}`,
+    get,
+    {
+      onSuccess: (data) => {
+        // When no more users can be found, trigger a toast to indicate why no
+        // changes have taken place.  We have to maintain a non-empty set of
+        // users otherwise we can't paginate using a cursor (since we've lost the
+        // cursor).
+        if (data.length === 0) {
```
rjmacarthy, line 76:
```suggestion
        if (!data.length) {
```

website/src/components/DataTable.tsx
rjmacarthy: Thank you,

I was mainly referring to a suggestion to use object for items and not an array which would enable memorable look up of filters.

```
const filters = {
  linkOperator: 'is',
  items: {
    1: {
      id: 1,
      columnField: 'rating',
      operatorValue: '>',
      value: '4',
    },
    2: {
      id: 2,
      columnField: 'isAdmin',
      operatorValue: 'is',
      value: 'true',
    },
  },
}

const getFilterById = (id) => filters.items[id]
```

backend/oasst_backend/user_repository.py
```diff
@@ -207,6 +207,7 @@ def query_users_ordered_by_display_name(
         limit: Optional[int] = 100,
         desc: bool = False,
     ) -> list[User]:
+
```
yk, line 210: this newline triggers my review as code owner. it's a very beautiful newline. I approve ;)

yk (approved):

fozziethebeat (approved):
Time to merge and see all the glory

PR #823, opened 2023-01-18
Convert update ranks in user_stat_repository to sqlalchemy

danielpatrickhug:
Pull request to convert the SQL text to SQL alchemy. This new code updates all time frames of users based on the leader score.

After running this you will notice the rank column in the user table is now updated where before it was null. 

There is a warning that appears in the terminal when running this. but from what I can tell this can be ignored. 
``` bash
Users/danielhug/Open-Assistant/backend/venv/lib/python3.10/site-packages/sqlmodel/orm/session.py:101: SAWarning: Evaluating non-mapped column expression 'anon_1.user_id' onto ORM instances; this is a deprecated use case.  Please make use of the actual mapped columns in ORM-evaluated UPDATE / DELETE expressions.
  return super().execute(  # type: ignore
/Users/danielhug/Open-Assistant/backend/venv/lib/python3.10/site-packages/sqlmodel/orm/session.py:101: SAWarning: Evaluating non-mapped column expression 'anon_1.time_frame' onto ORM instances; this is a deprecated use case.  Please make use of the actual mapped columns in ORM-evaluated UPDATE / DELETE expressions.
  return super().execute(  # type: ignore
/Users/danielhug/Open-Assistant/backend/venv/lib/python3.10/site-packages/sqlmodel/orm/session.py:101: SAWarning: Evaluating non-mapped column expression 'anon_1.rank' onto ORM instances; this is a deprecated use case.  Please make use of the actual mapped columns in ORM-evaluated UPDATE / DELETE expressions.
  return super().execute(  # type: ignore
 ```

### DIFF
--- a/backend/oasst_backend/user_stats_repository.py
+++ b/backend/oasst_backend/user_stats_repository.py
@@ -227,28 +227,30 @@ class UserStatsRepository:
         Update user_stats ranks. The persisted rank values allow to
         quickly the rank of a single user and to query nearby users.
         """
+        try:
+            #sql alchemy implementation
+            subquery = (
+                select([
+                    func.row_number()
+                    .over(partition_by=UserStats.time_frame, order_by=[UserStats.leader_score.desc(), UserStats.user_id])
+                    .label('rank'),
+                    UserStats.user_id,
+                    UserStats.time_frame
+                ])
+                .where(UserStats.time_frame == time_frame.value if time_frame is not None else None)
+                .alias()
+            )
 
-        #sql alchemy implementation
-        subquery = (
-            select([
-                func.row_number()
-                .over(partition_by=UserStats.time_frame, order_by=[UserStats.leader_score.desc(), UserStats.user_id])
-                .label('rank'),
-                UserStats.user_id,
-                UserStats.time_frame
-            ])
-            .where(UserStats.time_frame == time_frame.value if time_frame is not None else None)
-            .alias()
-        )
-
-        update_stmt = (
-            update(UserStats)
-            .where(UserStats.user_id == subquery.c.user_id)
-            .where(UserStats.time_frame == subquery.c.time_frame)
-            .values(rank= subquery.c.rank)
-        )
-        qry = self.session.execute(update_stmt)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: backend/oasst_backend/user_stats_repository.py [hunk 1]
```diff
@@ -226,35 +226,35 @@ def update_ranks(self, time_frame: UserStatsTimeFrame = None):
...
+
+            update_stmt = (
+                update(UserStats)
+                .where(UserStats.user_id == subquery.c.user_id)
+                .where(UserStats.time_frame == subquery.c.time_frame)
+                .values(rank=subquery.c.rank)
+            )
+            qry = self.session.execute(update_stmt)
+            logger.debug(f"pre_compute_ranks updated({time_frame=}) {qry.rowcount} rows.")
+        except Exception:
```
andreaskoepf, line 256: catch-all is not a good idea here, exceptions should bubble up ...

backend/oasst_backend/user_stats_repository.py
```diff
@@ -226,35 +226,35 @@ def update_ranks(self, time_frame: UserStatsTimeFrame = None):
...
+                        .over(
+                            partition_by=UserStats.time_frame,
+                            order_by=[UserStats.leader_score.desc(), UserStats.user_id],
+                        )
+                        .label("rank"),
+                        UserStats.user_id,
+                        UserStats.time_frame,
+                    ]
+                )
+                .where(UserStats.time_frame == time_frame.value if time_frame is not None else None)
```
andreaskoepf, line 244: Did you verify (e.g. by printing the query to inspect the generated sql) that `time_frame.value if time_frame is not None else None` is not making it into the SQL? .. otherwise I would recommend to build the query dynamically based on the value of `time_frame`.

andreaskoepf (changes requested):
Thanks for the sa conversion. The try-except should be removed. Regarding the warnings: Any chance to get rid of them?

backend/oasst_backend/user_stats_repository.py
danielpatrickhug: got it, i'll remove. thank you.

53jk1 (approved):

PR #825, opened 2023-01-18
Fix leaderboard sorting

Vechtomov:
User ranks with the same score don't correspond to their position in the leaderboard.
leaderboard

Also this solution matches index `ix_user_stats__timeframe__rank__user_id`


andreaskoepf (approved):
good catch, thanks!

Issue #829, opened 2023-01-18
Add configuration endpoint for trusted-api_clients


andreaskoepf:
Allow trusted api_clients to query the server configuration values.

Issue #822, opened 2023-01-18
Enforce 2000 character message size limit


andreaskoepf:
The backend should reject incoming text-messages (initial prompt, assistant-reply, prompter-reply) that exceed a maximum message size limit with an OasstError (define new code). The limit should be defined in the backend settings with a default value of 2000 characters.

oriyonay:
i can take this as a first issue :)

AbdBarho:
Is there anyway the website knows the value? maybe we can also enforce there, we can definitely hardcode 2000 at the beginning.

andreaskoepf:
> Is there anyway the website knows the value? maybe we can also enforce there, we can definitely hardcode 2000 at the beginning.

Not yet, but I can add a configuration endpoint for trusted_api clients.

andreaskoepf:
@oriyonay we will add the message size limit setting also to the `PublicSettings` model used for `/api/v1/backend_settings/public` .. see   #830

oriyonay:
@andreaskoepf thank you! will take a look in a few hours (after classes :))

PR #830, opened 2023-01-18
add GET /api/v1/admin/backend_settings/{full/public} endpoint for api-clients

andreaskoepf:
`/api/v1/admin/backend_settings/public` -> all clients
`/api/v1/admin/backend_settings/full` -> only accessible for trusted api clients

solves #829
related to #822


yk (approved):
lgtm, but probably a better way would be to have settings inherit from publicsettings and amend the private properties, then just specify publicsettings as response model, no conversion needed

andreaskoepf:
> lgtm, but probably a better way would be to have settings inherit from publicsettings and amend the private properties, then just specify publicsettings as response model, no conversion needed

I thought about that too .. but it gives less control over what should be handed out, in general I think exposing in internal settings object like it is done for the trusted-endoint should be an exception .. for the same reason that we normally don't directly return db-models but convert them into response-models which are sent back via json/http.

PR #834, opened 2023-01-19
add admin purge user function

andreaskoepf:
- database should be backed up before trying this in the production system


yk (approved):
lgtm. We'll need to come away from the raw sql statements eventually, those things are mighty scary

PR #839, opened 2023-01-19
PR: Create notebook to convert r/changemyview data

kayjay-is-here:
you can access the dataset here:
https://huggingface.co/datasets/kjl3080/OA_CMV_Arguments


andrewm4894:
@kayjay-is-here great - can you:

- [ ] add a open in colab button similar to in the [example notebook](https://github.com/LAION-AI/Open-Assistant/blob/main/notebooks/example/example.ipynb).
- [ ] run all cells so users can see expected sort of output when just reading the notebook in github.

kayjay-is-here:
I need to make some additional changes before PR

kayjay-is-here:
I'm stupid and didn't read the note about the "open in collab" button

andreaskoepf:
@yk this pr contains a 500kb parquet file. Should the parquet file be stored at HF and only the notebook here?

kayjay-is-here:
I think a 500kb one should be stored here as a test set, and the full one is stored on HF. I can remove it if you think that it's not a good idea.

yk:
any chance we can get it down to like 10 or 50kb? if not, it's still fine, but if it's possible to make it very small, it would save a bit of time of everyone doing `git clone`

kayjay-is-here:
No, but I can take the parquet file out if you want

andrewm4894:
@kayjay-is-here can you limit this to printing top 5 or so just so the wall of text is not so big for someone just browsing it on GH: 
image

andrewm4894:
@kayjay-is-here since `output. Parquet` is created by the notebook i think you can just remove it from this PR as not really needed to get the notebook to run from what i can see. 

So i think its just:

- [ ] only print 5 samples of `original_post_train`
- [ ] remove `output.parquet` file from the PR since last couple of cells will generate it anyway for the user. .

andrewm4894 (changes requested):
@kayjay-is-here since `output.parquet` is created by the notebook i think you can just remove it from this PR as not really needed to get the notebook to run from what i can see. 

So i think its just:

- [ ] only print 5 samples of `original_post_train`
- [ ] remove `output.parquet` file from the PR since last couple of cells will generate it anyway for the user.

andrewm4894:
@kayjay-is-here can you trim and reduce all the walls of text produced by any cells. The notebook is 1.4mb in size and won't actually render on my phone on GitHub since so much text. 

I'd suggest add some control like just printing 5 or a small sample of the data to any cell that prints lots of text as output.

andrewm4894 (approved):

PR #844, opened 2023-01-19
Send the most detailed user credentials to the backend

### LINKED ISSUE #827: auth_method is always shown as `local` for all users
In the admin dashboard in the `AUTH METHOD` column currently shows `local` for all users. For users that signed up via discord it should be `discord`. It is not clear to me what `AUTH ID` is - I would expect this to be the discord user id but that should be a number and it is showing there some alphanumeric tokens.

fozziethebeat:
Fixes #827 

This adds a new helper method that checks what linked accounts a user has.  If a linked account exists, we use those credentials when communicating with the backend rather than the web's local credentials.

### DIFF
--- a/website/cypress/contract/oasst_api_contract_tests.cy.ts
+++ b/website/cypress/contract/oasst_api_contract_tests.cy.ts
@@ -1,34 +1,27 @@
 import { OasstApiClient, OasstError } from "src/lib/oasst_api_client";
+import type { BackendUserCore } from "src/types/Users";
 
 describe("Contract test for Oasst API", function () {
   // Assumes this is running the mock server.
   const oasstApiClient = new OasstApiClient("http://localhost:8080", "test");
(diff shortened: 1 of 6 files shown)


AbdBarho (approved):
taking the "first" of an array is probably not the final solution right?

do we plan to merge all accounts together?

fozziethebeat:
> 

This is a problem we need to discuss at the next web meeting.  We're kind of making this harder for ourselves by having a web user table and a backend user table.  We should ideally have just one and always use the UUIDv6 created in that unified user table.  I've been wanting to investigate how to setup JWT and Oauth in the backend so we can totally throw away the web table and web API but that may be hard.

The other way to solve this is by other frontends always talking to the web to get user IDs and using that.

Either way we have a big mess of user IDs everywhere that we should solve soon.

PR #846, opened 2023-01-19
Soda dataset in an instruction format

sanagno:
Added soda dataset loading as preprocessed by @Huu to be closer to a user/assistant format.


theblackcat102 (commented):
looks good

theblackcat102 (approved):
approved

PR #851, opened 2023-01-19
Add keyset pagination for users ordered by `username` / `display_name`

andreaskoepf:
for keyset pagination pass lowest or highest values of last result-set:
`gt_id` & `gte_username` or
`lt_id` & `lte_username`

`/api/v1/users/by_username`

all parameters are optional:
```
api_client_id: str
gte_username: str
gt_id: uuid
lte_username: str
lt_id: uuid
auth_method: str
search_text: str
max_count: int
```

`/api/v1/users/by_display_name`

all parameters are optional:
```
api_client_id: str
gte_display_name: str
gt_id: uuid
lte_display_name: str
lt_id: uuid
auth_method: str
search_text: str
max_count: int
```


notmd:
Could this simplify to 1 endpoint and add an extra `sort_by` column and make default to `username` or `display_name`? Also, unify all `gt_*` and `lt_*` param into `gt` and `lt` too. The back end should return 2 extra fields, `next` and `previous`, which are the cursors encoded or encrypted by the backend itself. I'm feeling the front currently needs to specify the param and endpoint just to get users which is too much work here. Let's push more work to the backend instead.

yk (approved):
thank you

PR #852, opened 2023-01-19
website: Add localization to pages, header and footer

rjmacarthy:
Hey, this is a follow up PR for https://github.com/LAION-AI/Open-Assistant/pull/779 which includes the following changes:

- Add `i18n` provider for server side rendered pages using `common` namespace.
- Add localization to Header and Footer and associated components from `common` namespace.

Unfortunately its necessary to call `serverSideTranslations` on each page level component see https://github.com/i18next/next-i18next#serversidetranslations for more information.

Please let me know if you have any questions or concerns.


website/src/pages/terms-of-service.tsx
```diff
@@ -189,4 +190,10 @@ const TermsOfService = () => {
 
 TermsOfService.getLayout = getTransparentHeaderLayout;
 
+export const getStaticProps = async ({ locale }) => ({
+  props: {
+    ...(await serverSideTranslations(locale, ["common"])),
+  },
+});
+
```
AbdBarho, line 198: can you extract this into a function?

fozziethebeat: I agree, ideally this is a little simpler so that the same 6 lines aren't copied for every page.

fozziethebeat (changes requested):

website/src/pages/terms-of-service.tsx
rjmacarthy: Thanks for the review, I agree with you.  I referenced the instructions from `next-i18next` in the PR description which advises to add this to each page level component.  I  have refactored this to a export function which is only one line now for default namespace and common static props, please take a look. 

See also this related issue: https://github.com/i18next/next-i18next/issues/1215

fozziethebeat (approved):

PR #853, opened 2023-01-20
new SFT datasets

theblackcat102:
Added translation and additional datasets to the pool:

translation:
- WMT 2019
- TED Talk translation
- DiveMT : expert written translation dataset

safety:
- Prosocial dialogue in both dialogue generation and explanation format

others:
- @Rallio67 instruction tuning dataset **instruct_tuning**
- Other summarization dataset : debate_sum, tldr_news

WMT and TED translation can be a big ( few GB per pair ), so becareful not to include too many language pairs.


sanagno (approved):
Looks great, I also agree that we should upload datasets into the hub. Perhaps we can have a dataset_preprocess folder where we create the hub datasets and then we simply download the final versions.

Issue #747, opened 2023-01-15
Add user stats to the account page


AbdBarho:
From https://github.com/LAION-AI/Open-Assistant/pull/744 it is possible to get stats for every user, add a component to the account page to allow each logged-in user to see their stats.

Question:
* should add the ability to view other user's stats (from the leaderboard)?

MostlyGenius:
can you assign this issue to me?

fozziethebeat:
Common voice has something like the following on their user profile page.  I think this is a good first start.

<img width="1082" alt="Screenshot 2023-01-16 at 1 04 26 PM" src="https://user-images.githubusercontent.com/398211/212596328-784e2be1-2e92-4528-b5cd-0c4f3e54773b.png">

realsh9dy:
can you assign me this issue

AbdBarho:
@realsh9dy gladly!

cryptidv:
Any updates on this?

AbdBarho:
@cryptidv I started working on it

PR #858, opened 2023-01-20
Restyle account page

AbdBarho:
Related #747 

image

### DIFF
--- a/website/src/components/Survey/SurveyCard.tsx
+++ b/website/src/components/Survey/SurveyCard.tsx
@@ -1,22 +1,18 @@
 import { Box, BoxProps, useColorModeValue } from "@chakra-ui/react";
+import { PropsWithChildren } from "react";
 
-interface SurveyCardProps {
-  className?: string;
-  children: React.ReactNode;
-}
-
-export const SurveyCard = (props: SurveyCardProps) => {
+export const SurveyCard = (props: PropsWithChildren<{ className?: string }>) => {
   const backgroundColor = useColorModeValue("white", "gray.700");
 
   const BoxClasses: BoxProps = {
     gap: "2",
     borderRadius: "xl",
     shadow: "base",
-    className: "p-4 sm:p-6",
+    className: "p-4 sm:p-6 " + (props.className ?? ""),
   };
 
   return (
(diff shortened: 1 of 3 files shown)


website/src/components/Survey/SurveyCard.tsx
```diff
@@ -1,22 +1,18 @@
...
-export const SurveyCard = (props: SurveyCardProps) => {
+export const SurveyCard = (props: PropsWithChildren<{ className?: string }>) => {
   const backgroundColor = useColorModeValue("white", "gray.700");
 
   const BoxClasses: BoxProps = {
     gap: "2",
     borderRadius: "xl",
     shadow: "base",
-    className: "p-4 sm:p-6",
+    className: "p-4 sm:p-6 " + (props.className ?? ""),
```
notmd, line 11: You can use `clsx` here to simplify the code

website/src/pages/account/index.tsx
```diff
@@ -1,9 +1,11 @@
-import { Button } from "@chakra-ui/react";
+import { Button, Divider, Flex, Grid, Icon, Text } from "@chakra-ui/react";
 import Head from "next/head";
 import Link from "next/link";
 import { useSession } from "next-auth/react";
 import React from "react";
 export { getDefaultStaticProps as getStaticProps } from "src/lib/default_static_props";
+import { MdOutlineEdit } from "react-icons/md";
```
notmd, line 7: We are trying to use one type of icon. Can you switch to Feather icon here?

website/src/components/Survey/SurveyCard.tsx
AbdBarho: Of cousre!

website/src/pages/account/index.tsx
AbdBarho: I am not sure we talked about that.

Feather icon is not installed anyway, I will leave it to a different PR.

fozziethebeat: I've been using Font Awesome (`react-icons/fa`) but they don't don't have an outline edit icon.  I think this is okay for now.

Issue #854, opened 2023-01-20
Allow user to set their preferred language on web


fozziethebeat:
A user should be able to set their preferred language.  Once it's set, we should send that anytime the user requests a task or submits an answer so that the backend can ensure tasks fit within the user's requested language.

andreaskoepf:
I am making the necessary changes in the backend.

fozziethebeat:
After we fix #875 I'll be able to prepare a web PR.

We'll do this by:
1.  Adding a little language selector to the header
2. Updating the `NEXT_LOCALE` cookie with the value to keep it saved across page visits
3. Extract `NEXT_LOCALE` when fetching tasks or submitting answers.

This should work pretty smoothly and allow us to avoid having to store anything in a database.  It also seems to be the recommended solution by NextJS

PR #863, opened 2023-01-20
Add lang-tag based task selection (lang-separation)

andreaskoepf:
- `ConversationMessage`, `TextReplyToMessage` and `TaskRequest` now have `lang` property, if no value is provided 'en' is assumed
- POST `api/v1/tasks/availability` endpoint has a `lang` param
- default for lang column of message table  is now 'en' (was 'en-US')
- task selection in TreeManager is language-dependent


yk (approved):
really cool, thank you.
It may be worth to think about always considering a set of languages, instead of only one, i.e. a user could ask for a task in either of their languages, and a task could be appropriate for either of two languages (i.e. translate from one to the other, or some math quiz, etc.).

but up to you

andreaskoepf:
> It may be worth to think about always considering a set of languages, instead of only one, i.e. a user could ask for a task in either of their languages, and a task could be appropriate for either of two languages (i.e. translate from one to the other, or some math quiz, etc.).

For task selection that is a future option. For submission of the message I would suggest to use only a single language tag(even when multiple languages are mixed).  When the uses has more than one active language it would be necessary to specify the language used for a submission ... or to use the task's language as default.

Translation is a use-case, but a special one that won't even be solved by using a list of languages, e.g. like instruction: "Translate this text into Chinese:  ..." .. while the instruction-message would be in "en" the task could only be fulfilled by users with "en" + "zh-cn" or similar, so somehow the user would need to specify the desired language of the reply.

I think we should first try the simple solution of 'single language set per session" which will be an improvement for 95% of all use-cases.

yk:
sounds good

andreaskoepf:
link #854

PR #866, opened 2023-01-20
add users/cursor keyset_pagination endpoint

andreaskoepf:


notmd (approved):
Very clean, thank you.

backend/oasst_backend/api/v1/messages.py
```diff
@@ -16,38 +16,90 @@
...
+    max_count: Optional[int] = Query(10, gt=0, le=1000),
+    desc: Optional[bool] = False,
+    api_client: ApiClient = Depends(deps.get_api_client),
+    db: Session = Depends(deps.get_db),
+):
+    def split_cursor(x: str | None) -> tuple[datetime.datetime, UUID]:
+        if not x:
+            return None, None
+        f = x.split("$", maxsplit=1)
+        if len(f) == 2 and len(f[0]) == 36:
```
yk, line 69: other than `36`, maybe `len("some_constant")` or a comment why 36

backend/oasst_backend/api/v1/users.py
```diff
@@ -70,6 +72,71 @@ def get_users_ordered_by_display_name(
...
+    gt: Optional[str] = None,
+    sort_key: Optional[str] = Query("username", max_length=32),
+    max_count: Optional[int] = Query(100, gt=0, le=10000),
+    api_client_id: Optional[UUID] = None,
+    search_text: Optional[str] = None,
+    auth_method: Optional[str] = None,
+    api_client: ApiClient = Depends(deps.get_api_client),
+    db: Session = Depends(deps.get_db),
+):
+    def split_cursor(x: str | None) -> tuple[str, UUID]:
```
yk, line 87: duplicate code?

yk (approved):
thank you!

backend/oasst_backend/api/v1/users.py
andreaskoepf: one splits into str, uuid, the other in datetime, uuid ... datetime parsing is the part where the two functions differ .. I can give them different names to highlight this.

backend/oasst_backend/api/v1/messages.py
andreaskoepf: ok, 36 is the length of a full UUID converted with str() (with dashes) ... I could also validate with a regex ..

PR #869, opened 2023-01-20
Initial implementation of the inference system

yk:
This PR introduces:
- A server for coordination
- A minimal worker
- A text client

all building on redis lists to stream data as it is being produced

### DIFF
--- a/inference/server/main.py
+++ b/inference/server/main.py
@@ -1,6 +1,5 @@
 import asyncio
 import enum
-import random
 import uuid
 
 import fastapi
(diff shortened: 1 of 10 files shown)


inference/server/main.py
```diff
@@ -0,0 +1,193 @@
...
+            if (request := chat.pending_message_request) is not None:
+                if chat.message_request_state == MessageRequestState.pending:
+                    if request.compatible_with(worker_config):
+                        break
+        else:
+            logger.debug("No pending tasks")
+            await asyncio.sleep(1)
+            continue
+
+        chat.message_request_state = MessageRequestState.in_progress
```
andreaskoepf, line 171: if we have >1 "message-broker" instances and CHATS in db/redis then this "dequeue" operation will become a congestion point. One idea would be to define clear "configuration" tiers, e.g. based on GPU memory requirements and have independent task queues for them.

andreaskoepf (approved):
let's get the initial impl in.

PR #872, opened 2023-01-21
website: Switch to likert style labeling

othrayte:
This PR proposes to change the labeling task and messaging labeling to use a likert style scale of 5 options instead of a slider. The advantages include:
* it's clearer what to do
* the user has to think less about the value they are entering (i.e. 41% vs 44%)
* the labels are at the ends of the scale that they represent
* the interface is more friendly to touch screens and small screens
* there is consistency between the labeling task and the user initiated message flagging.

Things to note:
* This PR doesn't yet update the valid labels API to return the new format for labeling data and as such it is just in the website at the moment.
* We still in another PR need to reduce the options and make them specific to the type of message being labelled (backend).
* I still need to figure out how to write the cypress E2E test that clicks on one of the radio buttons for each label.
* We should review the content, the "opposite" labels are just the ones I came up with and have not really been reviewed.

image
image
image
image
image


website/cypress/e2e/tasks/random.cy.ts
```diff
@@ -44,44 +44,11 @@ describe("handles random tasks", () => {
...
-                        break;
-                      }
-                      case "radio": {
-                        // Clicking on the slider will set the value to about the middle where it clicks
-                        parent
-                          .get('[aria-roledescription="radio-button"]')
-                          .last()
-                          .click();
+                // TODO: Reimplement to click on a aria-roledescription="radio" in each incomplete aria-roledescription="radiogroup"
+                cy.get('[aria-roledescription="radiogroup"]');
```
AbdBarho, line 48: we can assign custom data properties for testing only, similar to what we had previously `data-cy="label-group-item"`

website/src/components/Buttons/LikertButtons.tsx
```diff
@@ -0,0 +1,45 @@
+import { Button, SimpleGrid } from "@chakra-ui/react";
+import { PropsWithChildren, ReactNode } from "react";
+
+export const LikertButtons = ({
+  isDisabled,
+  options,
+  value,
+  onChange,
+}: PropsWithChildren<{
+  isDisabled: boolean;
```
AbdBarho, line 10: is there a situation where we want to disable the buttons?

website/src/components/Buttons/LikertButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+}>) => {
+  return (
+    <SimpleGrid
+      aria-roledescription="radiogroup"
+      data-cy-complete={value !== null}
+      columns={options.length}
+      spacing={[1, 4]}
+    >
+      {options.map((option, idx) => {
+        const indexValue = idx / (options.length - 1);
```
AbdBarho, line 23: I am not really sure I understand what the purpose of this `indexValue` is

website/src/components/Buttons/LikertButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+            aria-checked={indexValue === value}
+            key={idx}
+            onClick={() => {
+              onChange(indexValue === value ? null : indexValue);
+            }}
+            isDisabled={isDisabled}
+            isActive={indexValue === value}
+            textOverflow="ellipsis"
+            paddingLeft="0"
+            paddingRight="0"
```
AbdBarho, line 36: `px='0'`?

website/src/components/FlaggableElement.tsx
```diff
@@ -1,152 +1,84 @@
...
-    report.label_values.forEach(({ label, checked, value }) => {
-      if (checked) {
-        label_map.set(label.name, value);
-      }
-    });
+    console.assert(valid_labels.length === values.length);
+    values
+      .map((value, idx) => {
+        return { labelName: valid_labels[idx].name, value };
+      })
```
AbdBarho, line 67: this `map` is redundant, you can get the name from the index in the `forEach` directly.

website/src/components/FlaggableElement.tsx
```diff
@@ -161,26 +93,23 @@ export const FlaggableElement = (props: FlaggableElementProps) => {
...
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader>Select one or more labels that apply.</ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            <LabelInputGroup
+              simple
+              labelIDs={valid_labels.map(({ name }) => {
+                return name;
+              })}
```
AbdBarho, line 106:
```
valid_labels.map(({ name }) => name)}
```

website/src/components/Survey/LabelInputGroup.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+interface LabelInfo {
+  zeroText: string;
+  oneText: string;
+  zeroDescription: string[];
+  oneDescription: string[];
+  inverted: boolean;
+}
+
+// This should be moved to the valid labels api endpoint
```
AbdBarho, line 23: I wonder how annoying it would be to localize this.

website/src/components/Survey/LabelInputGroup.tsx
```diff
@@ -0,0 +1,198 @@
...
+                {descriptionA.length > 0 ? <Explain explanation={descriptionA} /> : null}
+                <Spacer minWidth="1em" />
+                <Text textAlign="right">{textB}</Text>
+                {descriptionB.length > 0 ? <Explain explanation={descriptionB} /> : null}
+              </Flex>
+              <LikertButtons
+                isDisabled={!isEditable}
+                options={[
+                  // eslint-disable-next-line react/jsx-key
+                  <TbChevronsLeft />,
```
AbdBarho, line 175: you can just assign a key here, you don't need to deactivate eslint.

website/src/components/Survey/LabelInputGroup.tsx
```diff
@@ -0,0 +1,198 @@
...
+  return (
+    <Grid templateColumns={"minmax(min-content, 30em)"} rowGap={2}>
+      {labelIDs.map((labelId, idx) => {
+        const { zeroText, oneText, zeroDescription, oneDescription, inverted } = label_messages[labelId];
+
+        let textA = zeroText;
+        let textB = oneText;
+        let descriptionA = zeroDescription;
+        let descriptionB = oneDescription;
+        if (inverted) [textA, textB, descriptionA, descriptionB] = [textB, textA, descriptionB, descriptionA];
```
AbdBarho, line 159: I am not sure why do we need an `inverted` flag, can't we just switch the labels around in `label_messages`?

AbdBarho (commented):
Thank you!

One final thing which I would like to discuss is using a standard radio group with toggles ([like this](https://chakra-ui.com/docs/components/radio)) instead of buttons with `<<`, `<`, etc...

The former is well known and I think it signifies the intention with much less room for misinterpretation.

website/cypress/e2e/tasks/random.cy.ts
fozziethebeat: > 

Rather than radio buttons or new radio group buttons, we could use a Chakra [Slider](https://chakra-ui.com/docs/components/slider#discrete-sliders) with discrete options and [custom tics](https://chakra-ui.com/docs/components/slider#slider-with-custom-labels-and-marks)

othrayte: @fozziethebeat, would this still look similar or would it look like a slider again. I couldn't see a way that the custom ticks reproduce the same N options interface, Chakra's examples look like they don't change how the actual sliders work or appear.

website/src/components/Buttons/LikertButtons.tsx
othrayte: They are disabled when in review mode, in review mode we disable all the inputs.

website/src/components/Buttons/LikertButtons.tsx
othrayte: The likert buttons are a series of N buttons, the values passed in are normalised 0.0-1.0, the indexValue is the value from 0.0-1.0 represented by button i of N, the value that the button at this index represents/controls.

website/src/components/Survey/LabelInputGroup.tsx
othrayte: To localise it where it is now, or once it is moved to the API?

website/src/components/Survey/LabelInputGroup.tsx
othrayte: Inverting switches the appearance without changing the values assigned to the different labels. For some labels the backend expects 1 for the negative case and for others it expects 1 for the positive case, the names oneText and zeroText are tied to the meanings that the backend gives. Given we wanted to try having the positive options on the right some of them need to return 1 for the left option, to achieve this the inverted option was added. In the end it seemed clearer to have oneText and zeroText and swap them in one place here rather than have leftText and rightText and need to switch the text manually when we decide that the options should be swapped. Fundamentally it comes down to whether we want to show options in an order not tied to the values and whether the backend will change to put them in the desired order.

website/cypress/e2e/tasks/random.cy.ts
fozziethebeat: I'm not sure it's a great option but you can customize the slider appearance and option.  It still ultimately would be a slider however.

website/src/components/Buttons/LikertButtons.tsx
othrayte: In the end I decided the padded wasn't required. In general I find the px and similar short names quite cryptic as someone mostly familiar with traditional css, would paddingX have been acceptable in this case or are we trying to be consistent about using the short names?

website/src/components/FlaggableElement.tsx
othrayte: Yeah, not sure how I missed that, I'll blame organic development of the code :laughing:.

othrayte:
> Thank you!
> 
> One final thing which I would like to discuss is using a standard radio group with toggles ([like this](https://chakra-ui.com/docs/components/radio)) instead of buttons with `<<`, `<`, etc...
> 
> The former is well known and I think it signifies the intention with much less room for misinterpretation.

@AbdBarho, Do you mean with " (*) name   (  ) name2   (  ) name3 " or are you refering to [custom radio options](https://chakra-ui.com/docs/components/radio#custom-radio-buttons) from that page? I think the former has significant clarity issues is we show it on narrow phone screens or in the report message modal popup. The latter is effectively what we have at the moment. Initially I thought the custom appearance support would do more but when I went back to read it again during development I couldn't see how it helped, maybe I missed something.

website/cypress/e2e/tasks/random.cy.ts
othrayte: > we can assign custom data properties for testing only, similar to what we had previously `data-cy="label-group-item"`

I've made an attempt at this, but it is a little odd as you can't use data-cy on components that don't render to the dom and expect cypress to find them so I had to pass it down. For finding the options in the radio group I'm still using the aria role as that should work regardless of how we implement the actual radio buttons.

othrayte:
FYI we are discussion radio option appearance in discord. https://discordapp.com/channels/1055935572465700980/1065268675357179924/1066626112521764884

website/src/components/Survey/LabelInputGroup.tsx
AbdBarho: oof, we will see...

andreaskoepf:
Nice job! :-)

AbdBarho (approved):

PR #880, opened 2023-01-21
added message size limit of 2000 characters (issue 822)

oriyonay:
added a default message size limit (of 2000 characters) as per issue #822. the size limit is defined as a constant in config.py and referenced in the PublicSettings class in api/v1/admin.py. Messages that go over this limit will not go through and raise an OasstError code 1008 (TASK_MESSAGE_TOO_LONG).


andreaskoepf (approved):
looks good, please make sure pre-commit is happy so that we can merge it.

PR #881, opened 2023-01-22
Adding language preferences for the UI and tasks

### LINKED ISSUE #854: Allow user to set their preferred language on web

fozziethebeat:
Fixes #854 

This creates a new UI component that lets a user select their preferred language.  The selection will change both the UI display language via `next-i18next` and which language tasks we fetch.

When fetching tasks or answering them, we extract the user's language and pass that along to the backend.


AbdBarho (approved):
nice! I wonder if we could use this to translate all ranking tasks, instead of having the backend send the descriptions.

PR #882, opened 2023-01-22
Fix recent messages not showing

notmd:
The `auth_method` param is required.

### DIFF
--- a/website/src/pages/api/messages/user.ts
+++ b/website/src/pages/api/messages/user.ts
@@ -4,6 +4,7 @@ const handler = withoutRole("banned", async (req, res, token) => {
   //TODO: add params if needed
   const params = new URLSearchParams({
     username: token.sub,
+    auth_method: "local",
   });
 
   const messagesRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/messages?${params}`, {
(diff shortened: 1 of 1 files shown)


website/src/pages/api/messages/user.ts
```diff
@@ -4,6 +4,7 @@ const handler = withoutRole("banned", async (req, res, token) => {
   //TODO: add params if needed
   const params = new URLSearchParams({
     username: token.sub,
+    auth_method: "local",
```
fozziethebeat, line 7: Can we make use of `getBackendUserCore` (from `src/lib/users`) and use whatever `auth_method` that returns?  username and auth method are a little depending on if the user signed in with discord or not.

fozziethebeat (changes requested):

website/src/pages/api/messages/user.ts
```diff
@@ -1,9 +1,12 @@
 import { withoutRole } from "src/lib/auth";
+import { getBackendUserCore } from "src/lib/users";
 
 const handler = withoutRole("banned", async (req, res, token) => {
   //TODO: add params if needed
+  const user = await getBackendUserCore(token.sub);
   const params = new URLSearchParams({
     username: token.sub,
```
fozziethebeat, line 8: I forgot to mention, can you also use `user.id` instead of `token.sub`?

fozziethebeat (changes requested):

AbdBarho (approved):

fozziethebeat (approved):

PR #885, opened 2023-01-22
Fix prefix formatting

theblackcat102:
Move question and answer token to formatting.py and added in dataset phase instead of collator. This fix soda and prosocial dialogue which previously added the prefix as follow:

```
<human><prefix>Your prefix</prefix>prompt question<bot>answer text <eos token>
```

dataset with prefix (soda, prosocial_dialogue) should now be as follow :

```
<prefix>Your prefix</prefix><human>prompt question<bot>answer text <eos token>
```


model/supervised_finetuning/custom_datasets/formatting.py
```diff
@@ -0,0 +1,5 @@
+QA_SPECIAL_TOKENS = {"Question": "<human>", "Answer": "<bot>", "StartPrefix": "<prefix>", "EndPrefix": "</prefix>"}
+
+
+def format_pair(pair):
+    return "{} {} {}".format(QA_SPECIAL_TOKENS["Question"], pair[0], QA_SPECIAL_TOKENS["Answer"]), pair[1]
```
sanagno, line 5: Maybe we can make this into
-+ return "{}{}{}".format(QA_SPECIAL_TOKENS["Question"], pair[0], QA_SPECIAL_TOKENS["Answer"]), pair[1]

Not sure how all the tokenizer handle spaces for the models we have

sanagno (approved):
ProsocialDialogue and SODA should also have the format_parirs

theblackcat102:
@sanagno I added the prefix token in the preprocess phase, so there's no reason to add during get_item

model/supervised_finetuning/custom_datasets/formatting.py
theblackcat102: done

sanagno:
Thanks, looks great

PR #901, opened 2023-01-23
Send lang parameter the correct way

### LINKED ISSUE #898: Available tasks don't update when switching user-language
Tested on dev-instance after #890 was merged: It seems still to show the same options for all languages and when I click on one with German selected I get "Looks like no tasks were found." ...

AbdBarho:
Closes #898


andreaskoepf (approved):
ok, quick fix

fozziethebeat (approved):

Issue #896, opened 2023-01-23
Add 'lang' filter option to messages endpoints


andreaskoepf:
Add an optional 'lang' query param that filters messages based on language to the REST API endpoints that return a list of message, 

- `api/v1/messages` [messages.py#L18](https://github.com/LAION-AI/Open-Assistant/blob/274d03c5f0f75ba87102271384f20089ea6c0ef8/backend/oasst_backend/api/v1/messages.py#L18)
- `api/v1/messages/cursor` [messages.py#L51](https://github.com/LAION-AI/Open-Assistant/blob/274d03c5f0f75ba87102271384f20089ea6c0ef8/backend/oasst_backend/api/v1/messages.py#L51)
- `api/v1/users/{user-id}/messages` [users.py#L216](https://github.com/LAION-AI/Open-Assistant/blob/274d03c5f0f75ba87102271384f20089ea6c0ef8/backend/oasst_backend/api/v1/users.py#L216)
- `api/v1/users/{user-id}/messages/cursor` [users.py#L247](https://github.com/LAION-AI/Open-Assistant/blob/274d03c5f0f75ba87102271384f20089ea6c0ef8/backend/oasst_backend/api/v1/users.py#L247)

The primary function to modify is [query_messages_ordered_by_created_date()](https://github.com/LAION-AI/Open-Assistant/blob/274d03c5f0f75ba87102271384f20089ea6c0ef8/backend/oasst_backend/prompt_repository.py#L683).

If no lang query param is provided messages of all languages should be returned (e.g. default value None, behavior as it is right now).

j0rd1smit:
Sounds like fun. I would like to do it 😀

andreaskoepf:
@j0rd1smit nice, I assigned you!

kushalnl7:
Hi @j0rd1smit, let me know if anytime you drop out of this task. I would like to take over this. Thanks!

PR #902, opened 2023-01-23
add lang filter option to message endpoints

j0rd1smit:
# add lang filter option to message endpoints
Fixes issues #896

As descriped in the issues, I added a lang filter option to:
- `api/v1/messages`
- `api/v1/messages/cursor`
- `api/v1/users/{user-id}/messages`
- `api/v1/users/{user-id}/messages/cursor`

I saw that `query_messages_ordered_by_created_date()` is also used in `/api/v1/frontend_users/{auth_method}/{username}/messages` and ``/api/v1/frontend_users/{auth_method}/{username}/messages/cursor`. However, I did not add the lang filter to this endpoint since the issues did not specify this endpoint. So, I was unsure if this was desired or not. Let me know if you also want it there, it is an easy fix.

### DIFF
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -70,9 +70,9 @@ def query_frontend_user_messages(
     only_roots: bool = False,
     desc: bool = True,
     include_deleted: bool = False,
+    lang: Optional[str] = None,
     api_client: ApiClient = Depends(deps.get_api_client),
     db: Session = Depends(deps.get_db),
-    lang: Optional[str] = None,
 ):
     """
     Query frontend user messages.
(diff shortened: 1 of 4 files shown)


andreaskoepf (approved):
nice job. :-)

andreaskoepf:
> However, I did not add the lang filter to this endpoint since the issues did not specify this endpoint. So, I was unsure if this was desired or not. Let me know if you also want it there, it is an easy fix.

Yes, I think it would be useful to have the same lang query-param for the other places that you identified, too. If you have time to do it, please add it. Otherwise I will merge this PR in 24h.

j0rd1smit:
> If you have time to do it, please add it

Done 😀

andreaskoepf (approved):
lgtm!

Issue #857, opened 2023-01-20
Support OAuth and JWT in the backend


fozziethebeat:
This is a big issue.

When complete, the web frontend should be able to directly call the backend service using JSON Web Tokens (JWTs) and users should be able to complete an OAuth flow through the backend.  This would let us (in a separate issue) clean up the web service and remove the API routes completely.

To do this I propose the following steps:
- [x] Get JWT authentication working on the backend using the web generated JWTs.  This requires [HKDF](https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/) to encrypt and decrypt the tokens and sharing the current secret.
- [ ] Test out a few API calls from the web client directly to the backend using JWT.  This will require changing some sharing permissions for the token on the web side so it can be shared cross site.
- [x] Implement [Discord OAuth 2.0 Flow](https://discord.com/developers/docs/topics/oauth2) on the backend.  A user should be able to go through it and get a valid JWT sent from the backend.
- [ ] Implement some kind of email magic link login equivalent to Next Auth's
- [ ] Replicate the User and Account tables from the web database to the Backend's database
- [ ] Copy over data from the web database to the backend database.

fozziethebeat:
Step 1 should replicate [this](https://github.com/nextauthjs/next-auth/blob/b2450ef625b3873cb3d4f4b1b34926beefd4432c/packages/next-auth/src/jwt/index.ts#L119) next auth code

AbdBarho:
Ok just to get this straight: we want to move authentication and authorization entirely to the python backend?

occupytheweb:
@fozziethebeat instead of fleshing out an entire auth-n and auth-z implementation on the backend, why not offload this responsability to a dedicated IAM server, like [Keycloak](https://www.keycloak.org)?
This would basically require close to no implementation, since Keycloak covers most of the requirements out of the box.
It would handle the issuance of short-lived access tokens to authenticated clients, and ensure their validity when used. 
Securing the OA stack with Keycloak requires, in essence, that the client-adapters are configured on the backend and frontend side. 
The backend would use [this adapter](https://pypi.org/project/python-keycloak/), 
while for the SPA, we'd follow [this approach](https://mobihack.me/blog/2021-12-31-keycloak-nextjs/) or [this implementation](https://github.com/jplew/next-keycloak).
We would also use this [extension to provide Discord auth](https://www.keycloak.org/extensions.html#:~:text=Discord%20Identity%20Provider). 

[...]

PR #920, opened 2023-01-24
857 backend token decrypt

fozziethebeat:
First change for #857 

This adds routes and configurations to the backend such that it can decrypt web generated JSON Web Tokens using the JWE format:
- A new auth handler that gets the token and returns the decoded email
- FastAPI decryption using `jose` and `hkdf` encryption methods
- Settings that copy over the web's auth settings
- A demo call from the web client side directly to the backend to verify everything works
- Some commented out configuration that might be necessary for cross domain calls


backend/oasst_backend/api/v1/auth.py
```diff
@@ -0,0 +1,49 @@
...
+        salt=settings.AUTH_SALT,
+        info=settings.AUTH_INFO,
+    )
+    key = hkdf.derive(settings.AUTH_SECRET)
+    # Next we decrypt the JWE token.
+    payload = jwe.decrypt(token, key)
+    # Finally we have the real token JSON payload and can do whatever we want.
+    content = json.loads(payload)
+    email = content["email"]
+    return TokenData(email=email)
```
yk, line 43: here you could do `TokenData.parse_raw`

backend/oasst_backend/api/v1/auth.py
```diff
@@ -0,0 +1,49 @@
...
+
+oauth2_scheme = APIKeyCookie(name=settings.AUTH_COOKIE_NAME)
+
+
+class TokenData(BaseModel):
+    """
+    A minimal re-creation of the web's token type.  To be expanded later.
+    """
+
+    email: Union[str, None] = None
```
yk, line 22: I believe pydantic has email types, but depends if validation is an issue here

yk (commented):
looks good to me. I've left minor pydantic comments, otherwise really neat

backend/oasst_backend/api/v1/auth.py
fozziethebeat: Done.  Didn't know about this.  Cool that it also silently ignores additional fields that aren't used.

backend/oasst_backend/api/v1/auth.py
fozziethebeat: Nice! switched to `EmailStr` and installed the required sub-package.

backend/oasst_backend/api/v1/auth.py
```diff
@@ -0,0 +1,46 @@
...
+    TokenData.
+    """
+    # We first generate a key from the auth secret.
+    hkdf = HKDF(
+        algorithm=hashes.SHA256(),
+        length=settings.AUTH_LENGTH,
+        salt=settings.AUTH_SALT,
+        info=settings.AUTH_INFO,
+    )
+    key = hkdf.derive(settings.AUTH_SECRET)
```
AbdBarho, line 36: these are constant for the entire lifetime of the application, correct? can we move them out of the function?

fozziethebeat: I tried that.  HKDR didn't properly generate the key.  I'd like to keep this as is since I copied it from how NextAuthJS does their JWE encryption/decryption.

AbdBarho: ok, if it didn't work then there is some magic happening that I don't understand.

AbdBarho (approved):

fozziethebeat:
@yk Web review is in, if either you or @andreaskoepf can approve for the backend changes, I'll merge this in.

Also, I can apply the same changes to the inference server so we're setup there with user auth

yk (approved):

backend/oasst_backend/api/v1/auth.py
fozziethebeat: The cryptography library described this as dangerous like Dinosaurs with Laser guns.  So there's definitely some magic we should be careful about.

Issue #712, opened 2023-01-14
UnifiedQA Data -> Instructions


vale95ntino:
Follow-up issue to #560 

Expand on the notebook UnifiedQA (notebook/data-argumentation) and try to convert the Q-A pairs into clearer Instruction-Answer pairs.

Proposal for approach:
- Individual workflow for each dataset in UnifiedQA (that is not in xP3 - I don't mind doing xP3 in a second step if no one takes the task, but Id start with UnifiedQA)
- Create multiple templates for each dataset
- Generate the instruction- conversation pairs from the original QA pairs

Question:
- I think its important to have multiple templates for the same task to create a more diverse dataset. Is it an issue if the same QA pair is represented into several Instruction-Response pairs? Or should I try to have a 1-1 match? On one hand we get a larger dataset by having each QA pair mapped to multiple Instruction-Response pairs, but it would also be a very redundant dataset.

ontocord:
Agree on adding diversity! Thank you!

vale95ntino:
Hey guys, sorry that I havent gotten to this yet. Been quite sick but Im feeling better now so I should do a PR next week :)

ontocord:
Hope you feel better :)

pruksmhc:
Seems like the PR is mergerd. Can this be closed? @vale95ntino

vale95ntino:
Yes it can, thank you for noticing, I had forgotten to close it!

PR #922, opened 2023-01-24
Data/unified qa conversation

vale95ntino:
Based on Issue #712

As discussed, I converted each dataset from a QA pair to a conversation with the agent.

A couple templates are used for each dataset and the notebook takes a random template function, generates a conversation (sometimes Q-A, sometimes longer, like Q-A-Q-A-Q-A, etc) and saves everything to json files

Already ran pre-commit


andrewm4894 (approved):
Thanks!

53jk1 (approved):

PR #923, opened 2023-01-24
907 avoid duplicate labeling & ranking tasks

andreaskoepf:
- add new boolean debug option DEBUG_ALLOW_DUPLICATE_TASKS (default false)
- exclude tasks to which the current user already responded in tree_manager if DEBUG_ALLOW_DUPLICATE_TASKS is False


yk (approved):
crazy!

PR #926, opened 2023-01-25
Duplicate message reply content filter

danielpatrickhug:
Draft PR to add a check in handle interaction to see if a user has recent messages in the last 24 hours that are a duplicate of the most recent message task. raises an error if a duplicate is found.


backend/oasst_backend/prompt_repository.py
```diff
@@ -344,6 +344,26 @@ def insert_toxicity(self, message_id: UUID, model: str, score: float, label: str
         self.db.add(message_toxicity)
         return message_toxicity
 
+    def check_users_recent_replies_for_duplicates(self, task_interaction: protocol_schema.AnyInteraction) -> bool:
+        """Check if the user has a task of the same type with the same text in the last 24 hours."""
+        username = task_interaction.user.id
+        user = self.db.query(User).filter(User.username == username).first()
```
andreaskoepf, line 350: no need to query, the user is available in self.user_id (btw. username is not unique.. only the combination of username + auth_method, see `lookup_client_user()`)

backend/oasst_backend/prompt_repository.py
```diff
@@ -344,6 +344,26 @@ def insert_toxicity(self, message_id: UUID, model: str, score: float, label: str
...
+        user_id = user.id
+        logger.debug(f"Checking for duplicate tasks for user {user_id}")
+        # messages in the past 24 hours
+        messages = (
+            self.db.query(Message)
+            .filter(Message.user_id == user.id)
+            .order_by(Message.created_date.desc())
+            .filter(Message.created_date > datetime.now() - timedelta(days=1))
```
andreaskoepf, line 358: all time we use is utc + tz-aware ... please use our helper function `utcnow()` (utils.py).

andreaskoepf, line 358: the timespan in minutes could be part of the settings (config.py)

andreaskoepf (commented):
left minor comments

andreaskoepf:
replaced by #958

Issue #911, opened 2023-01-24
Reduce ability for users to register with fake email accounts


fozziethebeat:
This can be solved in two steps:
- [ ] Add an environment variable we can set during deployment that disables email login.
- [ ] Add a ReCaptcha when user is logging in via email

mashdragon:
Although registering with "fake email accounts" might truly be a serious problem, requiring all users to authenticate via Oauth using accounts strongly tied with their identity is a huge privacy drawback. For example, perhaps there are high-quality volunteers that prefer to use a secondary email (myself included) for logging in.

Perhaps you mean to disable email registration from new users? Then, at least, administrators could still manually create email-based accounts for known-trustworthy volunteers who wish to have them.

> Add a ReCaptcha when user is logging in via email

reCAPTCHA actively discriminates against VPN and Tor users by marking their correct solutions as invalid several times in a row and slowing down image reveals in a way that only hurts humans, not bots. Please consider hCaptcha instead; the host would even be able to earn some rewards for the captcha solving work its users are doing.

notmd:
Another alternative is [Cloudfare captcha](https://blog.cloudflare.com/turnstile-private-captcha-alternative/).

fozziethebeat:
This issue is to guard against a worst case scenario where the data collection platform is being flooded by malicious bots.

In that worst case scenario pretty much all good users will be negatively affected in someway.  Temporarily turning off email login access while dealing with a bot attack seems reasonable as long as it's temporary.

Regarding captchas that guard against attacks, I'm open to any solution.  hCaptcha or Cloudfare Captcha both seem like good alternatives.

If someone wants to setup a captcha that'd be awesome

notmd:
Let's go with Cloudfare. I will send a PR soon.

mashdragon:
> Let's go with Cloudfare. I will send a PR soon.

Cloudflare would be good except that it does not function very consistently over VPN or Tor. It gives me HTTP 400 responses and refreshes infinitely. I would be really sad if it took a lot of work for me to log in.

fozziethebeat:
Is there a reason you need to log into Open Assistant to contribute data via VPN or Tor?

notmd:
> > Let's go with Cloudfare. I will send a PR soon.
> 
> Cloudflare would be good except that it does not function very consistently over VPN or Tor. It gives me HTTP 400 responses and refreshes infinitely. I would be really sad if it took a lot of work for me to log in.
> 
[...]

PR #932, opened 2023-01-25
911 signin captcha

notmd:
Part of #911 
- Add Cloudflare captcha.


website/src/lib/captcha.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (action) {
+    if (cdata) {
+      return response.action === action && response.cdata === cdata;
+    }
+    return response.action === action;
+  }
+
+  return false;
+};
```
notmd, line 56: I'm unable to test this code cause with a demo key Cloudflare doesn't send action and cdata in response.

website/src/middleware.ts
```diff
@@ -14,5 +16,38 @@ export const config = {
...
+    }
+
+    const url = req.nextUrl.clone();
+    url.pathname = "/api/auth/invalid-captcha";
+
+    return NextResponse.redirect(url);
+  }
+
+  return withAuth(req);
+};
```
notmd, line 39: This is extremely hacky. I tried 2 other solutions but they did not work.
- `NextResponse.redirect('/auth/singin?errorCode=somecode')` cause next-auth `signIn` function doesn't follow redirect.
- `return NextResponse.json({url: '/auth/singin?errorCode=somecode'})` cause Nextjs doesn't allow to modify response body in middleware

.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
   "python.formatting.provider": "black",
-  "python.analysis.extraPaths": ["${workspaceFolder}/oasst-shared"]
+  "python.analysis.extraPaths": ["${workspaceFolder}/oasst-shared"],
+  "prettier.singleQuote": false
```
yk, line 4: isn't this the default value?

notmd: Yes but my prettier config in the local is different. Without this change prettier will use single quote.

notmd:
There is a small problem when the user fills in the email and clicks the sign-in button too fast, the captcha may not be loaded yet (it can take 2s depending on the internet connection). And the website will inform the captcha is invalid which may confuse the users because they don't see any captcha on the page. Here are a few solutions.
- Disable the button until the `onSuccess` callback gets called. But maybe there is a situation that the captcha never gets successful, and the user won't be able to log in.
- Check for the captcha response when the user clicks login and shows a toast.
- Switch to visible captcha.

website/src/lib/captcha.ts
fozziethebeat: Can you add that as a comment?

website/src/middleware.ts
fozziethebeat: did you try using the signIn callback within NextAuth? If you modify `src/pages/api/auth/[...[nextauth].ts` You can fill out the `signIn` call back.  Returning true means signin works.  Returning false rejects the signin.  So I think you could put very similar code in that callback when the user is using the email provider.  Not entirely sure how yo get the captcha data tho.

fozziethebeat (changes requested):
If you could, can you try the callback approach for captcha validation?

Also, for ensuring the captcha triggers I am a fan of
-  Disable email signin until captcha loads
- If loading fails for any reason, trigger a toast to tell the user to reload the page and try again

yk:
maybe I'm just overlooking this right now, but is it possible that we have a setting to circumvent (i.e. disable) the captcha, maybe via an environment variable?

notmd:
> maybe I'm just overlooking this right now, but is it possible that we have a setting to circumvent (i.e. disable) the captcha, maybe via an environment variable?

Sure, I will try to add it now.

53jk1 (approved):

notmd:
I have no idea why the tests only fail in CI, in local they still passing.

fozziethebeat (approved):
The web changes look good to me.  Does the local docker compose stack still work fine using these fake captcha credentials?

fozziethebeat:
Let me copy your branch and see if i can replicate this test failure.  The fact that everything fails is worrisome

notmd:
> The web changes look good to me. Does the local docker compose stack still work fine using these fake captcha credentials?

Yes, it still works.

> Let me copy your branch and see if i can replicate this test failure. The fact that everything fails is worrisome

All the tests failed because can't log in, which probably related to my changes

notmd:
@fozziethebeat  I managed to reproduce the test fails in the local by running a full build (docker compose up). Will try to fix the tests tonight.

notmd:
@fozziethebeat I have no idea to fix the tests so I disable them for now, lol. If you have time, please help me out.

fozziethebeat:
@yk can you do a final review of the non-web changes?  They seem to address your request of an env variable to disable the captcha

andreaskoepf (approved):
.vscode default OK

yk (approved):
lgtm, thank you for the adjustments

fozziethebeat:
Merging!

@notmd can you share in discord what env variables we have to setup on the deployment instances?  If you can generate them and send them to me privately I can add them to Github secrets

PR #933, opened 2023-01-25
Ranked pairs

MattAlexMiracle:
quick and dirty fix for tie braking in rankings (just pick the first one...)


andreaskoepf (approved):
thanks a lot for the quick fix!

PR #937, opened 2023-01-26
Add user emoji augmentation for message queries

andreaskoepf:
Add a join to message_emoji for all major message queries in PromptRepository to get a list of emojis that ware added by the current user. This allows the frontend to display the user which emojis were added by the user.

### DIFF
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -78,6 +78,7 @@ def query_frontend_user_messages(
     Query frontend user messages.
     """
     pr = PromptRepository(db, api_client)
+    pr.init_user(authmethod=auth_method, username=username)
     messages = pr.query_messages_ordered_by_created_date(
         auth_method=auth_method,
         username=username,
--- a/backend/oasst_backend/api/v1/frontend_users.py
+++ b/backend/oasst_backend/api/v1/frontend_users.py
@@ -77,8 +77,7 @@ def query_frontend_user_messages(
     """
     Query frontend user messages.
     """
-    pr = PromptRepository(db, api_client)
-    pr.init_user(auth_method=auth_method, username=username)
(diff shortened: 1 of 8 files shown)


backend/oasst_backend/api/v1/frontend_users.py
```diff
@@ -78,6 +78,7 @@ def query_frontend_user_messages(
     Query frontend user messages.
     """
     pr = PromptRepository(db, api_client)
+    pr.init_user(auth_method=auth_method, username=username)
```
yk, line 81: is `init_user` always coupled with creating a prompt repository? if yes, can we make it a part of PR's init function?

backend/oasst_backend/api/v1/utils.py
```diff
@@ -4,6 +4,9 @@
 from oasst_backend.models import Message
 from oasst_shared.schemas import protocol
 
+EMPTY_LIST = []
+EMPTY_DICT = {}
```
yk, line 8: dangerous because stateful. would rather just write `[]` and `{}` inline, even if we are "sure" we'll never mutate them

yk (approved):
thank you!

backend/oasst_backend/api/v1/frontend_users.py
andreaskoepf: it normally is but the call through init already is another way to specify the user, e.g. passing the protocol_user structure (including display_name but without the option to specify the user_id (UUID). The init_user() method is only used for the message query web-endpoints.

backend/oasst_backend/api/v1/utils.py
andreaskoepf: OK, I can remove it .. otherwise hundreds of thousands of empty objects may be created per request (only to have fallback objects). But if you think that's no problem I can remove these.

andreaskoepf: removed the globals

PR #943, opened 2023-01-26
857: Implement backend Discord authentication

olliestanley:
Addresses #857 

- Adds Account table to the backend to link external provider accounts to OA users
- Adds ability to authenticate via Discord to obtain a backend-generated JWT for a Discord-linked user

Does **not** add ability to link accounts yet.


fozziethebeat (commented):
I'm glad to see someone picking up where I'm slacking off now that I got a little busy with work.

I think this generally makes sense.  Before we add the login methods to the backend, I was initially planning on making sure we can share the website's JWT token with the backend so the backend could verify users via JWT.

This would let the website team then remove all our custom website API routes that proxy calls to the backend and instead call the backend API directly from the client side safely.

Then I wanted to figure out how to create the JWTs on the backend using both Discord and Email Magic Link type oauth methods.

I mainly thought that ordering would make sense since we ideally could port the website's auth table into backend's auth table.  Those auth tables are somewhat complex so I didn't want to do it too quickly.

What do you think?

olliestanley:
> I'm glad to see someone picking up where I'm slacking off now that I got a little busy with work.
> 
> I think this generally makes sense. Before we add the login methods to the backend, I was initially planning on making sure we can share the website's JWT token with the backend so the backend could verify users via JWT.
> 
> This would let the website team then remove all our custom website API routes that proxy calls to the backend and instead call the backend API directly from the client side safely.
> 
> Then I wanted to figure out how to create the JWTs on the backend using both Discord and Email Magic Link type oauth methods.
> 
> I mainly thought that ordering would make sense since we ideally could port the website's auth table into backend's auth table. Those auth tables are somewhat complex so I didn't want to do it too quickly.
> 
> What do you think?

This seems logical to me. To be honest I don't really know how the website works currently (w.r.t your mentions of website JWT tokens, website auth table). My approach here was going to be OAuth with Discord (and then Email) with the backend then generating a JWT. The backend `User` table is currently a bit of a mess so I was also planning to use a new table (`AuthenticatedUser`) but we could use the existing one with a bit of extra work

fozziethebeat:
> > I'm glad to see someone picking up where I'm slacking off now that I got a little busy with work.
> > I think this generally makes sense. Before we add the login methods to the backend, I was initially planning on making sure we can share the website's JWT token with the backend so the backend could verify users via JWT.
> > This would let the website team then remove all our custom website API routes that proxy calls to the backend and instead call the backend API directly from the client side safely.
> > Then I wanted to figure out how to create the JWTs on the backend using both Discord and Email Magic Link type oauth methods.
> > I mainly thought that ordering would make sense since we ideally could port the website's auth table into backend's auth table. Those auth tables are somewhat complex so I didn't want to do it too quickly.
> > What do you think?
> 
> This seems logical to me. To be honest I don't really know how the website works currently (w.r.t your mentions of website JWT tokens, website auth table). My approach here was going to be OAuth with Discord (and then Email) with the backend then generating a JWT. The backend `User` table is currently a bit of a mess so I was also planning to use a new table (`AuthenticatedUser`) but we could use the existing one with a bit of extra work

Okay cool, starting with Discord OAuth on the backend to generate a backend generated JWT makes sense.  Its possible we can skip trying to deal with the web's JWT token (that's turning out to create a lot of decryption problems in my branch).

For table structure., I think what NextAuthJS put together makes a lot of sense. There's a User table that has all the core fields for a user and then a linked Accounts table which includes any third party auth data (like from Discord).  You can see the schema in [website/prisma/schema.prisma](https://github.com/LAION-AI/Open-Assistant/blob/main/website/prisma/schema.prisma).  Replicating that would allow us to include Email, Discord, and later other OAuth providers like Github pretty naturally.

olliestanley:
> Note: I think I just got JWT decryption working in my branch. We can get and decrypt the web's created token [branch](https://github.com/LAION-AI/Open-Assistant/pull/920/files).
> 
> Whatever we do here, we should replicate JWT encryption and decryption. Since this is what NextAuth does it seems like reasonable defaults.

Ok sounds good. Tbh I had not noticed you already had a branch when I started this! Sounds like the best use of my time here is going to be looking at the Discord OAuth

fozziethebeat:
> > Note: I think I just got JWT decryption working in my branch. We can get and decrypt the web's created token [branch](https://github.com/LAION-AI/Open-Assistant/pull/920/files).
> > Whatever we do here, we should replicate JWT encryption and decryption. Since this is what NextAuth does it seems like reasonable defaults.
> 
> Ok sounds good. Tbh I had not noticed you already had a branch when I started this! Sounds like the best use of my time here is going to be looking at the Discord OAuth

That would be exceptionally helpful!

olliestanley:
Had a look into Discord OAuth this morning and just pushed some initial code. Needs a lot of cleanup so I'll be going back over it later today but if anyone has comments/suggestions feel free to jump in

53jk1 (approved):

backend/oasst_backend/api/v1/login.py
```diff
@@ -0,0 +1,68 @@
...
+from oasst_backend.models import Account
+from oasst_shared.exceptions.oasst_api_error import OasstError, OasstErrorCode
+from oasst_shared.schemas import protocol as protocol_schema
+from sqlmodel import Session
+from starlette.status import HTTP_401_UNAUTHORIZED
+
+router = APIRouter()
+
+# TODO: replace placeholder
+REDIRECT_URI = "http://localhost:8000/api/v1/callback_discord"
```
fozziethebeat, line 15: For callback URLS, I'd recommend the pattern `/api/v1/callback/{provider}`.

Then later we can drop in other provides as we develop them: `email`, `github` etc.

This seems to be the URL pattern NextAuthJS uses.

backend/oasst_backend/config.py
```diff
@@ -133,6 +133,9 @@ class Settings(BaseSettings):
     AUTH_SECRET: bytes = b"O/M2uIbGj+lDD2oyNa8ax4jEOJqCPJzO53UbWShmq98="
     AUTH_COOKIE_NAME: str = "next-auth.session-token"
 
+    DISCORD_CLIENT_ID: str = ""
```
fozziethebeat, line 136: Maybe prefix these with `AUTH_` to make clear they're auth specific settings.

fozziethebeat (approved):
OAuth stuff looks good enough for now.

backend/oasst_backend/api/v1/login.py
```diff
@@ -0,0 +1,74 @@
...
+@router.get("/callback/discord", response_model=protocol_schema.Token)
+def callback_discord(
+    auth_code: str,
+    request: Request,
+    db: Session = Depends(deps.get_db),
+):
+    redirect_uri = f"{get_callback_uri(request)}/discord"
+
+    # Exchange the auth code for an access token
+    token_response = requests.post(
```
andreaskoepf, line 31: Could we make these requests with async calls with `aiohttp` instead of requests?

olliestanley: Switched

andreaskoepf (approved):
OK, looks good .. Thanks Ollie!

PR #950, opened 2023-01-27
949_transaction error handling

melvinebenezer:
- No DB Rollback on FLUSH
- Appropriate Error Handling for other cases


backend/oasst_backend/utils/database_utils.py
```diff
@@ -46,17 +51,39 @@ def wrapped_f(self, *args, **kwargs):
...
+                    except OperationalError as e:
+                        if e.orig is not None and isinstance(
+                            e.orig, (SerializationFailure, DeadlockDetected, UniqueViolation, ExclusionViolation)
+                        ):
+                            logger.info(f"{type(e.orig)} Inner {e.orig.pgcode} {type(e.orig.pgcode)}")
+                            if auto_commit != CommitMode.COMMIT:
+                                logger.info(
+                                    "Since it is a flush or rollback, let the outer call handle the transaction retry"
+                                )
+                                # TODO: Is returning result ok ?
```
andreaskoepf, line 63: we can't return here, please re-raise exception in this case!

backend/oasst_backend/utils/database_utils.py
```diff
@@ -46,17 +51,39 @@ def wrapped_f(self, *args, **kwargs):
...
                         self.db.rollback()
                 raise OasstError(
                     "DATABASE_MAX_RETIRES_EXHAUSTED",
                     error_code=OasstErrorCode.DATABASE_MAX_RETRIES_EXHAUSTED,
                     http_status_code=HTTPStatus.SERVICE_UNAVAILABLE,
                 )
             except Exception as e:
-                logger.error("DB Rollback Failure")
-                raise e
+                logger.error(f"DB Failure {type(e)} {e}")
```
andreaskoepf, line 81: since we have here an additional catch-block maybe we should split the function into to branches .. one with retry (Commit-case) and the other without.

backend/oasst_backend/utils/database_utils.py
```diff
@@ -82,17 +109,39 @@ async def wrapped_f(self, *args, **kwargs):
...
                         return result
-                    except OperationalError:
-                        logger.info(f"Retry {i+1}/{num_retries} after possible DB concurrent update conflict.")
+                    except OperationalError as e:
+                        if e.orig is not None and isinstance(
+                            e.orig, (SerializationFailure, DeadlockDetected, UniqueViolation, ExclusionViolation)
+                        ):
+                            logger.info(f"{type(e.orig)} Inner {e.orig.pgcode} {type(e.orig.pgcode)}")
+                            if auto_commit != CommitMode.COMMIT:
+                                logger.info(
```
andreaskoepf, line 118: applies to all replications as well

andreaskoepf (changes requested):

backend/oasst_backend/utils/database_utils.py
```diff
@@ -34,28 +39,42 @@ def decorator(f):
...
+                        except PendingRollbackError as e:
+                            logger.info(f"{e}")
                             self.db.rollback()
-                        if isinstance(result, SQLModel):
-                            self.db.refresh(result)
-                        return result
-                    except OperationalError:
-                        logger.info(f"Retry {i+1}/{num_retries} after possible DB concurrent update conflict.")
+                        except OperationalError as e:
+                            if e.orig is not None and isinstance(
```
andreaskoepf, line 55: the else case for this if seems to be missing, e.g. when we cannot retry we should re-raise the exception

backend/oasst_backend/utils/database_utils.py
```diff
@@ -34,28 +39,42 @@ def decorator(f):
...
-                    "DATABASE_MAX_RETIRES_EXHAUSTED",
-                    error_code=OasstErrorCode.DATABASE_MAX_RETRIES_EXHAUSTED,
-                    http_status_code=HTTPStatus.SERVICE_UNAVAILABLE,
-                )
+                if isinstance(result, SQLModel):
+                    self.db.refresh(result)
+                return result
             except Exception as e:
-                logger.error("DB Rollback Failure")
+                logger.error(f"{str(e)}")
```
andreaskoepf, line 77: just `str(e)` would be shorter

oasst-shared/oasst_shared/exceptions/oasst_api_error.py
```diff
@@ -19,6 +19,7 @@ class OasstErrorCode(IntEnum):
     API_CLIENT_NOT_AUTHORIZED = 2
     ROOT_TOKEN_NOT_AUTHORIZED = 3
     DATABASE_MAX_RETRIES_EXHAUSTED = 4
+    DATABASE_OPERATION_ERROR = 5
```
andreaskoepf, line 22: this error-code seems not to be used anymore

andreaskoepf (approved):
ok nice, thanks a lot!

Issue #917, opened 2023-01-24
Simpify message feedback UI to Thumbs Up/Down


fozziethebeat:
Right now responding to arbitrary messages is fairly complicated, it's for reporting and has a long list of labels.  To simplify it and get cleaner more frequent data we should switch to a simple flow:
1.  A inviting review button
2. A simple thumbs up/down option
3. An ability to provide more detailed labeling only if desired (via a `say more` link).

othrayte:
If someone can assign me to this one I'll start the website side.

fozziethebeat:
When this is done we should make sure the user experience is better than what's described in #927

andreaskoepf:
> If someone can assign me to this one I'll start the website side.

Super cool. I merged the inital emoji PR. I guess we need the information about which emojis were set by the current user. Will try to add this later today.

Not sure where we would want to place the add-emoji functions ... some ideas here (only showing possible positions ... no design proposal ;-)):

![image](https://user-images.githubusercontent.com/9976399/214522037-a965dbe8-e2da-45e3-b5c7-1775a7884576.png)

andreaskoepf:
(probably we should add :clap: to the emoji set too)

mashdragon:
We might want to keep the emoji button away from the report button.

othrayte:
This is what I have so far: (keep in mind that this is a UI mock up and not yet hooked up to the API and I haven't yet aligned this with the API so I still need to add a way to add the rest of the emoji reactions.

The reactions will appear if any user has reacted with that emoji. Note this use has up voted with a thumbs up but not down voted with the thumbs down and noone has reacted with the other emoji. Clicking on a reaction would undo the reaction if the user had that reaction already, otherwise it will add to the current count.
![image](https://user-images.githubusercontent.com/241372/214786050-180c752e-e645-4f14-a469-4a497683840a.png)

The overflow menu is always present and adds a way to react, label, or report.
![image](https://user-images.githubusercontent.com/241372/214786096-64412d2f-9529-4617-aa2e-11e1b65d54e4.png)

Unrelated: I'm also thinking that we would add an option to the menu to open the message in a new tab. At the moment you can only go to the messages in the messages view and not from the tasks. We can add the ability to go to the task in a new task by clicking on it in a task but I don't think that is a good experience on mobile.

fozziethebeat:
I like how clean and simple this UI is.  It can easily be ignored but also looking at it I intuitively know what to do.

[...]

PR #952, opened 2023-01-27
Add emoji reactions and reporting for messages (website)

othrayte:
Implements web side of #917

Move labeling to message ... menu and add reporting and emoji reactions

This new menu appears to the right of the message, and floats down if it doesn't fit. The emoji reactions appear to the left of the menu button and float down with the button when there isn't enough space. The menu contains a list of emoji the user can react with (currently just thumbs up and down) formatted as a quick access grid.

The user can also click on an existing emoji to remove them from it or add them to its count, the emoji disappears when the count reaches 0.

Also added "open in new tab" option so that messages can be navigated to from tasks on mobile.

emoji1
emoji2

Notes:
 * There is a bug where the checked status for emojis is missing from the recent messages list. This appears to be due to a limitation of the backend were we need to provide a user to get emoji for, but doing so filters the list to only contain messages from that user.
 * We can add more emoji easily in future, we just need to pick ones that we have consistent icons for.
 * A future task will add some buttons to the report popup to allow the user to select labels to apply with the report (i.e. spam, contains PII etc)
 * The backend adds a flag emoji when the user files a report, the user can remove their flag emoji, but I don't think the backend does anything with it. The user doesn't currently see the flag emoji until they refresh or change another emoji (this is fixable but there are higher priority things I need to work on first),
 * Other users can currently click on other users flag emoji to also flag, but at the moment they don't get a popup to fill anything in.
 * Currently the emojis are updated when they are changed, this greatly simplifies the UI code but means that the user can technically see other people adding reactions when they do, as rare as that probably is.


website/src/components/Messages/LabelPopup.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Label</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <LabelInputGroup labelIDs={valid_labels.map(({ name }) => name)} onChange={setValues} />
+        </ModalBody>
+        <ModalFooter>
+          <Button colorScheme="blue" mr={3} onClick={submit}>
+            Submit
```
fozziethebeat, line 68: Any chance you can use the i18n stuff here?

website/src/components/Messages/MessageEmojiButton.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
fozziethebeat, line 1: Yay stories~

website/src/pages/api/report.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Adds a report for a message
+ *
+ */
+const handler = withoutRole("banned", async (req, res, token) => {
+  // Parse out the local message_id, and the interaction contents.
+  const { message_id, text } = req.body;
+
+  const interactionRes = await fetch(`${process.env.FASTAPI_URL}/api/v1/text_labels`, {
```
fozziethebeat, line 11: Can this go in the oasst library?

fozziethebeat (changes requested):
Nice! This is a lot of changes but it was pretty easy to review.  Please at least comment on the few comments I made and I'd like to wait for @AbdBarho to review since this is a lot.

website/src/components/Messages/LabelPopup.tsx
othrayte: Sure, I'll take a pass over it and see if I can find them all.

website/src/types/Conversation.ts
```diff
@@ -1,4 +1,16 @@
...
+  name: string;
+  count: number;
+}
+
+export interface MessageEmojis {
+  emojis: { [emoji: string]: number };
+  user_emojis: string[];
+}
+
+export interface Message extends MessageEmojis {
```
AbdBarho, line 13: the `Message` interface is what gets received from the backend, does the backend also send the emojis now?

if not, please create a separate types that combines message & emojis.

AbdBarho (commented):
Thank you, this was definitely a lot of work!!

website/src/pages/api/report.ts
othrayte: Done

website/src/components/Messages/LabelPopup.tsx
othrayte: Think I got them all.

website/src/components/Messages/MessageEmojiButton.tsx
```diff
@@ -0,0 +1,48 @@
+import { Button } from "@chakra-ui/react";
+import { BoxSelect, Flag, LucideProps, ThumbsDown, ThumbsUp } from "lucide-react";
+import { ReactElement } from "react";
+import { MessageEmoji } from "src/types/Conversation";
+
+type EmojiIconPurpose = "MINI_BUTTON" | "NORMAL";
```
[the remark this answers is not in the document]
othrayte: TBH that might depend on whether we need to add extra styling, at the moment it just changes the size, but the differentiation is between putting it in normal parts of the UI and putting it in the little buttons on the message. I envisioned that we might have other changes like color and stuff but haven't needed it yet.

andreaskoepf:
> * The backend adds a flag emoji when the user files a report, the user can remove their flag emoji, but I don't think the backend does anything with it. The user doesn't currently see the flag emoji until they refresh or change another emoji (this is fixable but there are higher priority things I need to work on first),

We could for example disallow directly setting the red-flag for ordinary users, so that it only would be set through reporting the message... users could then not un-set the flag on their own.

website/src/components/Messages/MessageEmojiButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+      onClick={onClick}
+      variant={checked ? "solid" : "ghost"}
+      colorScheme={checked ? "blue" : undefined}
+      size="sm"
+      height="1.6em"
+      minWidth={0}
+      padding="0"
+    >
+      {getEmojiIcon(emoji.name, "MINI_BUTTON")}
+      <span style={{ position: "relative", left: "-0.2em", marginRight: "0.25em" }}>{emoji.count}</span>
```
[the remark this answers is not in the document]
othrayte: Ok, those are new to me, I'll take a look.

AbdBarho (approved):

website/src/types/Conversation.ts
othrayte: Yep, we get emojis for all messages.

othrayte:
@AbdBarho, thanks so much for fixing those last issues 🙏. I had started on it but had to stop and have an early night as I had a splitting headache and nausea by that point.

PR #958, opened 2023-01-27
Duplicate message reply filter

danielpatrickhug:
PR continuation from DRAFT PR #926 
- changed user lookup to use lookup_client_user
- changed message query to use utcnow() and moved time in minutes variable to settings
- added a check in store_text reply to confirm there is no white space only messages.


backend/oasst_backend/prompt_repository.py
```diff
@@ -550,6 +556,30 @@ def fetch_message_tree(
             qry = qry.filter(not_(Message.deleted))
         return self._add_user_emojis_all(qry)
 
+    def check_users_recent_replies_for_duplicates(self, task_interaction: protocol_schema.AnyInteraction) -> bool:
+        """
+        Checks if the user has recently replied with the same text within a given time period.
+        """
+        user = self.user_repository.lookup_client_user(task_interaction.user)
+        user_id = user.id
```
andreaskoepf, line 564: we don't need an additional user lookup here .. e.g. simply use `self.user_id` in `PromptRepository` (technically it would be possible to pass a different user to the PromptRepository() ctor than the one used in the interaction .. but currently it is made sure that this is not the case .. if you wanted to be 100% sure you could compare the values of self.user.auth_method & self.user.username to the ones passed in the interaction.user .. for example in an assert.)

backend/oasst_backend/tree_manager.py
```diff
@@ -522,6 +522,9 @@ async def handle_interaction(self, interaction: protocol_schema.AnyInteraction)
                     )
```
andreaskoepf, line 522: let's move the message size check also to `store_text_reply()` in `PromptRepository` ..

andreaskoepf (approved):
awesome, thanks!

Issue #953, opened 2023-01-27
Send backend HTTP requests with `x-oasst-user` header 


andreaskoepf:
Add a `x-oasst-user: <auth_method>:<username>` whenever possible (when a user is logged in) to all HTTP requests to the backend. Username is the user-id used by the frontend (not to mix with display-name).

baddhidurgarao:
[Buy Github Followers, Stars, Forks & Watchers](https://baddhi.shop/product/buy-github-followers/) from Baddhi Shop

hemangjoshi37a:
@andreaskoepf,

I understand the requirement to add the x-oasst-user header to all HTTP requests to the backend, with the format x-oasst-user: <auth_method>:<username> where username is the user-id used by the frontend.

In addition to implementing this change, I would also suggest adding a check to ensure that the auth_method and username values are properly formatted, and handling any errors that may occur as a result. Additionally, I would recommend adding proper logging to track when the x-oasst-user header is added and its values for debugging and auditing purposes.

I would also suggest that we should mention the related people like @hemangjoshi37a, and also mention the related labels like backend, website so that they can also check it and give suggestions.

Please let me know if this is in line with your expectations and I can proceed to implement these changes.

andreaskoepf:
The backend part is very trivial, I created this issue mainly for the web-team. I think @AbdBarho will take care of it.

AbdBarho:
@andreaskoepf should we add the header to ALL endpoints?

andreaskoepf:
> @andreaskoepf should we add the header to all endpoints, even the ones that did not need user info before?

If you can simply pass on the current logged in user everywhere, then please yes. The user-info is not evaluated everywhere yet .. but it could be used for logging etc to identify on behalf of which frontend-user the calls were made. I created a backend PR https://github.com/LAION-AI/Open-Assistant/pull/957  ...

If if is too cumbersome to do it everywhere than please for now only add it to all message requests.

AbdBarho:
@andreaskoepf I added it everywhere, even if we don't need it now, we can use it generally for rate limiting. #959

andreaskoepf:
@AbdBarho thanks a lot for adapting everything so fast!

PR #959, opened 2023-01-27
Send `x-oasst-user` header to backend

AbdBarho:
#957 #953 

Also fix some UI errors regarding illegal nesting of html types.

### DIFF
--- a/website/src/components/Tasks/LabelTask/LabelTask.tsx
+++ b/website/src/components/Tasks/LabelTask/LabelTask.tsx
@@ -1,14 +1,11 @@
 import { Box, Button, Flex, HStack, Text, useColorModeValue } from "@chakra-ui/react";
-import { useEffect, useMemo, useState } from "react";
-import { useTranslation } from "react-i18next";
+import { useEffect, useState } from "react";
 import { MessageView } from "src/components/Messages";
 import { MessageTable } from "src/components/Messages/MessageTable";
 import { LabelInputGroup } from "src/components/Survey/LabelInputGroup";
 import { TwoColumnsWithCards } from "src/components/Survey/TwoColumnsWithCards";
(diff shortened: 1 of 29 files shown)


website/src/components/Tasks/LabelTask/LabelTask.tsx
```diff
@@ -1,11 +1,13 @@
 import { Box, Button, Flex, HStack, Text, useColorModeValue } from "@chakra-ui/react";
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
```
rjmacarthy, line 3: Should it be `next-i18next`?

AbdBarho: correct, this was leftover code from some experiment so I removed it.

rjmacarthy (approved):
LGTM

website/src/pages/api/admin/status.ts
```diff
@@ -1,17 +1,17 @@
...
 import { withRole } from "src/lib/auth";
-import { oasstApiClient } from "src/lib/oasst_api_client";
-import { getBackendUserCore } from "src/lib/users";
+import { createApiClientFromUser } from "src/lib/oasst_client_factory";
 
 /**
  * Returns tasks availability, stats, and tree manager stats.
  */
 const handler = withRole("admin", async (req, res) => {
+  // NOTE: why are we using a dummy user here?
```
fozziethebeat, line 8: This is to fetch global stats that apply to all users.  By fetching stats for the current admin user, the server would report stats that take into account what the admin user has done.  This is a hacky solution to simulate a global stats query.

fozziethebeat (approved):

PR #963, opened 2023-01-27
language classifier

MattAlexMiracle:
language classifier that takes in a text and returns the most used language (internally it moves a shifting window of N words, classifies them and does majority voting).

For running the model, you first have to `load(modelpath)` and then run `inference_voter(model,text)`.

I also have pretrained weights (trained on wikipedia top 200 most common languages, see https://lukelindemann.com/wiki_corpus.html), but the training runs in ~5min so retraining is also easily possible.

The accuracy is relatively high (average ~80%) especially since some  language duplicates (like "English" and "English simplified") exist.

### DIFF
--- a/backend/oasst_backend/utils/language_classification.py
+++ b/backend/oasst_backend/utils/language_classification.py
@@ -1,0 +1,111 @@
+import os
+import pickle
+from collections import Counter
+
+from sklearn import metrics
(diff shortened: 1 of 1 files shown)


andreaskoepf (approved):
very nice .. thank you.

PR #966, opened 2023-01-28
website: Support new widget types for labelling

othrayte:
Adds proper support for yes/no spam style questions as well as a simple interface for flag style labels.

Also cleaned up the Task component to fix some rerender issues.

Known Issues:
* There are more flags I need to provide user strings for

"Spam" task:
image

Full labeling task:
image

On demand message labeling:
image


andreaskoepf:
Great work, thank you Adrain!

Regarding UI:
- the text in the free labeling dialog refers to a highlighted message .. that makes sense for the task-ui where we show a conversation with highlighted message .. but probably wrong for free-labeling?
- can you change "!helpfullness" <-> "helpfullness" ... to  "Unhelpful" <-> "Helpful" & add the two other missing texts "Moral Judgement" & "Political Content" .. and maybe for creativity make "Unoriginal" the low end?
- IMO fails task should not be shown for "free" text-labeling (or only in cases where the frontend knows that is was an assistant message)
- I would prefer the yes/no buttons of spam+fails_task to be aligned, e.g. centered each on their own line or in a dialog with limited width right-aligned

othrayte:
That should all be doable. On the inclusion of certain labels we currently just show what we get from the valid_labels API, maybe we should have a free labelling one or else some way to indicate the message id so that we can get a more specific list.

website/src/components/Tasks/Task/Task.tsx
```diff
@@ -26,13 +64,64 @@ export interface TaskSurveyProps<T> {
...
+          return status.mode === "REVIEW" ? { mode: "SUBMITTED" } : status;
+        }
+      }
+    },
+    { mode: "EDIT", replyValidity: "INVALID" }
+  );
+
   const replyContent = useRef<TaskContent>(null);
-  const [showUnchangedWarning, setShowUnchangedWarning] = useState(false);
+  console.log("RENDER", taskStatus);
```
notmd, line 109:
```suggestion
```

notmd (changes requested):

AbdBarho (approved):
Thanks!

PR #865, opened 2023-01-20
#769 mock service worker in storybook

jojopirker:
I used MSW to mock some API calls (#769). 
The calls for the `/api/valid_labels` endpoint gets mocked for all stories. 
Code is still kinda messy thats why it is a draft, feedback appreciated! :)

Issue #769, opened 2023-01-16
Get Storybook working with fake API calls


fozziethebeat:
We have quite a few components that fetch dynamic data from API routes.  We need some way to mock those in Storybook stories.

One possible solution is [Mock Service Worker](https://storybook.js.org/addons/msw-storybook-addon).

jojopirker:
I can work on this :)

PR #967, opened 2023-01-28
Copying from #769 to implement mock service worker

fozziethebeat:
Copying from the initial PR that won't be completed to include a Mock Service Worker inside of Storybook so we can more easily mock components that have to fetch data.

This includes some small changes to `MessageTableEntry` that are required due to how storybook mishandles the state objects.

[this is the original PR](https://github.com/LAION-AI/Open-Assistant/pull/865).  I wasn't able to get the pre-commit to run properly and fix it so I had to make a new branch with the copied changes.


website/.prettierignore
```diff
@@ -1,3 +1,4 @@
 .eslintrc.json
 tailwind.config.js
 .storybook/*
+public/mockServiceWorker.js
```
AbdBarho, line 4: this file will also be available in production right? at `/mockServiceWorker.js`?
intended behavior or side effect?

website/.storybook/preview.js
```diff
@@ -12,6 +20,43 @@ export const parameters = {
...
+                  name: "not_appropriate",
+                  display_text: "Inappropriate for customer assistant",
+                  help_text: null,
+                },
+                {
+                  name: "violence",
+                  display_text:
+                    "Encourages or fails to discourage violence/abuse/terrorism/self-harm",
+                  help_text: null,
+                },
```
AbdBarho, line 52: could you move this object out of the function? the nesting will be reduced significantly, making it easier to read.

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -71,6 +71,8 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
     sendEmojiChange({ op: state ? "add" : "remove", emoji });
   };
 
+  console.log(emojiState);
+  console.log(message);
```
AbdBarho, line 75: !!

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -93,7 +95,7 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
           style={{ float: "right", position: "relative", right: "-0.3em", bottom: "-0em", marginLeft: "1em" }}
           onClick={(e) => e.stopPropagation()}
         >
-          {Object.entries(emojiState.emojis).map(([emoji, count]) => (
+          {Object.entries(emojiState?.emojis || {}).map(([emoji, count]) => (
```
AbdBarho, line 98: How could values in `emojiState` be undefined? they areis assigned values on init? maybe it is the `useEffect`? if so, we should provide the default values there.

website/.prettierignore
fozziethebeat: According to the [mock service worker documentation](https://mswjs.io/docs/getting-started/integrate/browser) this seems to be intended (this file is auto generated by their tooling).

website/src/components/Messages/MessageTableEntry.tsx
fozziethebeat: That's a good deduction.  Done.

website/src/components/Messages/MessageTableEntry.tsx
fozziethebeat: Removed!!

website/.storybook/preview.js
fozziethebeat: Done

website/.prettierignore
AbdBarho: https://mswjs.io/docs/getting-started/integrate/browser#start-worker

> It's not recommended to include Mock Service Worker in production. Doing so may lead to a distorted experience for your users.

I think we can leave it in right now, since we don't use a service worker at all, but we should take care of this soon, maybe in the docker image directly.

AbdBarho (approved):

PR #970, opened 2023-01-28
website: Improve reliability of e2e tests

othrayte:
The main change is to useGenericTaskAPI() to have clean cases when the task is set and when it is awaiting progress.
I also changed the frontend to pass through the 503 code we get when tasks are not available, but then decided that "no tasks available" is better than just freezing up on other errors, so all errors are currently handled the same.

Some other type consistency changes were also make to cleanup the task types and also some of the prop names.

### DIFF
--- a/website/src/components/TaskPage/TaskPage.tsx
+++ b/website/src/components/TaskPage/TaskPage.tsx
@@ -4,9 +4,10 @@ import { TaskEmptyState } from "src/components/EmptyState";
 import { LoadingScreen } from "src/components/Loading/LoadingScreen";
 import { Task } from "src/components/Tasks/Task";
 import { TaskInfos } from "src/components/Tasks/TaskTypes";
-import { ERROR_CODES, taskApiHooks } from "src/lib/constants";
+import { taskApiHooks } from "src/lib/constants";
 import { getTypeSafei18nKey } from "src/lib/i18n";
 import { TaskType } from "src/types/Task";
+import { KnownTaskType } from "src/types/Tasks";
 
 type TaskPageProps = {
   type: TaskType;
@@ -15,26 +16,30 @@ type TaskPageProps = {
 export const TaskPage = ({ type }: TaskPageProps) => {
   const { t } = useTranslation(["tasks", "common"]);
   const taskApiHook = taskApiHooks[type];
-  const { tasks, isLoading, reset, trigger, error } = taskApiHook(type);
+  const { response, isLoading, completeTask, skipTask } = taskApiHook(type);
   const taskInfo = TaskInfos.find((taskType) => taskType.type === type);
 
-  if (isLoading) {
-    return <LoadingScreen text={t("common:loading")} />;
+  switch (response.status) {
+    case "AWAITING_INITIAL":
+      return <LoadingScreen text={t("common:loading")} />;
+    case "NONE_AVAILABLE":
+      return <TaskEmptyState />;
+    case "AVAILABLE":
+      return (
+        <>
+          <Head>
+            <title>{t(getTypeSafei18nKey(`${taskInfo.id}.label`))}</title>
+            <meta name="description" content={t(getTypeSafei18nKey(`${taskInfo.id}.desc`))} />
+          </Head>
+          <Task
+            key={response.task.id}
+            frontendId={response.id}
+            task={response.task as KnownTaskType}
+            isLoading={isLoading}
+            completeTask={completeTask}
+            skipTask={skipTask}
+          />
+        </>
+      );
   }
-
-  if (tasks.length === 0 || error?.errorCode === ERROR_CODES.TASK_REQUESTED_TYPE_NOT_AVAILABLE) {
-    return <TaskEmptyState />;
-  }
-
-  const task = tasks[0];
-
-  return (
-    <>
-      <Head>
-        <title>{t(getTypeSafei18nKey(`${taskInfo.id}.label`))}</title>
-        <meta name="description" content={t(getTypeSafei18nKey(`${taskInfo.id}.desc`))} />
-      </Head>
-      <Task key={task.task.id} frontendId={task.id} task={task.task} trigger={trigger} mutate={reset} />
-    </>
-  );
 };
--- a/website/src/components/TaskPage/TaskPage.tsx
+++ b/website/src/components/TaskPage/TaskPage.tsx
@@ -19,27 +19,35 @@ export const TaskPage = ({ type }: TaskPageProps) => {
   const { response, isLoading, completeTask, skipTask } = taskApiHook(type);
   const taskInfo = TaskInfos.find((taskType) => taskType.type === type);
 
-  switch (response.status) {
+  let body;
+  switch (response.taskAvailability) {
     case "AWAITING_INITIAL":
-      return <LoadingScreen text={t("common:loading")} />;
+      body = <LoadingScreen text={t("common:loading")} />;
+      break;
     case "NONE_AVAILABLE":
-      return <TaskEmptyState />;
+      body = <TaskEmptyState />;
+      break;
     case "AVAILABLE":
-      return (
-        <>
-          <Head>
-            <title>{t(getTypeSafei18nKey(`${taskInfo.id}.label`))}</title>
-            <meta name="description" content={t(getTypeSafei18nKey(`${taskInfo.id}.desc`))} />
-          </Head>
-          <Task
-            key={response.task.id}
-            frontendId={response.id}
-            task={response.task as KnownTaskType}
-            isLoading={isLoading}
-            completeTask={completeTask}
-            skipTask={skipTask}
-          />
-        </>
+      body = (
+        <Task
+          key={response.task.id}
+          frontendId={response.id}
+          task={response.task as KnownTaskType}
+          isLoading={isLoading}
+          completeTask={completeTask}
+          skipTask={skipTask}
+        />
       );
+      break;
   }
+
+  return (
+    <>
+      <Head>
+        <title>{t(getTypeSafei18nKey(`${taskInfo.id}.label`))}</title>
+        <meta name="description" content={t(getTypeSafei18nKey(`${taskInfo.id}.desc`))} />
+      </Head>
+      {body}
+    </>
+  );
 };
--- a/website/src/hooks/tasks/useGenericTaskAPI.tsx
+++ b/website/src/hooks/tasks/useGenericTaskAPI.tsx
@@ -1,26 +1,43 @@
 import { useState } from "react";
 import { get, post } from "src/lib/api";
-import { BaseTask, TaskResponse, TaskType as TaskTypeEnum } from "src/types/Task";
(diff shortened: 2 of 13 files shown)


othrayte:
Oh look, the E2E tests passed. 😆

website/src/hooks/tasks/useGenericTaskAPI.tsx
```diff
@@ -1,26 +1,43 @@
...
+  const [isLoading, setIsLoading] = useState(true);
 
-  const [tasks, setTasks] = useState<ConcreteTaskResponse[]>([]);
+  const { mutate: requestNewTask } = useSWRImmutable<TaskAvailableResponse<TaskType>>(
+    "/api/new_task/" + taskType,
+    get,
+    {
+      onSuccess: (response) => {
+        setIsLoading(false);
+        setReponse({ status: "AVAILABLE", ...response });
```
AbdBarho, line 18: could you move the status to its own variable? no need to pollute the actual response.

website/src/hooks/tasks/useGenericTaskAPI.tsx
```diff
@@ -1,26 +1,43 @@
...
+import { TaskApiHook } from "src/types/Hooks";
+import { BaseTask, TaskAvailableResponse, TaskResponse, TaskType as TaskTypeEnum } from "src/types/Task";
 import useSWRImmutable from "swr/immutable";
 import useSWRMutation from "swr/mutation";
 
-export const useGenericTaskAPI = <TaskType extends BaseTask>(taskType: TaskTypeEnum) => {
-  type ConcreteTaskResponse = TaskResponse<TaskType>;
+export const useGenericTaskAPI = <TaskType extends BaseTask>(taskType: TaskTypeEnum): TaskApiHook<TaskType> => {
+  const [response, setReponse] = useState<TaskResponse<TaskType>>({ status: "AWAITING_INITIAL" });
+  const [isLoading, setIsLoading] = useState(true);
```
AbdBarho, line 10: I am not sure how  the behavior of `isLoading` in this hook is different from the default one from  `useSWR`.

website/src/hooks/tasks/useGenericTaskAPI.tsx
```diff
@@ -1,26 +1,43 @@
...
-  });
-
-  const { trigger } = useSWRMutation("/api/update_task", post, {
-    onSuccess: async (newTask: ConcreteTaskResponse) => {
-      setTasks((oldTasks) => [...oldTasks, newTask]);
-      mutate();
+  const { trigger: completeTask } = useSWRMutation<TaskAvailableResponse<TaskType>>("/api/update_task", post, {
+    onSuccess: () => {
+      setIsLoading(true);
+      requestNewTask();
```
AbdBarho, line 33: no set status here?

website/src/pages/api/new_task/[task_type].ts
```diff
@@ -22,8 +24,13 @@ const handler = withoutRole("banned", async (req, res, token) => {
   try {
     task = await oasstApiClient.fetchTask(task_type as string, user, userLanguage);
   } catch (err) {
-    console.error(err);
-    res.status(500).json(err);
+    if (err instanceof OasstError && err.errorCode === ERROR_CODES.TASK_REQUESTED_TYPE_NOT_AVAILABLE) {
+      res.status(503).json({});
+      return;
```
Comment (AbdBarho) [inline website/src/pages/api/new_task/[task_type].ts:29]: return not needed

website/src/types/Task.ts
```diff
@@ -29,12 +29,26 @@ export interface BaseTask {
...
+  status: "AVAILABLE";
+}
+
+interface AwaitingInitialTask {
+  status: "AWAITING_INITIAL";
+}
+
+interface NoTaskAvailable {
+  status: "NONE_AVAILABLE";
+}
```
AbdBarho, line 48: these could be an enum

AbdBarho (changes requested):

website/src/components/TaskPage/TaskPage.tsx
```diff
@@ -15,26 +16,30 @@ type TaskPageProps = {
...
-    return <LoadingScreen text={t("common:loading")} />;
+  switch (response.status) {
+    case "AWAITING_INITIAL":
+      return <LoadingScreen text={t("common:loading")} />;
+    case "NONE_AVAILABLE":
+      return <TaskEmptyState />;
+    case "AVAILABLE":
+      return (
+        <>
+          <Head>
```
rjmacarthy, line 30: Do you think we should move `<Head>` out of the switch statement, and only change the inner page content?

website/src/components/TaskPage/TaskPage.tsx
```diff
@@ -15,26 +16,30 @@ type TaskPageProps = {
 export const TaskPage = ({ type }: TaskPageProps) => {
   const { t } = useTranslation(["tasks", "common"]);
   const taskApiHook = taskApiHooks[type];
-  const { tasks, isLoading, reset, trigger, error } = taskApiHook(type);
+  const { response, isLoading, completeTask, skipTask } = taskApiHook(type);
   const taskInfo = TaskInfos.find((taskType) => taskType.type === type);
 
-  if (isLoading) {
-    return <LoadingScreen text={t("common:loading")} />;
+  switch (response.status) {
```
rjmacarthy, line 22: Hey, do you agree that `response.status` makes the reader think this is related to an response status code e.g 200 or 500?

website/src/hooks/tasks/useGenericTaskAPI.tsx
rjmacarthy, line 18: I'm inclined to think that if we get a `200` from the API the task should be assumed available.

rjmacarthy: I'm inclined to think that if we get a `200` the task should be assumed available, considering if it's not we get an error and a corresponding error code.

website/src/components/TaskPage/TaskPage.tsx
othrayte: Good point, I'll try to find another name for it.

website/src/hooks/tasks/useGenericTaskAPI.tsx
othrayte: The difference, which I only just found out about (I was all ready to change to isLoading from useSWR), is that the isLoading for useSWRImmutable doesn't change back to false when you call mutate().

website/src/hooks/tasks/useGenericTaskAPI.tsx
othrayte: The intent was to make it so that when you have a TaskAvailable you have a task, but if you have an AwaitingInitialTask or NoTaskAvailable you don't. If we separate the variables we are back in the situation where the variables only update one at a time and we would have to handle the cases that you have the "AVAILABLE" state but no task and the "NONE_AVAILABLE" state yet have a task. Originally the loading was also part of the tagged union `TaskResponse` but in testing I (thought I) discovered that we actually intend the isLoading to be possible regardless of the current state of having a task.

website/src/hooks/tasks/useGenericTaskAPI.tsx
othrayte: Correct, the data returned from /api/update_task is just a "task" with the task type "done" (or some equivalent, didn't check the wording), we don't actually want to try to show such a non-task task and so we just need to request a new task.

website/src/types/Task.ts
othrayte: Are you referring to the values for `status`? Sure I can do that, but can you elaborate on what additional benefit we get from replacing the strings with enum values? The `TaskResponse` is already a tagged union and as far as I know (which isn't much, I'm new to typescript) the type system already checks that the strings match, that we have handled all cases in a switch etc, and the autocomplete (at least for me) completes them. Just looking to understand it better and learn more.

fozziethebeat:
I'll leave the review to @AbdBarho

website/src/hooks/tasks/useGenericTaskAPI.tsx
AbdBarho: just to make sure I understood: we use one variable instead of two because they are updated independently, and react might render in-between the updates, leading to an undefined behaviour?

website/src/hooks/tasks/useGenericTaskAPI.tsx
AbdBarho: >  isLoading for useSWRImmutable doesn't change back to false when you call mutate().

That is a good point, could you please write it as a comment in the code

website/src/types/Task.ts
AbdBarho: Yes I am referring to `status`, the enum could also be a string, that's ok, an enum helps show intent better than 3 different interfaces (and another type that combines these interfaces), also reduces the amount of code you need to write, and most importantly, understand.
```ts
export enum TaskAvailability {
   AVAILABLE = "AVAILABLE",
   AWAITING_INITIAL = "AWAITING_INITIAL",
   NONE_AVAILABLE = "NONE_AVAILABLE"
}
```

website/src/hooks/tasks/useGenericTaskAPI.tsx
othrayte: Good idea

website/src/types/Task.ts
othrayte: The reason we can't use an enum instead of the interfaces is that the TaskAvailable one has data (via TaskAvailableResponse), unless of course there is a way to do this with enums, I'm still new to typescript.

website/src/hooks/tasks/useGenericTaskAPI.tsx
othrayte: That is part of it, but the other advantage is it reflects reality more accurately, by definition if the task is available we have task data and when it's not we don't, reflecting that in the types just leads to code that is only as complex as reality.

AbdBarho: I see what you mean, you can set the `task` to `null` while loading, this will also reflect the reality as accurately as before.

The reason they should be separate is to separate the concerns, the status is only used in the `TaskPage`.

Regarding:
> reflecting that in the types just leads to code that is only as complex as reality.

if we want to have an interface for every possible state our system could be in, we will end up with exponentially many, which is unmaintainable in the long run.

website/src/types/Task.ts
AbdBarho: that's a point against combining the status with the task, a task being available or awaiting first load has nothing to do with which task nor its contents.

othrayte: I disagree, the task is what we are loading, it being available and having it are the same thing.

Anyway, now that we have discussed it a bunch and I think we are on the same page, what changes would you like me to make.

AbdBarho: Yeah it dragged along a bit too much, you can just leave it like this for now.

AbdBarho (approved):

PR #971, opened 2023-01-28
website: request message specific labels for on demand labelling

othrayte:
This just passes the message id when getting the labels for the message so that we only get the labels that make sense for the message. (e.g. the "fails task" yes/no question is omitted for prompter messages.

### DIFF
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -160,8 +160,8 @@ export class OasstApiClient {
   /**
    * Returns the valid labels for messages.
    */
-  async fetch_valid_text(): Promise<any> {
-    return this.get(`/api/v1/text_labels/valid_labels`);
+  async fetch_valid_text(messageId?: string): Promise<any> {
+    return this.get("/api/v1/text_labels/valid_labels", { message_id: messageId });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


website/src/lib/oasst_api_client.ts
```diff
@@ -160,8 +160,8 @@ export class OasstApiClient {
   /**
    * Returns the valid labels for messages.
    */
-  async fetch_valid_text(): Promise<any> {
-    return this.get(`/api/v1/text_labels/valid_labels`);
+  async fetch_valid_text(messageId: string = null): Promise<any> {
+    return this.get(`/api/v1/text_labels/valid_labels${messageId && `?message_id=${messageId}`}`);
```
notmd, line 164: You can pass an object as the 2nd params of the get method.

website/src/pages/api/valid_labels.ts
```diff
@@ -5,8 +5,10 @@ import { createApiClient } from "src/lib/oasst_client_factory";
  * Returns the set of valid labels that can be applied to messages.
  */
 const handler = withoutRole("banned", async (req, res, token) => {
+  const { message_id } = req.query;
+  console.log("message_id", message_id);
```
notmd, line 9: Left over?

website/src/lib/oasst_api_client.ts
othrayte: Ah, much nicer

fozziethebeat (approved):

PR #991, opened 2023-01-29
Add anthropic RLHF dataset & deepspeed support for reward model

theblackcat102:
- Added [Anthropic/hh-rlhf](https://huggingface.co/datasets/Anthropic/hh-rlhf) to list of available RLHF datasets. 

- The new trainer also support deepspeed config via deepspeed flag ( same as supervised training ). 

```
deepspeed trainer.py configs/deberta-v2-xxlarge-a100.yml --deepspeed
```

I have ran deberta-v2-xxlarge model for 12 hours ([OA wandb](https://wandb.ai/open-assistant/reward-model?workspace=)) with no major issue except the validation seems to stuck with no apparent error, I feels like its a cuda deadlock or something since the GPU utilization was fix at 100% with no fluctuation. So a work around is simply save the needed checkpoint and run evaluation offline.

### DIFF
--- a/model/reward/instructor/configs/deberta-v2-xxlarge-a100.yaml
+++ b/model/reward/instructor/configs/deberta-v2-xxlarge-a100.yaml
@@ -1,0 +1,17 @@
+model_name: microsoft/deberta-v2-xxlarge
+learning_rate: 2e-6
+scheduler: cosine
+gradient_checkpointing: false
+gradient_accumulation_steps: 12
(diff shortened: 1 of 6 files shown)


sanagno (approved):
Looks good!

zhangsanfeng86:
"the validation seems to stuck with no apparent error"
--------------------
@theblackcat102  Do you solve this problem?
I just set evaluation_strategy="no' to bypass eval.

PR #992, opened 2023-01-29
Update prompting.md

bair82:
I think it's important that everyone who is writing assistant responses should use spellchecking


yk (approved):
thank you

PR #993, opened 2023-01-29
website: Simplify isLoading handling

othrayte:
As a follow up to this comment (https://github.com/LAION-AI/Open-Assistant/pull/970#discussion_r1089880512) in #970 I read up more on useSWR and realised the behavior we wanted is provided by isValidating. I still thought it deserved a comment as renaming isValidating isLoading is still a bit odd.


AbdBarho (approved):
Cool! the less we maintain the better

PR #994, opened 2023-01-29
Consolidate API calls to TaskContext

AbdBarho:
code clean up, delegate all API calls to hook, so that the react code can focus on the data and representation.

Many of the changes is just shuffling interfaces around.

I want to make a couple of PRs after this one to continue cleaning up the code, I am not sure if the `TaskContext` as a provider is the ultimate solution for the long run, but it made sense for now.

### DIFF
--- a/website/src/types/Hooks.ts
+++ b/website/src/types/Hooks.ts
@@ -1,16 +1,11 @@
-import { BaseTask, TaskContent, TaskResponse, TaskType } from "src/types/Task";
+import { BaseTask, TaskResponse, TaskType } from "src/types/Task";
 
-interface TaskInteraction {
-  id: string;
-  update_type: string;
-  content: TaskContent;
-}
-
-export type TaskApiHook<Task extends BaseTask> = {
+export type TaskApiHook<Task extends BaseTask, ResponseContent> = {
   response: TaskResponse<Task>;
   isLoading: boolean;
-  completeTask: (interaction: TaskInteraction) => void;
-  skipTask: () => void;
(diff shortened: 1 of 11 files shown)


fozziethebeat (approved):

fozziethebeat:
Overall looks like a good cleanup once it builds right.

AbdBarho:
@fozziethebeat I made significant changes in the last commit, I will re-request the review.

AbdBarho:
@othrayte you can also review, this touches many of your recent changes.

53jk1 (approved):

website/src/types/Hooks.ts
```diff
@@ -1,16 +1,11 @@
...
-}
-
-export type TaskApiHook<Task extends BaseTask> = {
+export type TaskApiHook<Task extends BaseTask, ResponseContent> = {
   response: TaskResponse<Task>;
   isLoading: boolean;
-  completeTask: (interaction: TaskInteraction) => void;
-  skipTask: () => void;
+  completeTask: (interaction: ResponseContent) => Promise<void>;
+  skipTask: () => Promise<void>;
```
othrayte, line 7: I don't think we need skipTask if we have rejectTask. Probably best not to expose a way to accidentally skip without informing the server we are rejecting.

othrayte (approved):
Great to see some refactoring. 👷

website/src/types/Hooks.ts
AbdBarho: Agree, I will remove it in the next PR.

PR #1004, opened 2023-01-29
fix message tree export & add emojis

andreaskoepf:
I noticed that the changes I made for importing message trees broke the export code. This PR fixes this issue. Additionally I have added code for exporting emoji-stats (might be helpful info for the RM).


yk (approved):
let's hope we never encounter circular trees ;)

PR #1016, opened 2023-01-30
906: update user streaks

melvinebenezer:
1. Remove streak_last_day_date & streak_days from UserStats and add them to User (add a server default current_timestamp() for streak_last_day_date) + add a nullable additional last_activity_date datetimetz column to User.
2. Update last_activity to the current time whenever the user sends a task results, e.g. after tm.handle_interaction() in the [tasks_interaction endpoint](https://github.com/LAION-AI/Open-Assistant/blob/b7fab83d36c1a50de985f1203d14440870d94989/backend/oasst_backend/api/v1/tasks.py#L141).
3. Store the process start time in a global variable.
4. Add a new periodically executed function to main.py (similar to the stats updates) that runs every 4h and updates the streak value but for the first time not before the process ran for more than 24h (e.g. to gracefully recover from potential system outages, simply return if timedelta to procses-start time < 24h). Set the streak_days value to 0 of all users who have a last_activity_date older than 24 hours and streak_days > 0. Increments the streak value by 1 and set streak_last_day_date to the current time for all user's having current_time() - streak_last_day_date > 24 h.

Test Evidence: 
Screenshot 2023-01-30 at 8 20 57 PM


backend/main.py
```diff
@@ -268,6 +271,54 @@ def update_leader_board_total(session: Session) -> None:
...
 
+@app.on_event("startup")
+def set_startup_time() -> None:
+    global startup_time
+    startup_time = utcnow()
+
+
+def update_all_user_streak(session: Session) -> None:
+    statement = select(User)
+    result = session.exec(statement).all()
```
andreaskoepf, line 282: Not sure what this is supposed to do? returns all users by type annotation says None? The name says update but nothing is updated.

backend/main.py
```diff
@@ -268,6 +271,54 @@ def update_leader_board_total(session: Session) -> None:
...
+
+@app.on_event("startup")
+@repeat_every(seconds=60 * settings.USER_STREAK_UPDATE_INTERVAL, wait_first=False)
+@managed_tx_function(auto_commit=CommitMode.COMMIT)
+def update_user_streak(session: Session) -> None:
+    try:
+        global startup_time
+        current_time = utcnow()
+        timedelta = current_time - startup_time
+        result = update_all_user_streak(session=session)
```
andreaskoepf, line 294: the query should be moved after the if

backend/alembic/versions/2023_01_29_1207-7b8f0011e0b0_move_user_streak_from_user_stats_to_.py
```diff
@@ -0,0 +1,31 @@
...
+# revision identifiers, used by Alembic.
+revision = "7b8f0011e0b0"
+down_revision = "49d8445b4c90"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("user", sa.Column("streak_last_day_date", postgresql.TIMESTAMP(), autoincrement=False, nullable=True))
```
andreaskoepf, line 21: all timestamps should be `timestamptz`, e.g. `type_=sa.DateTime(timezone=True)` .. this will make sure that we always deal with timezone aware python datetime objects as results from the db (the offset is normally 0 since our DB tz is set to UTC and internally we also use UTC date ..)

backend/main.py
```diff
@@ -268,6 +271,54 @@ def update_leader_board_total(session: Session) -> None:
...
+        if timedelta.days >= 0:
+            # Update only greater than 24 hours . Do nothing
+            logger.debug("Process timedelta greater than 24h")
+            if result is not None:
+                for user in result:
+                    logger.debug(f"{type(user)}... {user}")
+                    last_activity_date = user.last_activity_date
+                    streak_last_day_date = user.streak_last_day_date
+                    current_time_no_tz = current_time.replace(tzinfo=None)
```
andreaskoepf, line 303: this should not be necessary, see comment regarding `timestamptz`.

backend/main.py
```diff
@@ -268,6 +271,54 @@ def update_leader_board_total(session: Session) -> None:
...
+                    logger.debug(f"{current_time_no_tz}: {last_activity_date}, {streak_last_day_date}")
+                    if last_activity_date is not None:
+                        lastactitvitydelta = current_time_no_tz - last_activity_date
+                        if lastactitvitydelta.days > 1 or user.streak_days is None:
+                            user.streak_days = 0
```
andreaskoepf, line 308: we need to set `user.streak_last_day_date` to current time here too, don't we?

backend/oasst_backend/user_repository.py
```diff
@@ -309,3 +310,12 @@ def query_users_ordered_by_display_name(
             qry = qry.limit(limit)
 
         return qry.all()
+
+    @managed_tx_method(CommitMode.COMMIT)
```
andreaskoepf, line 314: This should be not commit .. either flush or no managed-scope at all .. will then write the changes when the interaction commits ..

backend/oasst_backend/models/user.py
```diff
@@ -32,6 +32,15 @@ class User(SQLModel, table=True):
     deleted: bool = Field(sa_column=sa.Column(sa.Boolean, nullable=False, server_default=sa.false()))
     show_on_leaderboard: bool = Field(sa_column=sa.Column(sa.Boolean, nullable=False, server_default=sa.true()))
 
+    # only used for time span "total"
+    streak_last_day_date: Optional[datetime] = Field(
+        sa_column=sa.Column(sa.DateTime(timezone=True), nullable=True, server_default=sa.func.current_timestamp())
+    )
```
andreaskoepf, line 38: hm. here you specified `timezone=True` .. but the alembic migration script doesn't seem to use them.

backend/main.py
melvinebenezer: yes, it is needed. I added it.

backend/oasst_backend/user_repository.py
melvinebenezer: By Interaction , did you mean the task interaction ? 
This will commit the changes to streak_days and streak_last_day_date based on the consecutive days calculation. 
The other commit will be to update just after the task_interaction. 
Not sure which one you referred to. Let me know.
Anyways I removed the managed-scope to an explicit session.commit()

melvinebenezer: will do the changes as per the discussion on discord

backend/main.py
```diff
@@ -268,6 +271,58 @@ def update_leader_board_total(session: Session) -> None:
...
+
+
+@app.on_event("startup")
+@repeat_every(seconds=60 * settings.USER_STREAK_UPDATE_INTERVAL, wait_first=False)
+def update_user_streak(session: Session) -> None:
+    try:
+        global startup_time
+        current_time = utcnow()
+        timedelta = current_time - startup_time
+        if timedelta.days >= 0:
```
andreaskoepf, line 287: should be > 0

backend/main.py
```diff
@@ -268,6 +271,58 @@ def update_leader_board_total(session: Session) -> None:
...
+def set_startup_time() -> None:
+    global startup_time
+    startup_time = utcnow()
+
+
+@app.on_event("startup")
+@repeat_every(seconds=60 * settings.USER_STREAK_UPDATE_INTERVAL, wait_first=False)
+def update_user_streak(session: Session) -> None:
+    try:
+        global startup_time
```
andreaskoepf, line 284: not necessary, since `startup_time` is not changed in this function.

backend/main.py
```diff
@@ -19,15 +19,18 @@
...
 from oasst_shared.exceptions import OasstError, OasstErrorCode
 from oasst_shared.schemas import protocol as protocol_schema
+from oasst_shared.utils import utcnow
 from pydantic import BaseModel
-from sqlmodel import Session
+from sqlmodel import Session, select
 from starlette.middleware.cors import CORSMiddleware
 
 app = fastapi.FastAPI(title=settings.PROJECT_NAME, openapi_url=f"{settings.API_V1_STR}/openapi.json")
+startup_time = None
```
andreaskoepf, line 33: could you add `datetime` type annotation?

backend/main.py
```diff
@@ -268,6 +271,58 @@ def update_leader_board_total(session: Session) -> None:
...
+            statement = select(User)
+            result = session.exec(statement).all()
+            if result is not None:
+                for user in result:
+                    last_activity_date = user.last_activity_date
+                    streak_last_day_date = user.streak_last_day_date
+                    # set NULL streak_days to 0
+                    if user.streak_days is None:
+                        user.streak_days = 0
+                    # if the user had completed a task
```
andreaskoepf, line 299: had *never* completed a task?

backend/main.py
melvinebenezer: yes

andreaskoepf (approved):
OK, nice feature addition!

PR #1024, opened 2023-01-30
inference: Define container images and compose descriptor for the inference stack

### LINKED ISSUE #913: Setup docker configurations for live inference server

occupytheweb:
This changeset defines the container images required for the inference stack, and integrates them to the unified compose descriptor.

This should, in principle, supersede the present `inference/full-dev-setup.sh`.

Notes:
- Here, the more standard `profiles` key of the compose file is used to group the inference stack (instead of a relying on a sleep-only service with a `depends_on` specification). In my opinion, this approach could be extended to other services for improved DX.
- BuildKit is leveraged to enable caching pip's operations across builds. This is described in the docs supplied.

Fixes #913.

Demo:
open-assistance-inference-docker-setup-test


yk:
Very cool, thank you! I see you're mounting various things such as the code folders. Am I right to assume we would use different images for actual deployment to prod? Like, ones that copy the code instead of mounting? If yes, could we call the dockerfiles you just made something that makes it clear they're for development?

occupytheweb:
> Very cool, thank you! I see you're mounting various things such as the code folders. Am I right to assume we would use different images for actual deployment to prod? Like, ones that copy the code instead of mounting? If yes, could we call the dockerfiles you just made something that makes it clear they're for development?

Hi @yk! 
We can use the same dockerfiles to build the prod images.
During build, the code is already copied-in. I just cater for the mounts of `oasst-shared`, and `inference-{server,worker}` to favour a development setup.

The current final stage of the dockerfiles is actually `dev`. I can add a `prod` stage.
We'd then just target the `dev` stage and `prod` stage, as required, at build time.

It would just imply dropping any volume declarations, running uvicorn without `--reload`, and building `oasst-shared` as non-editable.

We'd then be able to do, for instance:
```yaml
  inference-server:
    build:
      dockerfile: docker/inference/Dockerfile.server
      context: .
      target: prod
```

<br/>
Thus, do we want a dev or prod setup in the root compose file?

yk (approved):
thanks for the elaborations! the compose setup should definitely have the dev versions, that makes a lot of sense. I would just be nice if we could re-use the dockerfiles for building the prod containers, like you say.

occupytheweb:
@yk, ok, I'll add the prod stage.

For the compose setup, I was thinking about having the root compose file with a prod setup, and an _override_ file for the dev-only additions.
Which would yield:

```yaml
# docker-compose.yaml
...
  inference-server:
    build:
      dockerfile: docker/inference/Dockerfile.server
      context: .
    image: oasst-inference-server
    environment:
      - "PORT=8000"
      - "REDIS_HOST=redis"
    restart: unless-stopped
    depends_on:
      redis:
        condition: service_healthy
    profiles: ["inference"]

  inference-worker:
    build:
      dockerfile: docker/inference/Dockerfile.worker
      context: .
    image: oasst-inference-worker
    environment:
      - "BACKEND_URL=ws://inference-server:8000"
      - "INFERENCE_SERVER_URL=http://inference-text-generation-server"
    depends_on:
      - inference-server
    deploy:
      replicas: 1
    profiles: ["inference"]

  inference-text-client:
    build:
      dockerfile: docker/inference/Dockerfile.text-client
      context: .
    image: oasst-inference-text-client
    environment:
      - "BACKEND_URL=http://inference-server:8000"
    tty: true
    stdin_open: true
    restart: unless-stopped
    depends_on:
      - inference-server
    profiles: ["inference"]
```

and:
```yaml
# docker-compose.dev.yaml
  inference-server:
    build:
      target: dev
    image: oasst-inference-server:dev
    volumes:
      - "./oasst-shared:/opt/inference/lib/oasst-shared"
      - "./inference/server:/opt/inference/server"

  inference-worker:
    build:
      target: dev
    image: oasst-inference-worker:dev
    volumes:
      - "./oasst-shared:/opt/inference/lib/oasst-shared"
      - "./inference/worker:/opt/inference/worker"
    deploy:
      replicas: 1

  inference-text-client:
    volumes:
      - "./inference/worker:/opt/inference/worker"
```
<br/>

To use the dev stack, we'd do:
```shell
dk compose --profile inference -f docker-compose.yaml -f docker-compose.dev.yaml up 
```

This pattern could also be used for the services as well, removing the need for the sleep containers.
What do you think?

yk:
I'm totally in favor of removing the sleep containers. I think the prod setup in the compose is less necessary than the dev setup, since for prod we deploy with ansible, it's really just about building the images without mounts.

occupytheweb:
@yk 
Ok great. I added the prod stage to the images, and updated the compose file to target the dev stage.

> I'm totally in favor of removing the sleep containers. 

I can think of a clean way to do this, and submit another PR.
I also found some places where we could streamline the DX by reducing bloat and time required to build the containers. I'll look into it.

PR #1026, opened 2023-01-30
docs: prompting.md: fix typo and add some markup

lainisourgod:
it's easier to read dialog with mark up


andrewm4894 (approved):

lainisourgod:
@andrewm4894 I believe there's some bug with second `pre-commit` workflow that is constantly pending waiting for smth. I guess it stayed there from first commit. Maybe we should drop it?

PR #1036, opened 2023-01-31
Limit number of replies to assistant messages

andreaskoepf:
Before this PR the system tried to generate max_children_count prompter reply tasks. With this PR we stop collecting prompter messages more early (defined by num_prompter_replies).

- include make rate limit settings in config
- minor improvements to auto_main.py test code (labeling tasks)
- use uuid for frontend message id in auto_main (avoid duplicates)


yk (approved):
thank you

PR #1038, opened 2023-01-31
Add endpoint to query nearby leaderboard rows

andreaskoepf:
- add /api/v1/users/{user_id}/stats/{time_frame}/window" endpoint
- add `highlighted` bool flag `UserScore` which is true for current user (when frontend user info is passed in header) for normal leaderboards and it is true for the entry of the user the user_id of the /window request

### DIFF
--- a/oasst-shared/oasst_shared/schemas/protocol.py
+++ b/oasst-shared/oasst_shared/schemas/protocol.py
@@ -26,7 +26,7 @@ class TaskRequestType(str, enum.Enum):
 class User(BaseModel):
     id: str
     display_name: str
-    auth_method: Literal["discord", "local"]
+    auth_method: Literal["discord", "local", "system"]
 
 
 class FrontEndUser(User):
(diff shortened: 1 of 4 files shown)


oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -26,7 +26,7 @@ class TaskRequestType(str, enum.Enum):
 class User(BaseModel):
     id: str
     display_name: str
-    auth_method: Literal["discord", "local"]
+    auth_method: Literal["discord", "local", "system"]
```
yk, line 29: is it ok that this is in this PR?

yk (approved):
thank you, very neat!

oasst-shared/oasst_shared/schemas/protocol.py
andreaskoepf: The user created for the importer has `auth_method == "system"`. This user is for example returned by the user_cursor queries .. pydantic otherwise complains when we try to create a FontendUser object for it which is derived from protocol.User ...

Issue #1022, opened 2023-01-30
Implement critical admin functions


andreaskoepf:
We need Admin UI for the following functions (**only displayed for admin users**): 

- [x] User-Id link in every message that navigates to the admin user-page of the author (e.g. to disable a user, inspect message history) (resolved via https://github.com/LAION-AI/Open-Assistant/pull/1021)
- [x] "delete message" command for all messages (e.g. in "..." drop down) -> call `DELETE` `/api/v1/messages/{message-id}` ..
- [x] "stop tree" command in all messages (set them into halted-by-moderator state) (e.g. in "..." drop down) (backend function to halt tree added by: https://github.com/LAION-AI/Open-Assistant/pull/1025)
- [ ] browse through all the user's messages (not just last 10, e.g. user message cursor endpoint)
- [ ] browse all message trees of the system (e.g. via general message cursor endpoint, only roots first and then view the whole message tree)

rjmacarthy:
I have opened a PR for point number 1.

kingychiu:
I am looking at the 4th point here `browse through all the user's messages (not just last 10, e.g. user message cursor endpoint)`

The current state of this page is to have the latest 10 messages of a user, but the UI is showing the latest message at the top. Is it expected behavior?

![Image](https://user-images.githubusercontent.com/11500136/216825753-24d886c7-bcb9-46c4-ad16-e33664d6365b.png)


About the pagination/cursor feature here. I am thinking of having a button `Load older message" to load 10 more old messages at the top/bottom depending on the answer of the question above.

notmd:
@kingychiu have you start the work?

kingychiu:
@notmd  Not yet

notmd:
I think this can be closed now

PR #1041, opened 2023-01-31
website: Add delete message feature for admins

rjmacarthy:
This PR addresses point number two for issue #1022 which add the ability to delete a message from the actions drop-down menu.  The back-end requires some work for individual message not to return after being deleted.

Any questions or concerns please let me know.

### DIFF
--- a/website/public/locales/en/common.json
+++ b/website/public/locales/en/common.json
@@ -18,5 +18,6 @@
   "sign_out": "Sign Out",
   "terms_of_service": "Terms of Service",
   "title": "Open Assistant",
-  "yes": "Yes"
+  "yes": "Yes",
+  "delete": "Delete"
 }
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -14,7 +14,7 @@ import {
   useDisclosure,
 } from "@chakra-ui/react";
 import { boolean } from "boolean";
-import { ClipboardList, Flag, MessageSquare, MoreHorizontal, User } from "lucide-react";
+import { ClipboardList, Flag, MessageSquare, MoreHorizontal, Trash, User } from "lucide-react";
 import { useRouter } from "next/router";
 import { useSession } from "next-auth/react";
 import { useTranslation } from "next-i18next";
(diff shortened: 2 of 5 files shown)


website/src/pages/api/messages/[id]/index.ts
```diff
@@ -2,12 +2,48 @@ import { withoutRole } from "src/lib/auth";
...
+    const client = createApiClientFromUser(user);
+    await client.delete_message(id);
+    res.status(200).json({ message: "Message deleted" });
+  } catch (e) {
+    res.status(500).json({ message: "Failed to delete message" });
+  }
+};
+
+const handlers = {
+  GET: getMessage,
```
Comment (fozziethebeat) [inline website/src/pages/api/messages/[id]/index.ts:36]: I feel like this is kinda confusing.  I would much rather we have a pure delete API route in `website/src/pages/api/admin/delete_message.ts`.

That way you don't need all this meta handler wrappers and the handler can directly use `withRole('admin')`.

fozziethebeat (changes requested):

backend/oasst_backend/prompt_repository.py
```diff
@@ -673,7 +673,8 @@ def fetch_message(self, message_id: UUID, fail_if_missing: bool = True) -> Optio
         messages = self._add_user_emojis_all(qry)
         message = messages[0] if messages else None
 
-        message = self.db.query(Message).filter(Message.id == message_id).one_or_none()
+        message = self.db.query(Message).filter(Message.id == message_id, not_(Message.deleted)).one_or_none()
```
fozziethebeat, line 676: Oof, i can't approve for this.  You'll need to get the backend team to review this (or make it a separate PR)

fozziethebeat (approved):

backend/oasst_backend/prompt_repository.py
rjmacarthy: Understood. @andreaskoepf, please could you take a look?

andreaskoepf: FetchMessage is a low level function that needs to be able to pull every message out of the db. I will make changes for deleted messages later if that should be necessary. Please restore this file for now.

PR #1043, opened 2023-01-31
Fix #1034, split into paragraphs for full text view.

### LINKED ISSUE #1034: Full Text popup during ranking does not handle newlines correctly
image

I wrote this so I know there are a couple of newlines in there. probably just needs some white-space: pre-line or so

kpoeppel:
This should fix issue #1034. Text is split into paragraphs now, without any additional styling. The change in the dataset was used for testing.

### DIFF
--- a/website/src/components/CollapsableText.tsx
+++ b/website/src/components/CollapsableText.tsx
@@ -29,7 +29,7 @@ export const CollapsableText = ({
   if (typeof text !== "string" || text.length <= maxLength) {
     return <>{text}</>;
   } else {
-    const text_with_breaks = text.split('\n').map((str, index) => <p key={index}>{str}</p>);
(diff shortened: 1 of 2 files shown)


website/src/components/CollapsableText.tsx
```diff
@@ -29,6 +29,7 @@ export const CollapsableText = ({
   if (typeof text !== "string" || text.length <= maxLength) {
     return <>{text}</>;
   } else {
+    const text_with_breaks = text.split("\n").map((str, index) => <p key={index}>{str}</p>);
```
fozziethebeat, line 32: I feel like there's gotta be a less strange way to handle this.  Have you looked into alternatives?

fozziethebeat (changes requested):

fozziethebeat:
Double checking the changes submitted in another PR, I am going to close this as it fixes the problem with a simpler solution that let's css fix the problem for us.

PR #1049, opened 2023-02-01
Various website cleanup

notmd:
- Remove some unused dependencies.
- Remove the root styles folder.
- Disable ESlint during build because we already have pre-commit.
- Use the Card component in a few more places.


fozziethebeat (approved):
Deletions!

PR #1051, opened 2023-02-01
Setup bundle analyzer

notmd:


website/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
+    "build:analyze": "ANALYZE=true next build",
```
AbdBarho, line 9: this won't work on windows

notmd: Ok I will replace with cross-env

notmd: Ok I will replace with cross-env

AbdBarho (approved):

PR #1055, opened 2023-02-01
Add infty prompting + prompt-lottery

andreaskoepf:
- all initial prompts enter PROMPT_LOTTERY_WAITING state after they passed spam tests
- prompt lottery selects among available in waiting state randomly when the number of active trees for the current language falls below the target config setting
- prompt selection: uniform among distinct authors, then randomly among author's prompts .. every user who submitted a prompt has the same chance that one of her/his initial prompts is chosen


yk (approved):
lgtm, thank you

PR #1056, opened 2023-02-01
Feat/task handler

AlexanderHOtt:
Checklist
- [ ] get log_channel id from db instead of hard coding it

The diff for work.py is confusing because I basically refactored all of the logic into handler classes. I would look at the files side by side.


AlexanderHOtt:
This should be ready for review

andreaskoepf (approved):
ok, nice to see some progress here.

PR #1057, opened 2023-02-01
Allow to enable/disable email signin via env variable

### LINKED ISSUE #911: Reduce ability for users to register with fake email accounts

notmd:
close #911


andreaskoepf (approved):
Further changes neccessary to set env-var in dev, staging & production..

AbdBarho:
do we need to add the variable to https://github.com/LAION-AI/Open-Assistant/blob/main/ansible/deploy-to-node.yaml ?

notmd:
So do I need to update other config?

andreaskoepf:
> do we need to add the variable to https://github.com/LAION-AI/Open-Assistant/blob/main/ansible/deploy-to-node.yaml ?

Yes and in https://github.com/LAION-AI/Open-Assistant/blob/main/.github/workflows/deploy-to-node.yaml when we want to control the variable via github. As far as I understand the check for the variable defaults to false .. i.e. the feature is disabled if the variable is missing. This means we need to make sure it is properly set for staging and production .. otherwise email-signup will not work.

fozziethebeat (approved):
Yup this all looks right

fozziethebeat:
Actually wait, I'm going to un-approve this until I go through the sign up process and set the env variables.  Merging this too quickly could break things.

fozziethebeat (changes requested):
Requesting myself to make changes.  I'll do that tomorrow.

.github/workflows/deploy-to-node.yaml
```diff
@@ -33,6 +33,14 @@ jobs:
       WEB_EMAIL_SERVER_PORT: ${{ secrets.DEV_WEB_EMAIL_SERVER_PORT }}
       WEB_EMAIL_SERVER_USER: ${{ secrets.DEV_WEB_EMAIL_SERVER_USER }}
       WEB_NEXTAUTH_SECRET: ${{ secrets.NEXTAUTH_SECRET }}
+      WEB_NEXT_PUBLIC_CLOUDFLARE_CAPTCHA_SITE_KEY:
+        ${{ secrets.DEV_WEB_NEXT_PUBLIC_CLOUDFLARE_CAPTCHA_SITE_KEY }}
+      WEB_CLOUDFLARE_CAPTCHA_SERCERT_KEY:
+        ${{ secrets.DEV_WEB_CLOUDFLARE_CAPTCHA_SERCERT_KEY }}
```
andreaskoepf, line 39: @fozziethebeat thx for preparing.  I have pushed these changes now already because the captcha settings were missing. see https://github.com/LAION-AI/Open-Assistant/commit/776584503f6b7d0fd18ddcb48bcea6ded4119543

fozziethebeat (approved):

PR #1066, opened 2023-02-02
hide system emojis starting with '_'

andreaskoepf:
The skip state of messages is stored in special emoji codes... currently they show up as undefined emojis:

image

This PR filters emoji-codes with "_" prefix.

### DIFF
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -98,7 +98,7 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
           onClick={(e) => e.stopPropagation()}
         >
           {Object.entries(emojiState.emojis)
-            .filter(([k, _]) => !k.startsWith("_"))
+            .filter(([k]) => !k.startsWith("_"))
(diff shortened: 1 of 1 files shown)


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -97,14 +97,16 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
...
-          {Object.entries(emojiState.emojis).map(([emoji, count]) => (
-            <MessageEmojiButton
-              key={emoji}
-              emoji={{ name: emoji, count }}
-              checked={emojiState.user_emojis.includes(emoji)}
-              onClick={() => react(emoji, !emojiState.user_emojis.includes(emoji))}
-            />
-          ))}
+          {Object.entries(emojiState.emojis)
+            .filter(([k, _]) => !k.startsWith("_"))
```
notmd, line 101:
```suggestion
            .filter(([k]) => !k.startsWith("_"))
```

notmd (changes requested):

fozziethebeat (approved):

PR #1067, opened 2023-02-02
fix side menu regression

notmd:
fix regression introduced in #1049. just notice this bug when visiting our dev website with a mobile.

### DIFF
--- a/website/src/components/SideMenu.tsx
+++ b/website/src/components/SideMenu.tsx
@@ -21,8 +21,9 @@ export function SideMenu(props: SideMenuProps) {
   return (
     <main className="sticky top-0 sm:h-full">
       <Card
+        display={{ base: "grid", sm: "flex" }}
         width={["100%", "100%", "100px", "280px"]}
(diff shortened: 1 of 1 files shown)


website/src/components/SideMenu.tsx
```diff
@@ -21,8 +21,9 @@ export function SideMenu(props: SideMenuProps) {
   return (
     <main className="sticky top-0 sm:h-full">
       <Card
+        display={{ base: "grid", sm: "flex" }}
         width={["100%", "100%", "100px", "280px"]}
-        className="grid grid-cols-4 gap-2 sm:flex sm:flex-col sm:justify-between p-4 h-full"
+        className="grid-cols-4 gap-2 sm:flex-col sm:justify-between p-4 h-full"
```
AbdBarho, line 26: How is the new behvaior different than the one before? it seems to me that they are identical?

notmd: The Card component has display flex by default and It will override tailwindcss.
Here is how it looks on the dev instance.
image

AbdBarho (approved):

PR #1072, opened 2023-02-02
System config parameter view

### LINKED ISSUE #1068: Add minimalistic system config parameter view (readonly, just the json) to admin dashboard
- use backend API: GET `/api/v1/admin/backend_settings/full` and if that fails use GET `/api/v1/admin/backend_settings/public` as a fallback (only trusted api-clients can access the full settings).
- add a "Parameters" tab on the left side, if selected show simply the current config as JSON text (similar to other things in the admin interface).

notmd:
close #1068 

- Add new AdminArea component
- Add new json variant to Card component

image

### DIFF
--- a/website/src/components/AdminArea.tsx
+++ b/website/src/components/AdminArea.tsx
@@ -1,0 +1,19 @@
+import { useRouter } from "next/router";
+import { useSession } from "next-auth/react";
+import { ReactNode, useEffect } from "react";
+
+export const AdminArea = ({ children }: { children: ReactNode }) => {
+  const router = useRouter();
+  const { data: session, status } = useSession();
+
+  useEffect(() => {
+    if (status === "loading") {
+      return;
+    }
+    if (session?.user.role === "admin") {
+      return;
+    }
+    router.push("/");
+  }, [router, session, status]);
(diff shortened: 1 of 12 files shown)


website/src/components/AdminArea.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+  useEffect(() => {
+    if (status === "loading") {
+      return;
+    }
+    if (session?.user.role === "admin") {
+      return;
+    }
+    router.push("/");
+  }, [router, session, status]);
```
AbdBarho, line 17: this use effect can be simplified into 
```ts
if (session?.user.role !== "admin")
   router.push("/")
```

AbdBarho (commented):
looks good

website/src/components/AdminArea.tsx
notmd: > this use effect can be simplified into
> 
> ```ts
> if (session?.user.role !== "admin")
>    router.push("/")
> ```

This will rediect if session is null (when the status is loading or unauthenticated)
image

AbdBarho: oh I see! makes sense

AbdBarho (approved):

PR #1076, opened 2023-02-02
Update Spanish translation of the website

GuilleHoardings:
Better translation of some strings and translated new strings added to the website


AbdBarho (approved):
thanks

PR #1078, opened 2023-02-03
Add TaskReply types for task hooks

AbdBarho:
With the PR every part of the task process is now typed.

We still do a lot of casting around which is not ideal but this works for now.


website/src/hooks/tasks/useEvaluateReplies.ts
```diff
@@ -0,0 +1,13 @@
...
+import { EvaluateTaskReply } from "src/types/TaskResponses";
+import { RankAssistantRepliesTask, RankInitialPromptsTask, RankPrompterRepliesTask } from "src/types/Tasks";
+
+export const useRankAssistantRepliesTask = () =>
+  useGenericTaskAPI<RankAssistantRepliesTask, EvaluateTaskReply>(TaskType.rank_assistant_replies);
+
+export const useRankPrompterRepliesTask = () =>
+  useGenericTaskAPI<RankPrompterRepliesTask, EvaluateTaskReply>(TaskType.rank_prompter_replies);
+
+export const useRankInitialPromptsTask = () =>
```
AbdBarho, line 12: this file was renamed

fozziethebeat (approved):

PR #1079, opened 2023-02-03
Translate strings that were not using i18n yet

GuilleHoardings:
Only English and Spanish translations have been included.

### DIFF
--- a/website/public/locales/en/leaderboard.json
+++ b/website/public/locales/en/leaderboard.json
@@ -10,5 +10,9 @@
   "weekly": "Weekly",
   "prompt": "Prompts",
   "reply": "Replies",
-  "label": "Labels"
+  "label": "Labels",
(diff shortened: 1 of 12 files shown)


GuilleHoardings:
There is an issue: I was not able to translate `Layout.tsx` because `getDashboardLayout` is not a React hook, so I cannot call `useTranslation`. Actually, I can use it and it works, but the linter in the pre-commit fails. If anyone knows how to fix it, please let me know.

I see also that two checks fail here. I will take a look to then later.###

fozziethebeat (approved):
Looks good! thank you for adding more translations

GuilleHoardings:
Thank you, @fozziethebeat . I would like to complete the internationalization of `Layout.tsx`. Do you know where would be the best place to ask about the problem I'm having?

PR #1080, opened 2023-02-03
Add ToS/License Agreement to Website

### LINKED ISSUE #1047: Add ToS/License Agreement/Cookies popup to Web-App
Every user of the web application has to accept the OA ToS/license agreement before submitting data

I have added backend functionality to manage users and checks for the acceptance:
https://github.com/LAION-AI/Open-Assistant/pull/1046

Required changes in web:
1. on login or dashboard navigation check ToS acceptance status at backend (potentially create user via new POST `/api/v1/frontend_users/` endpoint.
2. check if user has accepted ToS - if not display popup dialog with legal text and checkbox or Yes / No buttons.
3. Navigate user to https://laion.ai/ if they select No (or log user out)

Legal team:
- needs to prepare document to show

AbdBarho:
Refs #1047 

A lot of changes are mostly clean up:

in  the `oasst_api_client`, the functions have been re-ordered, to move the general ones at the top.

in `terms-and-service`, extracted the constant part out of the react component, to make it re-usable.

~~TBH I am not happy with this solution, because it kills server side rendering, since we fetch the tos status on the client side.~~

This PR is getting big, further changes will be addressed in the next PR.


website/src/lib/oasst_api_client.ts
```diff
@@ -3,16 +3,10 @@ import { LeaderboardReply, LeaderboardTimeFrame } from "src/types/Leaderboard";
 import type { AvailableTasks } from "src/types/Task";
 import type { BackendUser, BackendUserCore, FetchUsersParams, FetchUsersResponse } from "src/types/Users";
 
-export class OasstError {
+export interface OasstError {
```
AbdBarho, line 6: changed to an interface since next js is not able serialize this object, it always shows `error [Object object]` in the console, which is not helpful.

when left as a normal object, nextjs per default will show all errors in the console

website/src/lib/oasst_api_client.ts
```diff
@@ -30,12 +24,66 @@ export class OasstApiClient {
     }
   }
 
-  fetch_full_settings() {
-    return this.get<Record<string, any>>("/api/v1/admin/backend_settings/full");
+  private async request<T>(
```
AbdBarho, line 27: this function has been moved to the top from below to reflect the call hierarchy in code

website/src/pages/dashboard.tsx
```diff
@@ -44,9 +46,11 @@ const Dashboard = () => {
...
       </Head>
       <Flex direction="column" gap="10">
-        <WelcomeCard />
-        <TaskOption content={availableTaskTypes} />
-        <LeaderboardWidget />
+        <ToSWrapper>
+          <WelcomeCard />
+          <TaskOption content={availableTaskTypes} />
+          <LeaderboardWidget />
+        </ToSWrapper>
```
AbdBarho, line 53: I added this wrapper here, although implementation-wise it could be added anywhere within the application.

fozziethebeat (changes requested):
I'm mostly ready to approve this but I wanted to ask if you considered embedding this tos acceptance stuff in the JWT and session?  The check against the backend could happen anytime someone logs in and then the wrapper could simply check the session object rather than call a new API route.

AbdBarho:
> I'm mostly ready to approve this but I wanted to ask if you considered embedding this tos acceptance stuff in the JWT and session? The check against the backend could happen anytime someone logs in and then the wrapper could simply check the session object rather than call a new API route.

I have not, I will poke around a bit and see if it is easily doable.

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -94,20 +99,44 @@ const authOptions: AuthOptions = {
...
+         * in the frontend, when the user accepts the tos, we do a full refresh
+         * which means this function will be called again.
+         */
+        tosAcceptanceDate = await oasstApiClient.fetch_tos_acceptance(user);
+      } catch (err) {
+        if (err.httpStatusCode !== 404) {
+          throw err;
+        }
+      }
+
```
Comment (AbdBarho) [inline website/src/pages/api/auth/[...nextauth].ts:135]: @fozziethebeat do you have any idea how to overcome this? or am I misinterpreting the situation?

AbdBarho (commented):
I think it is much cleaner with the token approach, great idea.

I reverted the changes to oasst error since I realized I can just override the tostring function

website/src/pages/api/auth/[...nextauth].ts
Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts]: from @andreaskoepf  on discord:

> I don't think that is already happening .. I added an endpoint for create-or-update user only very recently ... I fear the old code only indirectly notified the backend by sending a new user for the next-task query.

So this should be on our todo list

> Create or Update can be done by POST to  /api/v1/frontend_users/ ...
It returns the created/updated user info ...
(fields that are null are ignored .. eg. any of enabled, show_on_leaderboard, tos_acceptance, notes ) 


I have updated the comments to reflect this.

website/src/lib/oasst_api_client.ts
AbdBarho: I added a to string and it worked.

website/src/components/ToSWrapper.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const acceptToS = async () => {
+  await post("/api/tos", { arg: {} });
+  location.reload();
+};
+
+export const ToSWrapper = ({ children }: { children?: ReactNode | undefined }) => {
+  const { t } = useTranslation("tos");
+  const { data: session } = useSession();
+  const hasAcceptedTos = Boolean(session?.user.tosAcceptanceDate);
```
notmd, line 22: we should check for status === "loading" here. There is flash when hard reloading the page now because session is null

notmd (changes requested):

website/src/components/ToSWrapper.tsx
AbdBarho: https://github.com/LAION-AI/Open-Assistant/pull/1080/commits/61914e2f26d02e43d80230f2d5888d38b1d58c4e

notmd (approved):
lgtm

PR #1088, opened 2023-02-03
website: Add halt tree, and copy message id features

rjmacarthy:
This PR implements task number three for issue #1022.

- Add halt/stop tree feature 
- Add copy message id to clipboard feature
- Sort alphabetical to the translation files to prevent merge issues

If you have any questions or concerns please let me know.


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -155,16 +158,47 @@ const MessageActions = ({
...
+  const handleCopyId = () => {
+    navigator.clipboard.writeText(message.id);
+    toast({
+      title: t("common:copied"),
+      description: displayId,
+      status: "info",
+      duration: 5000,
+      isClosable: true,
+    });
+  };
```
notmd, line 200: CharkaUI has a useClipboard hook, we should use it here

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -155,16 +158,47 @@ const MessageActions = ({
...
+        description: t("tree_stopped", { id: displayId }),
+        status: "success",
+        duration: 5000,
+        isClosable: true,
+      });
+    },
+  });
+
+  const { data: session } = useSession();
+  const isAdmin = session?.user?.role === "admin";
```
notmd, line 180: This logic is quite often, maybe we should extract it to useHasRole(role: Role) hook?

website/src/components/Messages/MessageTableEntry.tsx
rjmacarthy: If you look here https://github.com/chakra-ui/chakra-ui/issues/6759 there is still some confusion regarding this hook, so I think it's simpler and cleaner to go with native solution for now.

website/src/components/Messages/MessageTableEntry.tsx
rjmacarthy: I think this is probably out of scope for this PR and better refactored in a separate pull request.

fozziethebeat: I agree we should clean this up into a hook.  In a follow up PR works for me.

website/src/components/Messages/MessageTableEntry.tsx
fozziethebeat: I'm not quite sure what the confusion is with useClipboard.  That mentioned issue is closed and in the [hook docs](https://chakra-ui.com/docs/hooks/use-clipboard) they show how to use `setValue`.  Can we use that following their example?

fozziethebeat: Also, why do we need to copy the id to the clipboard?

fozziethebeat (changes requested):

website/src/components/Messages/MessageTableEntry.tsx
rjmacarthy: I welcome you to try that. `setValue` does not work as expected and you need to use an effect to make it work because we don't have an on change function. I added this feature because there's no easy way to know what message you are deleting or halting on the UI.

website/src/components/Messages/MessageTableEntry.tsx
rjmacarthy: I welcome you to try that. setValue does not work as expected and you need to use an effect to make it work because we don't have an on change function. Read the last comment in the tracked issue. I added this feature because there's no easy way to know what message you are deleting or halting on the UI.

andreaskoepf:
Regarding "copy the id to the clipboard": Not sure who requested it .. but IMO it is a useful feature for admin stuff like DB editing.

fozziethebeat (approved):

PR #1090, opened 2023-02-04
Fixing the way we generate the backenduser object during jwt creation

fozziethebeat:


AbdBarho (approved):

AbdBarho:
I am still having a hard time wrapping my head around how our auth should work. maybe it will be simpler once we move it to the python backend.

PR #1092, opened 2023-02-04
Add inpainted biomedical dialogue dataset

### LINKED ISSUE #1029: Biomedical dialogue data
Rephrase some existing biomedical text datasets as dialogue or QA

uyhcire:
Closes #1029


olliestanley:
@uyhcire Would it be possible to add a brief summary of what the dataset actually is to the README file? We have a lot of data notebooks so it would make it easier for the ML team to check at a glance without having to go to external sources

PR #1115, opened 2023-02-04
add ko localization

jeonsworld:
* added ko localization files for website
* ran pre-commit to ensure the JSONs are correctly formatted


AbdBarho (approved):
Great, thanks!

PR #1122, opened 2023-02-04
Putting a button for toggling the light/dark color mode right into the navigation bar

theopfr:
Hey! The user can only change the UI color mode when s/he is signed in. Also, the current color toggle button in the dashboard seems a bit misplaced.

### Solution:
Putting the button for toggling the light/dark color mode right into the navigation bar. The button could just display the sun icon in order to not take away too much space and be on the right side of the header.

### Example images:
oa_color_mode_light
oa_color_mode_dark

I haven't removed the current button in the dashboard, but if the button works fine in the navigation bar, it could probably be removed.

### DIFF
--- a/website/src/components/Header/ColorModeToggler.jsx
+++ b/website/src/components/Header/ColorModeToggler.jsx
@@ -5,8 +5,8 @@ const ColorModeToggler = () => {
   const { colorMode, toggleColorMode } = useColorMode();
 
   return (
-    <Button size="md" width="70px" justifyContent="center" onClick={toggleColorMode} gap="4" variant="ghost">
-      <Sun size={"2em"} />
+    <Button size="md" p="0px" justifyContent="center" onClick={toggleColorMode} gap="4" variant="ghost">
+      <Sun size={"1.5em"} />
     </Button>
   );
 };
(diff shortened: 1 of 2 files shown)


AbdBarho:
Thank you, looks good.

When the user logs in, it look a bit weird because the user item is rounded.

Could change the user item to be more boxy?

image

AbdBarho:
or maybe move the button next to the language? not sure, whatever you think looks better

theopfr:
I think a nicer solution would be to remove the border around the sun icon to visually seperate it from the other buttons. This also takes care of the problem you just described, take a look

theopfr:
I noticed that when adding this button to the nav bar the mobile version looks even more squeezed and the title overlaps with the buttons. I removed the padding around the sun icon to lower the width and made the "Open Assistant" title slightly responsive by lowering the font size when the viewport gets more narrow.
The title still overlaps with the buttons on the right on mobile but I think some kind of burger menu should be implemented when there are more buttons being added to the navigation bar. As for now, I hope this still looks good to you

AbdBarho (approved):
I guess it is okay like this for now, we can always improve in the future.

PR #1128, opened 2023-02-04
Hotfix for Hungarian language

sedthh:
The fails_task.question was ambigous in Hungarian:
the original translation for "Is it a bad reply, as an answer to the prompt task?" was confusing as the question would require a different answer than yes or no in Hungarian.
The new translation is roughly "Do you think there is something wrong with this message, as it is not an answer to the question?"


AbdBarho (approved):

AbdBarho:
Thanks

PR #1129, opened 2023-02-04
feat(ci): enable `pip` caching in CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument of `actions/setup-python` to enable `pip` caching.


yk (approved):
thank you!

PR #1181, opened 2023-02-05
remove theme button from sidebar

swaptr:
Now that the theme switch button is moved to navbar, we can safely remove this button from sidebar.


AbdBarho:
@swaptr thank you, do you know how it looks on mobile?

swaptr:
Thanks for bringing to my attention.

Before:

Screenshot_20230206_000636

Now:

Screenshot_20230206_000614

AbdBarho (approved):
Thanks

PR #1185, opened 2023-02-05
Expose thumbs up/down buttons in the message table

mehdi-zibout:
Initialized the values of thumbs up/down buttons (the "+1 and "-1" keys in the emojiState objects) to 0. also sorted `Object.entries(emojiState.emojis)` so the location of the buttons is consistent.

### DIFF
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -40,8 +40,11 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
   const router = useRouter();
   const [emojiState, setEmojis] = useState<MessageEmojis>({ emojis: {}, user_emojis: [] });
   useEffect(() => {
+    const emojis = { ...message.emojis };
+    emojis["+1"] = emojis["+1"] || 0;
+    emojis["-1"] = emojis["-1"] || 0;
     setEmojis({
-      emojis: message?.emojis || {},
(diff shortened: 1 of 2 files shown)


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -102,14 +109,18 @@ export function MessageTableEntry({ message, enabled, highlight }: MessageTableE
...
-              <MessageEmojiButton
-                key={emoji}
-                emoji={{ name: emoji, count }}
-                checked={emojiState.user_emojis.includes(emoji)}
-                onClick={() => react(emoji, !emojiState.user_emojis.includes(emoji))}
-              />
-            ))}
+            .sort(([emoji]) => -emoji)
+            .map(([emoji, count]) => {
+              console.log(emoji);
```
AbdBarho, line 114: could you please remove the console log

mehdi-zibout: Sorry I forgot it 😅, while I'm at it should I show the count only when it's greater than 0?

AbdBarho: kind of, yeah, but not for the `+1` or `-1`, otherwise we are back to square one

mehdi-zibout: image
done

AbdBarho (approved):

PR #1186, opened 2023-02-05
add Basque language support

juletx:
Add Basque (eu) language support


AbdBarho (approved):
Thanks!

juletx:
Why are tests failing?

fozziethebeat:
> Why are tests failing?

I don't see any obvious explanation.  Can you first merge main in and resolve the conflicting file?

juletx:
Tests are still failing.

juletx:
> > Why are tests failing?
> 
> I don't see any obvious explanation. Can you first merge main in and resolve the conflicting file?

I've merged main again. I still have no idea why tests are failing.

AbdBarho:
The problem here is very weird:
```
Text content did not match. Server: "Euskara" Client: "Basque"
```
both are correct, I am not sure what should be updated.

juletx:
> The problem here is very weird:
> 
> ```
> Text content did not match. Server: "Euskara" Client: "Basque"
> ```
> 
> both are correct, I am not sure what should be updated.

Yes, very weird. As far as I know, it has only happened with Basque. Apparently, it is a hydration issue, but I have no idea why it happens and how we can solve it.

For now, I have added the translations that were missing.

fozziethebeat (approved):
I have no idea what fixed the tests but this works so I'm merging now! thank you~

fozziethebeat:
Spoke too soon.  Need to figure out what's going on here...

fozziethebeat:
Alright, I tested this locally.  It seems the test failure happens even without the language code addition.  I'm going to merge this and hope for the best.

juletx:
Seems to be working correctly. However, there is a mistake in the language selection menu. "Eurakasa" is wrong, it should be "Euskara".

PR #1197, opened 2023-02-05
fix sender email address

### LINKED ISSUE #1191: Sign in via email link - sender address misspelled
The sender is shown as open-**assistent**@laion.ai  - it should be assistant@

This could make it look like phishing or spam.

swaptr:
The sender email address has typo. This PR fixed that.
Fixes #1191


ansible/deploy-to-node.yaml
```diff
@@ -177,7 +177,7 @@
             "{{ lookup('ansible.builtin.env', 'WEB_DISCORD_CLIENT_ID') }}"
           DISCORD_CLIENT_SECRET:
             "{{ lookup('ansible.builtin.env', 'WEB_DISCORD_CLIENT_SECRET') }}"
-          EMAIL_FROM: open-assistent@laion.ai
```
andreaskoepf, line 180: Sorry, we cannot easily change this.

andreaskoepf (changes requested):

ansible/deploy-to-node.yaml
ArgiesDario: @andreaskoepf could you provide some insight on how this should be fixed? cause I was thinking of fixing it the same way @swaptr did

andreaskoepf (approved):
good news, we now got the correct E-Mail address

ansible/deploy-to-node.yaml
andreaskoepf: @ArgiesDario The problem was not the source change, but creation of the account/alias on the admin side.

PR #1201, opened 2023-02-05
Ukranian

letier3110:
Added Ukranian localization


AbdBarho:
can you also add the key code for ukranian to this array? https://github.com/LAION-AI/Open-Assistant/blob/dfda94f178acfc5dca2870db7032fe7412431a92/website/next-i18next.config.js#L4

AbdBarho (approved):

PR #1203, opened 2023-02-05
Add blog

andrewm4894:
- add blog to docs site.
- create a first example post based on recent youtube video.

Thinking is this can be useful for persistent project announcements and specific blog posts. Also another way that contributors can write up and share research reports etc. as opposed to a proliferation of .md files spread throughout the repo in hard to discover ways. Might also be nice way to incentivise people to get involved and share stuff via blog posts etc too.

Blog ends up looking like this:

image


AbdBarho:
ngl I think it is kinda cool, especially how easy it is to setup.

andreaskoepf (approved):
nice.

Issue #1204, opened 2023-02-05
500 when accepting TOS


ehartford:
500 when Accepting TOS.

I'm logged in with Discord account.

![image](https://user-images.githubusercontent.com/1117701/216842314-6b1b1883-baa7-4b90-b602-45940dc3c336.png)

Koloss0:
Possible duplicate of #1117

andreaskoepf:
Should be fixed now, thanks to @AbdBarho for quick fix. 
Please try again and let us know if it works now. :-)

ehartford:
Works now, thank you

Issue #1117, opened 2023-02-04
Initial Sign-on bug with TOS


Austin-Earl-Private:
If the servers are under a lot of load, the sign-on link doesn't load properly. Thus moving a user to the Dashboard where they are greeted with the TOS. A user will then attempt to agree with the TOS and it will error out. The solution is to set a username in account settings first then accept TOS.

Steps to reproduce:

1. Create new account
2. Dont set UserName
3. Attempt to agree to the TOS
4. Visually nothing happens, the User thinks they are locked out. The console reports many 500 errors.

AbdBarho:
Fixed in https://github.com/LAION-AI/Open-Assistant/pull/1207

PR #1207, opened 2023-02-05
Update ToS Endpoint

AbdBarho:
Create user if it does not exist

Refs #1204
Refs #1117

our `BackendUserCore` only contains the bare minimum, so it won't override any of the other properties.

https://github.com/LAION-AI/Open-Assistant/blob/dcb9dfe952fb0fa6fc11c719c6fc2881de19b29b/website/src/types/Users.ts#L1-L18


andreaskoepf (approved):
lgtm will fix the tos prob.

PR #1215, opened 2023-02-05
Update README.md

vcuculo:


andrewm4894:
`docker compose` is the more modern version: https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose

oh...actually we have an FAQ for this as has come up a few times :)

https://projects.laion.ai/Open-Assistant/docs/faq#docker-compose-instead-of-docker-compose

PR #1217, opened 2023-02-05
Rlhf

sanagno:
(mostly) Placeholder code for RLHF. @theblackcat102  Let me what you think about the structure. I think at this point we should choose one model and start to create scripts that go from start to finish.

### DIFF
--- a/model/supervised_finetuning/configs/config.yaml
+++ b/model/supervised_finetuning/configs/config.yaml
@@ -16,32 +16,32 @@ defaults:
   eval_accumulation_steps:
   freeze_layer:
   datasets:
-    - webgpt
-    - prompt_dialogue
-    - squad_v2
-    - adversarial_qa
-    - trivia_qa_nocontext
-    - xsum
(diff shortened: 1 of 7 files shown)


theblackcat102:
placing under supervised feels odd, I was thinking of placing it under reward/training(or learning ) or a new folder ( RL perhaps )?

sanagno:
That is how I started but then realised that most of the things are/will be already shared, including models, utils and partially pre-processing. We can also think about have some common utils somewhere for all models, whatever you prefer.

theblackcat102:
Hmm, how about renaming the supervised_tuning to like `rlhf` or model training? Since the first step for all related RLHF on language model start with SFT so I guess it makes sense to place our supervised training code inside?

theblackcat102:
But I would suggest doing it in a two step process, create the new directory in the first PR and we move the supervised_tuning code in the second PR. Cause some contributors might be working on the old scheme

sanagno:
Good idea actualy

PR #1236, opened 2023-02-06
fix: removed title on the navbar in the mobile view

MrlolDev:
I make this pr because in mobile devices the navbar have too many elements and the text mix with the language selector, so is better removing the title of the page and using the icon as branding giving more space in the navbar to extra elements.


AbdBarho:
@MrlolDev can you please include screenshots?

MrlolDev:
Before:
imagen
Now:
imagen

fozziethebeat (approved):

PR #1237, opened 2023-02-06
Implemented ability to press ctrl+enter (or cmd+enter) to Review/Subm…

### LINKED ISSUE #811: Keyboard Shortcuts
Would be nice to have a shortcut, something like Ctrl/Cmd+Enter to press Review and Submit.

cooperaustinj:
Implemented ability to press `ctrl+enter` (or `cmd+enter`) to Review/Submit Create prompts

The first press of the shortcut is the same as pressing the Review button. The second press of the shortcut is the same as pressing Submit.

image

Fixes #811


website/src/components/Tasks/CreateTask.tsx
```diff
@@ -55,6 +57,7 @@ export const CreateTask = ({
                 {t(getTypeSafei18nKey(`tasks:${taskType.id}.instruction`))}
               </Text>
             )}
+            <Text color={tipColor}>Tip: Use <Kbd>ctrl</Kbd>+<Kbd>Enter</Kbd> to Review and Submit</Text>
```
AbdBarho, line 60: please extract to localization, also we need to check how it works on macs, we also need to hide this text on mobile.

AbdBarho (changes requested):
Thanks, this is a really big subject, we can start small.

website/src/components/Tasks/CreateTask.tsx
cooperaustinj: Thanks for the feedback. Will swing back around on this tomorrow (or anyone else interested can jump in).

I do also have a Mac to test that `e.metaKey` works as it's supposed to.

cooperaustinj: @AbdBarho how are we doing translations for localization? Simply Google Translate or?

cooperaustinj: Working on Mac: 
image

fozziethebeat:
Would it be possible to resolve the merge conflicts? Then i think we can wrap this PR up and merge it

cooperaustinj:
The localizations are still missing for most languages.

fozziethebeat:
We can deal with localizations separately.  For most changes we add the localizations for the languages the PR author speaks and then over time we'll get the other languages added in.

website/src/components/Tasks/CreateTask.tsx
AbdBarho: Great! no need to worry about translating the text now, my comment with regarding not hard-coding the text in the react component, but just moving it the locale json file, so that it can be translated easily later.

cooperaustinj:
@fozziethebeat Merge conflicts resolved!

andreaskoepf:
@cooperaustinj could you please resolve conflicts one more time?

website/src/components/Survey/TrackedTextarea.tsx
```diff
@@ -50,6 +66,12 @@ export const TrackedTextarea = (props: TrackedTextboxProps) => {
       progressColor = "green";
   }
 
+  const onKeyDown = (e: React.KeyboardEvent<HTMLTextAreaElement>) => {
```
AbdBarho, line 69: `useCallback`?

AbdBarho (commented):
@notmd this code touches upon the tracked text area you created, it looks good from my side, maybe you could give it one final look.

notmd:
@AbdBarho I just pushed the change to allow submission via keyboard. The old implementation only allows previewing the answer only. In order to achieve that, a custom ref attached to the root element of the `CreateTask` component is required to listen for the keyboard event.

AbdBarho (approved):

PR #1258, opened 2023-02-06
Added Persian language locale

Torabi-srh:


fozziethebeat (changes requested):
I see there are changes to `index.tsx` and `delete_user.ts` and `package-lock.json`  Can those changes be undone? For localization changes ideally we just update the locale list and add the json files.

PR #1262, opened 2023-02-06
Add tokenizer config classes

### LINKED ISSUE #990: get_tokenizer in utils.py needs to be cleaned up
def get_tokenizer(tokenizer_name):
    if "t5" in tokenizer_name:  # rankgen
        tokenizer = T5Tokenizer.from_pretrained(tokenizer_name, truncation_side="left")
    else:
        tokenizer = AutoTokenizer.from_pretrained(tokenizer_name)
    if "galactica" in tokenizer_name:
        tokenizer.add_special_tokens({"pad_token": "<pad>", "eos_token": "</s>"})

    return tokenizer

The line of code 
    if "galactica" in tokenizer_name:
will not be executed, either it should be removed, or the "if .. else " branch logic needs to be re-written according to the the original design

jackapbutler:
Fixes #990 

I've just refactored out the tokenizer configuration details from the `get_tokenizer` function. This should hopefully;
* Make it easier to add new tokenizers without changing logic in `get_tokenizer`.
* Make it easier to see what `get_tokenizer` is doing without having to worry about the special tokens of each model type.

### DIFF
--- a/model/supervised_finetuning/utils.py
+++ b/model/supervised_finetuning/utils.py
@@ -35,13 +35,13 @@ TOKENIZER_CONFIGS = {
 }
 
 
-def match_tokenizer_name(name: str) -> TokenizerConfig:
+def match_tokenizer_name(model_name: str) -> TokenizerConfig:
     """Match a partial model name to a tokenizer configuration"""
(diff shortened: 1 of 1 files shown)


sanagno (approved):
Looks good although not too much more simple :))

Issue #1178, opened 2023-02-05
No way to check personal score


VisenDev:
There is no way to see how much score you have earned unless you get onto the leaderboard. Only the top 20 people can see how many points they've earned.

Adding this would help improve gamification.

swaptr:
#1158 will fix this.

jjmachan:
+1 to this issue, would be really nice to have

andreaskoepf:
> Only the top 20 people can see how many points

We will display the personal score .. just one note: There is a next button on the top right on the leaderboards which allows you to page through the top 100 users.

fozziethebeat:
I'm going to close this as a duplicate of #747.  That issue aims to achieve the same goal

PR #1263, opened 2023-02-06
Show current user rank in leaderboard

### LINKED ISSUE #1000: Ranking info
Above or below the ranking should show the stats of the person logged in, and it would also be nice to see how many prompts/responses/labels have already been made overall in this project out of a targeted 50 000.

notmd:
close #1000 maybe #1178 too
- Show current user rank in the leaderboard with +-1 user (only on leaderboard page).
- Extend auto_main script to use random user.
- Support colSpan in the DataTable component (I haven't verified colSpan in header yet, leave that until we need it)
- Refactor OasstError to include the path and request method.

Here is the UI

image


andreaskoepf (approved):
I approve the changes in auto_main .. at some point we should refactor auto_main a bit..

fozziethebeat (approved):

PR #1276, opened 2023-02-06
Clarify what README Docker instructions are for

olliestanley:
Many people enter the Discord thinking that the Docker command is to run the AI model locally and asking how it works. This updates the README instructions to avoid such confusion.


andrewm4894 (approved):

andreaskoepf (approved):
yes, makes sense! thanks for the change.

PR #1278, opened 2023-02-06
Ignore up/down votes (emojis) for own messages

andreaskoepf:
Ignore thumbs up/down add emoji requests from users for their own messages.

(I thought we had an issue for this but couldn't find it...)


notmd:
Can you also expose is_author field to all messages endpoint so the frontend can easily disable thumb up/down button? In later PR is fine

andreaskoepf:
gh does not update the pr although I commit changes and can see them here https://github.com/LAION-AI/Open-Assistant/tree/no_self_up_down_voting will close and retry

PR #1280, opened 2023-02-07
Create  training a transformer-based model

Kidus-berhanu:
Use the built-in torch.nn.functional.cross_entropy method to compute the loss instead of defining your own custom loss function.

Use the inputs.get("attention_mask", None) only if it is not None. This can be done inside the compute_loss method.

Use the torch.max method instead of torch.argmax to find the predictions and labels.

Use the targets.masked_fill_ method instead of the labels[~labels_mask.bool()] = -100 line to mask the padding values in the targets.

### DIFF
--- a/model/supervised_finetuning/training a transformer-based model
+++ b/model/supervised_finetuning/training a transformer-based model
@@ -1,0 +1,117 @@
+import argparse
+from distutils.util import strtobool
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+from transformers import PreTrainedModel, Trainer, TrainingArguments
+from transformers.training_args import OptimizerNames
+from utils import get_dataset, get_loss, get_metrics, get_model, get_tokenizer, read_yamls
+
+
+def compute_metrics(eval_pred, preprocess_fns, metrics):
+    out = {}
+    for metric, preprocess_fn in zip(metrics, preprocess_fns):
+        preds, labels = preprocess_fn(eval_pred)
+        out = dict(**out, **metric.compute(predictions=preds, references=labels))
+
+    return out
+
+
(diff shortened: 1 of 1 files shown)


theblackcat102:
@Kidus-berhanu Have you run a small toy model ( galactica 128m ) with the wandb results?

Kidus-berhanu:
yes, I first preprocessed my data and then trained the "Galactica 128m" model using a deep learning framework PyTorch. During the training process, I integrated W&B to track and visualize the progress of my model, including metrics such as accuracy, loss, and gradients.

After training the model, I was able to use W&B to compare the results of my model with other models I've run and make informed decisions on which model to use for deployment. W&B also allowed me to easily share my results with my team and collaborate on improving the model.
@theblackcat102 it would mean a lot if you would merge my code to the official repository.

theblackcat102:
@Kidus-berhanu Thanks! I just need to make sure the code works on training, some contributors doesn't run their code and I have to fix their bugs. One small problem, could you run pre-commit and make sure its fine to merge?

Kidus-berhanu:
Screenshot 2023-02-08 205427
@theblackcat102, Done! please let me know if there is anything else i could do.

theblackcat102:
@Kidus-berhanu  weird, commit checks still failed

Vechtomov:
The whole pull request looks like it was generated by a language model. @theblackcat102 I think we can close it.

Kidus-berhanu:
@Vechtomov. IT IS NOT GENERTAED BY  language model ,  i surprised on how you got to that conclusion.  wait a little i couldn't find a time  with classes and work ..to run pre-commit again!

sanagno:
Hi @Kidus-berhanu , thanks for the input. I don't think your contributions in the current format are going to help. I would suggest making small changes in the existing code for the issues that you raised.

Issue #879, opened 2023-01-21
Enable Typescript strict mode


notmd:
We need to enable Typescript strict mode to make out code safe. This needs to be incremental cause we don't want to break the build.

- Set TS `strict` option to true in the tsconfig file.
- Going through each file and checking if there is any error need to fix.

We should make small PRs to easily review and avoid conflict.

KarthikRaju391:
@fozziethebeat, I would like to work on this!

EpicLiem:
I'll do a few files

PR #1288, opened 2023-02-07
Ts strict mode

notmd:
part of #879


website/types/next-auth.d.ts
```diff
@@ -17,9 +16,11 @@ declare module "next-auth" {
 declare module "next-auth/jwt" {
   interface JWT {
     /** The user's role. */
-    role?: string;
+    role: string;
     /** True when the user is new. */
     isNew?: boolean;
+
+    sub: string;
```
notmd, line 23: these are guaranteed by the jwt callback

fozziethebeat (approved):

PR #1291, opened 2023-02-07
Force full page reload on language change

### LINKED ISSUE #1287: Possible language changing issues
When doing a initial prompt task in language A , you can switch to language B and complete the task. You will, after this task is completed, be presented with another task that seems to be of language B (not sure how I could check this) but that task would not be available if you would have passed trough the dashboard of language B (no initial prompt tasks are available). By that you are able to create an "unwanted" intitial prompt for language B.
When doing the same manipulation for classify tasks, when the first task in language A is done, then language B task marks as not available (that makes sense as there was none of these task on the dashboard of language B). I don't know about assistant replies. That could possibily explain the occasional wrong language prompts we get.

fozziethebeat:
Fixes #1287 by reloading the page every time a user changes their language.  When a user changes their language during the a task page, it will re-fetch the task and show either a proper task in their language or show that none exist for that task type.

### DIFF
--- a/website/src/components/LanguageSelector/LanguageSelector.tsx
+++ b/website/src/components/LanguageSelector/LanguageSelector.tsx
@@ -37,7 +37,8 @@ const LanguageSelector = () => {
       const locale = option.target.value;
       setCookie("NEXT_LOCALE", locale, { path: "/" });
       const path = router.asPath;
-      return router.push(path, path, { locale });
+      await router.push(path, path, { locale });
+      router.reload();
     },
     [router]
   );


website/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -37,7 +37,8 @@ const LanguageSelector = () => {
       const locale = option.target.value;
       setCookie("NEXT_LOCALE", locale, { path: "/" });
       const path = router.asPath;
-      return router.push(path, path, { locale });
+      await router.push(path, path, { locale });
+      router.reload();
```
rjmacarthy, line 41: Although it's obvious that this will work is there a better way that we can use `mutate` to be re-validate the task data?

fozziethebeat: Right now I want to make it work correctly.  I will leave it to others to find a more elegant solution.

rjmacarthy (approved):

othrayte:
Won't forcing the user to change task potentially throw away the content they have been writing. We currently tell the user to switch language (in a popup) if we detect it doesn't match the language that it will be submitted as.

fozziethebeat:
Note: we discussed this a bit in discord and agreed with this sledgehammer solution.  When we have to deal with intentionally multi-lingual tasks, we'll think about that more carefully.

AbdBarho (approved):

Issue #962, opened 2023-01-27
Messages on the message tab (`/messages`) should be filtered by language


andreaskoepf:
Currently the messages show in on the message tab (`/messages`) are not filtered by the currently active language. I think it would be helpful to add this filtering because this view allows quick and simple message feedback (e.g. emojis & reporting). When many messages are shown in foreign languages which the current user does not understand we lose potential message-review capacity...

The queries for the messages should send the lang-tag of the currently active language as query param to the backend. When the user switches the language (via language selector drop down) the message result lists should be refreshed.

jebbbbb:
Hey, I'm a research engineer currently working on a mix of ml research and software engineering with a focus on language modelling. I'd like to get involved and take this issue if it's still free?

fozziethebeat:
Hi @jebbbbb have you worked with NextAuth localization?  I'm happy to let you take a look at this if you want although I setup this part and I _think_ I know the quick fix to it.

jebbbbb:
Hey @fozziethebeat, no I haven't used it before. In that case it probably makes more sense for you to take it. I also noticed #990 is still open and think I've already got an idea how I would solve it, could you assign me to that one?

jackapbutler:
Hey @fozziethebeat, I've changed GitHub accounts to my personal username. No, I haven't worked with NextAuth localization before. In that case it probably makes more sense for you to take it. I'd be happy to take up another ticket like #990 if you could assign me to it?

fozziethebeat:
I'll get to this shortly

fozziethebeat:
@andreaskoepf to make this work, method `query_messages_ordered_by_created_date` in `prompt_repository.py` has to be updated.  The lang filtering has to happen before ordering (I just tested locally and it currently fails when you fetch recent messages with the lang field set).

I tried to make this in my local PR but for some reason the `check python ast` pre-commit is flagging this file because it uses match (which I guess isn't supported in CPython 3.9.13 that's being run).

Can you fix the backend side? I have a web side fix that'll request it in the user's language for the short term.

PR #1292, opened 2023-02-07
When loading recent messages, filter by the user's stored language.

fozziethebeat:
A simple short term fix.  Applies to #962.  To fully fix a small backend change is needed.

### DIFF
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -313,8 +313,8 @@ export class OasstApiClient {
     return this.get<Message[]>(`/api/v1/messages?${params}`);
   }
 
-  fetch_recent_messages() {
-    return this.get<Message[]>(`/api/v1/messages`);
+  fetch_recent_messages(lang: string) {
+    return this.get<Message[]>(`/api/v1/messages?lang=${lang}`);
   }
 
   fetch_message_children(messageId: string) {
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -314,7 +314,7 @@ export class OasstApiClient {
   }
 
   fetch_recent_messages(lang: string) {
-    return this.get<Message[]>(`/api/v1/messages?lang=${lang}`);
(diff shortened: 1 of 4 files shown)


rjmacarthy (approved):

website/src/lib/oasst_api_client.ts
```diff
@@ -313,8 +313,8 @@ export class OasstApiClient {
     return this.get<Message[]>(`/api/v1/messages?${params}`);
   }
 
-  fetch_recent_messages() {
-    return this.get<Message[]>(`/api/v1/messages`);
+  fetch_recent_messages(lang: string) {
+    return this.get<Message[]>(`/api/v1/messages?lang=${lang}`);
```
notmd, line 317: you can pass an object as a 2nd param of the get method here
```suggestion
    return this.get<Message[]>(`/api/v1/messages`, { lang });
```

AbdBarho:
Would it make sense to update the text in the UI to reflect this?

website/src/lib/oasst_api_client.ts
fozziethebeat: Done!

fozziethebeat:
> Would it make sense to update the text in the UI to reflect this?

Done!

AbdBarho (approved):

website/src/pages/messages/index.tsx
```diff
@@ -24,7 +27,9 @@ const MessagesDashboard = () => {
       <SimpleGrid columns={[1, 1, 1, 1, 1, 2]} gap={4}>
         <Box>
           <Text className="text-2xl font-bold" pb="4">
-            {t("recent_messages")}
+            {t("recent_messages", {
+              language: new Intl.DisplayNames([currentLanguage], { type: "language" }).of(currentLanguage),
```
rjmacarthy, line 31: Maybe it would be a good idea to extract this to a helper method since we use it now in more than one place?

fozziethebeat: Yes, that would make sense.  I'll try to do that in a follow up PR, not this one.

PR #1294, opened 2023-02-07
update the backend path to manage user status

fozziethebeat:
This uses the now proper backend URL path for updating user status


andreaskoepf (approved):
This is an important change. Not sure why it wasn't merged yet.

PR #1296, opened 2023-02-07
website: show language in text area

othrayte:
When the language hasn't been detected (too little text) the language that it will be submitted as is shown. When we detect the language is not what will be submitted we show the detected language in red. The tool tip informs the user that they don't appear to be writing in the correct language.

Note: this replaces the modal dialog that we use to popup if the language appeared to differ.

Also note: We might need to consider the interaction with #1291.
show-lang

RTL language support:
image


website/public/locales/en/language.json
```diff
@@ -0,0 +1,22 @@
+{
```
fozziethebeat, line 1: Let's delete the last sentence in this message and then i'll merge.

fozziethebeat (approved):

website/public/locales/en/language.json
othrayte: Done

PR #1300, opened 2023-02-07
website: "Wrong Language" -> "Not {lang}"

othrayte:
This PR changes the "Wrong Language" labelling flag to instead say "Not English", "Not French" etc depending on the messages tagged language.

Screenshot from 2023-02-08 00-13-03

not-lang

### DIFF
--- a/website/src/components/Messages/LabelFlagGroup.tsx
+++ b/website/src/components/Messages/LabelFlagGroup.tsx
@@ -5,16 +5,29 @@ import { getTypeSafei18nKey } from "src/lib/i18n";
 interface LabelFlagGroupProps {
   values: number[];
   labelNames: string[];
+  expectedLanguage: string;
   isEditable?: boolean;
   onChange: (values: number[]) => void;
 }
 
-export const LabelFlagGroup = ({ values, labelNames, isEditable = true, onChange }: LabelFlagGroupProps) => {
-  const { t } = useTranslation("labelling");
+export const LabelFlagGroup = ({
+  values,
+  labelNames,
+  expectedLanguage,
+  isEditable = true,
+  onChange,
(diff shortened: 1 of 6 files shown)


website/src/components/Messages/LabelFlagGroup.tsx
```diff
@@ -5,16 +5,29 @@ import { getTypeSafei18nKey } from "src/lib/i18n";
...
-  const { t } = useTranslation("labelling");
+export const LabelFlagGroup = ({
+  values,
+  labelNames,
+  expectedLanguage,
+  isEditable = true,
+  onChange,
+}: LabelFlagGroupProps) => {
+  const { t } = useTranslation(["labelling", "language"]);
+  const expectedLanguageName = t(getTypeSafei18nKey(`language:lang_${expectedLanguage}`));
```
rjmacarthy, line 21: We could potentially delete `language.json`. It's seems like another file to maintain when we add/remove locales. Where `locale` is the current code we could try this;

```suggestion
  const expectedLanguageName = new Intl.DisplayNames(locale, { type: "language" }).of(locale)
```

othrayte: Yeah, that would like a much better way to handle it. I won't be able to make that change till after work (18:00+1100), but I think we can do it as a separate change anyway, especially given that file is already in and I'll need to update the other code that uses it. If someone else wants to do it before I'm able to just give me a heads up.

fozziethebeat: I'm good with making this change in a follow up PR

fozziethebeat (approved):

fozziethebeat:
Something broke the pre-commit check.  can you try making a trivial commit to re-trigger it?

Issue #1271, opened 2023-02-06
Similar languages resulting in spam of language change suggestion popup


olliestanley:
![image](https://user-images.githubusercontent.com/1496060/217078836-5fdc28cb-e48c-4204-a84b-9d0c00861a42.png)

softyoda:
Thanks for opening it here for me.

mehdi-zibout:
Hello, I think I can fix this by storing the choice in local storage.

olliestanley:
> Hello, I think I can fix this by storing the choice in local storage.

Sounds good, we don't want it to persist forever though so maybe with some expiry

mehdi-zibout:
@olliestanley of course, for how long do you want it to persist?

olliestanley:
> @olliestanley of course, for how long do you want it to persist?

Not sure to be honest. If you have a length in mind go for it and if needed we can always change it later

mehdi-zibout:
I take back what I said, I don't think that was a good solution. In most cases we shouldn't let someone reply for example in Spanish to an English prompt.
This only poses a problem where the two languages are variant of the same language.
so I guess a better solution is to replace `"ca" === "fr"` by something like `is_variant_of("ca","fr")

olliestanley:
Hmm as in just adding some config for languages we deem similar enough not to show the popup for? That would make sense to me

mehdi-zibout:
@olliestanley  Yes exactly, we need a function that determines whether two languages are really different. So should I write one for the 14 supported languages, or search for one that cover all cases?
The languages are represented in iso6393.

olliestanley:
> @olliestanley Yes exactly, we need a function that determines whether two languages are really different. So should I write one for the 14 supported languages, or search for one that cover all cases? The languages are represented in iso6393.

I would advocate a config file containing a list of pairs which we consider highly similar. This way it will be easy to change without coding anything when new languages are added. Then whenever we detect a language mismatch, if the detected language and the required language form one of the pairs in the config, we don't show the popup.

Dryhb:
I got that one a lot too, got a new one about "ro" , romanian, even tho I is very different from french and, looking at the translation of waht I wrote in romanian doesn't look close at all. That is a different setting than the "ca" to "fr" which are basically the same language since french and romanian seems to be very different.
[...]

PR #1318, opened 2023-02-07
Only check for language conflict on review

mehdi-zibout:
#1271 
Sometimes the language detector mistakes french for other languages, and the users get spammed with a modal whenever they double the word count (starting from 10).
This solution checks for language conflict only when the user submit for review.

### DIFF
--- a/website/src/components/Survey/TrackedTextarea.tsx
+++ b/website/src/components/Survey/TrackedTextarea.tsx
@@ -1,15 +1,13 @@
 import {} from "@chakra-ui/react";
 import lande from "lande";
 import { LanguageAbbreviations } from "src/lib/iso6393";
-import { useCookies } from "react-cookie";
-import React from "react";
+import React, { useEffect } from "react";
 import {
   Progress,
   Stack,
   Textarea,
   TextareaProps,
   useColorModeValue,
-  Button,
   Modal,
   ModalBody,
   ModalCloseButton,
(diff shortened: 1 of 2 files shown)


website/src/components/Survey/TrackedTextarea.tsx
```diff
@@ -114,3 +68,49 @@ export const TrackedTextarea = (props: TrackedTextboxProps) => {
...
+        size="xl"
+        scrollBehavior={"inside"}
+      >
+        {/* we kill the event here to disable drag and drop, since it is in the same container */}
+        <ModalOverlay onMouseDown={killEvent}>
+          <ModalContent alignItems="center">
+            <ModalHeader>Switch Language?</ModalHeader>
+            <ModalCloseButton />
+            <ModalBody>
+              Do you want to switch language? The detected language is <b>{mostProbableLanguage}</b>, whereas your
```
fozziethebeat, line 108: we should change this language to be something like

`We've detected this content to be in {mostProbableLanguage} while your chosen language is {currentLanguage}.  Please write your answer in {currentLanguage} or go back to the <link>dashboard</link> and change your language.`

fozziethebeat (changes requested):

website/src/components/Survey/TrackedTextarea.tsx
mehdi-zibout: I didn't know that a good solution was already implemented (forgot to sync 😅)
image
If you still want to show the modal on review, then I'll do it. 
If not then I think this should be closed

fozziethebeat: If you think the changes that got merged already are sufficient, I'll let you close this out.  Otherwise, I'll review again after the merge conflicts are resolved.

Either way, thank you for putting something together for this problem.

mehdi-zibout: I would only change the progress bar to red in case of conflicting languages, as I think there could be some people who'll miss the letters on the corner. Apart from this I think the implemented solution is great.

Issue #1249, opened 2023-02-06
Add kernel fusion for GELU function for faster training and inference


hyunwoongko:
I will do it firstly. 
cc @ontocord

PR #1322, opened 2023-02-07
Add gelu fusion

hyunwoongko:
Add feature of gelu fusion.

The gelu function was copied from https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/model/fused_bias_gelu.py and modified by me.

---

related issue: #1249


sanagno (approved):
This looks great, thanks!

PR #1323, opened 2023-02-07
update api docs

andrewm4894:
Just wanted to update the api docs since this is still a manual process for now.

- update api docs
- small housekeeping on backend devcontainer


.devcontainer/backend-dev/post_create_command.sh
```diff
@@ -15,7 +15,7 @@ pip install -e .
 cd ..
 
 # docker compose up for backend-dev
-docker compose up backend-dev --build --attach-dependencies -d
+docker compose up backend-dev --build --attach-dependencies
```
andreaskoepf, line 18: Isn’t this in the start-docker.sh … that’s the script I use all the time … (there is now also run-local-no-limit.sh which uses the simpler seed data and disables rate limits.)

andrewm4894: Yeah, for some reason in the dev container it was complaining about the -d when it gets run by the script. 

So in the devcontainer it's just easier to have docker output to the terminal and you can just open new terminals etc. 

I'll take it out just so this is only doing just the api docs just to be cleaner

andreaskoepf (approved):

Issue #968, opened 2023-01-28
Add flagged_message table and admin management endpoints


andreaskoepf:
- Create a "flagged_message" table with message_id (FK, ondelete="CASCADE" similar to emoji table, [see here](https://github.com/LAION-AI/Open-Assistant/blob/19116f7251b366ac0dae5139e2a10bedc1bf7070/backend/oasst_backend/models/message_emoji.py#L16)), `processed` boolean column (not null default false) and `created_date` datetime not null default current_timestamp & create alembic migration script
- insert/update a row in `flagged_message` table whenever the `red_flag` emoji is added to the message, potentially in [PromptRepository.handle_message_emoji()](https://github.com/LAION-AI/Open-Assistant/blob/19116f7251b366ac0dae5139e2a10bedc1bf7070/backend/oasst_backend/prompt_repository.py#L989-L1017) (this happen for text-label requests with `is_report==True`, see [here](https://github.com/LAION-AI/Open-Assistant/blob/b2eb94962c99b3e021d723f9aa877b3681b607f1/backend/oasst_backend/prompt_repository.py#L487)). Check if there is an existing entry in the "flagged_message" table .. if found update it and set processed = False & created_date to current timestamp .  (Do nothing when the flag is removed.)
- Add GET `/admin/flagged_messages/` REST endpoints to query unprocessed reported messages ordered by create_date (with max_count limit)
- Add PUT `/admin/flagged_messages/{message_id}/processed` endpoint to mark the message as processed

Related issue: https://github.com/LAION-AI/Open-Assistant/issues/912

andreaskoepf:
Initial work on this has already been contributed by @GraemeHarris.

For a full solution we need now:
- (backend): Paging for the flag queries
- (web frontend): Admin pages to offer the backend functionality to the user

GraemeHarris:
Will dive into fleshing this out

andreaskoepf:
Since this issue was stale, I unassigned @GraemeHarris and set it to high prio.

GraemeHarris:
I will see if I redeem myself quickly! But if anyone does take it out sooner let me know :)

ruoniw:
Happy to contribute if work still needs to be done here!

PR #1325, opened 2023-02-07
968 add flagged message table and endpoints

GraemeHarris:
Updates for created flagged messages when messages are flagged via the red flag emoji

Issue detail can be seen here: https://github.com/LAION-AI/Open-Assistant/issues/968

Not sure if there's anything in particular I might need to add to admin endpoints but initial testing seems alright. 

Happy to update for any feedback <3


andreaskoepf (approved):
Awesome work .. thanks a lot!

PR #1335, opened 2023-02-08
 improvements with modern language features

marksmayo:
- converted some format strings to f-strings
 - rephrased imports
 - removed unnecessary brackets
 - replaced type() with faster isinstance()
 - included optional parameter maxsplit in split commands

### DIFF
--- a/backend/oasst_backend/api/v1/tasks.py
+++ b/backend/oasst_backend/api/v1/tasks.py
@@ -155,7 +155,7 @@ async def tasks_interaction(
         tm = TreeManager(session, pr)
         ur = UserRepository(session, api_client)
         task = await tm.handle_interaction(interaction)
-        if type(task) is protocol_schema.TaskDone:
+        if isinstance(task, protocol_schema.TaskDone):
             ur.update_user_last_activity(user=pr.user)
         return task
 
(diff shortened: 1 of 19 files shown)


oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -421,7 +421,7 @@ def has_message_id(self) -> bool:
     @pydantic.validator("labels")
     def check_label_values(cls, v):
         for key, value in v.items():
-            if not (0 <= value <= 1):
+            if not 0 <= value <= 1:
```
andreaskoepf, line 424: Removal of brackets while possible does not increases the readability here.

backend/oasst_backend/api/v1/tasks.py
```diff
@@ -155,7 +155,7 @@ async def interaction_tx(session: deps.Session):
         tm = TreeManager(session, pr)
         ur = UserRepository(session, api_client)
         task = await tm.handle_interaction(interaction)
-        if type(task) is protocol_schema.TaskDone:
+        if isinstance(task, protocol_schema.TaskDone):
```
andreaskoepf, line 158: `isinstance()` and `type() is`  are two different things, these checks were exact type checks. The rewritten form is longer than the old.

andreaskoepf:
partially implemented via https://github.com/LAION-AI/Open-Assistant/commit/6c935df3289e63d90ca055286bbefdc83f2b80a0

PR #1327, opened 2023-02-07
website: Fix hydration issues side menu, move side_menu.json to common.json

rjmacarthy:
Fixes #1326 

This PR moves all json in `side_menu.json` to `common.json`.  We have an issue with the side menu layout as it's a high level component and this causes hydration issues with SSR when global namespaces are not passed by default.

Any questions please reach out.

PR #1337, opened 2023-02-08
feat: add Bahasa Indonesia

adityapurwa:
# Context

Adding Bahasa Indonesia (id-ID) language localization.


website/next-i18next.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
       "vi",
       "zh",
       "tr",
+      "id-ID",
```
fozziethebeat, line 23: Other locales are using just the language code.  How essential is the country code here?

fozziethebeat (changes requested):
If possible, I'd prefer we use just `id` for the locale value and directory name.

website/next-i18next.config.js
adityapurwa: Thank you for the feedback! After a second thought, you are right. We don't need the country ID here. I updated the PR to only use language code.

fozziethebeat: Looks like this array entry needs to be updated as well to be `id`.

fozziethebeat (changes requested):

AbdBarho:
Blocked by #1327

website/public/locales/id/side_menu.json
```diff
@@ -0,0 +1,12 @@
+{
```
fozziethebeat, line 1: We had some refactoring in another PR.  Sorry for another ask but can you move all these entries in to `common.json`?

fozziethebeat (changes requested):

website/public/locales/id/side_menu.json
adityapurwa: Thank you for letting me know about the refactor. I moved these into common file in [65e210a](https://github.com/LAION-AI/Open-Assistant/pull/1337/commits/65e210a0e2a93c15e854704971fa8204e4e6e94f) 👍

fozziethebeat (approved):

PR #1359, opened 2023-02-08
added persian language

Torabi-srh:


AbdBarho:
Blocked by #1327

website/public/locales/fa/side_menu.json
```diff
@@ -0,0 +1,12 @@
+{
+  "dashboard": "داشبورد",
```
fozziethebeat, line 2: We had some refactoring in another PR. Sorry for another ask but can you move all these entries in to common.json?

website/next-i18next.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
       "vi",
       "zh",
       "tr",
+      "fa",
```
fozziethebeat, line 23: I see we're not totally ordering these alphabetically.  Could you put this as alphabetically as you can?

fozziethebeat (changes requested):

Issue #1208, opened 2023-02-05
Answer preview bug


XavierRex:
It is impossible to see some of the answers when rating the assistant responses. I tested on Chrome and Opera browser. Moving the UI are caused by the second scroll bar appearing on the right side.

https://user-images.githubusercontent.com/98437003/216843457-7d37dc53-793a-478a-a2bd-7174f21eaae1.mp4

AbdBarho:
Damn...

how often does this happen?

XavierRex:
In almost every task. I do not know if something has been changed, but now when the cursor is on the answer button other answers move and it is possible to see it.

mehdi-zibout:
I don't have this problem in Firefox, but I think a better UX is to make the whole card clickable.

AbdBarho:
@XavierRex one thing that confuses me is that you have two scroll bars, can you tell me your screen zoom scaling? is it 200%?

>  I don't have this problem in Firefox, but I think a better UX is to make the whole card clickable.

@mehdi-zibout we considered this, but you still have to be able to drag and drop the card without the window openning.

mehdi-zibout:
> @XavierRex one thing that confuses me is that you have two scroll bars, can you tell me your screen zoom scaling? is it 200%?
> 
> > I don't have this problem in Firefox, but I think a better UX is to make the whole card clickable.
> 
> @mehdi-zibout we considered this, but you still have to be able to drag and drop the card without the window openning.

yes it's possible in dnd-kit by modifying some of the sensors options.

XavierRex:
@AbdBarho No, my zoom scaling is 100%

andreaskoepf:
IMO there are multiple problems with the current ranking UI, e.g. the popup poorly uses the available screen space:
![image](https://user-images.githubusercontent.com/9976399/216846885-44be4252-fe68-4d7f-8c8f-1aaee60cebff.png)

In the image it looks as if I could scroll .. but as soon as I move my mouse after scrolling the popup disappears.

CactiStaccingCrane:
The better option is to move the [...] next to the drag icon. Problem solved, with an added benefit of not needing to move the mouse a lot to see the expanded message.

Lyrialtus:
[...]

PR #1386, opened 2023-02-09
Making cards clickable in Ranking replies page

mehdi-zibout:
Users were having problem with the [...] button #1208 
My proposed solution is to remove it and make the card clickable
Recording 2023-02-09 at 12 24 33
I changed the sensor options so that the dragging will start after the pointer moves by 10 pixels. 
I have tested this in my phone, and a laptop screen touch and so far haven't found any conflict between dragging and clicking.

Only the text area is clickable, in case of a problem one can use the handle. 
image

### DIFF
--- a/website/src/components/CollapsableText.tsx
+++ b/website/src/components/CollapsableText.tsx
@@ -1,16 +1,16 @@
-import React, { ReactNode } from "react";
+import { ReactNode } from "react";
 
-export const CollapsableText = ({
-  text,
-  maxLength = 220,
-}: {
-  text: ReactNode;
-  maxLength?: number;
-  isDisabled?: boolean;
-}) => {
+export const CollapsableText = ({ text, maxLength = 220 }: { text: ReactNode; maxLength?: number }) => {
   if (typeof text !== "string" || text.length <= maxLength) {
     return <>{text}</>;
   } else {
-    return <span>{text.substring(0, maxLength - 3)}...</span>;
+    const visibleText = text.substring(0, maxLength - 3);
+    const dots = visibleText[visibleText.length - 1] === "." ? ".." : "...";
+    return (
+      <span>
+        {visibleText}
+        {dots}
+      </span>
+    );
   }
 };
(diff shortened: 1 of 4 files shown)


fozziethebeat (changes requested):
Can you investigate the cypress test failure? It seems to be for the random task type and not a random test failure.  Before we merge this I'd like to see that test pass.

mehdi-zibout:
@fozziethebeat it was because cypress test clicks on the middle of the card (to get it focused so it can be dragged by keyboard), but that causes full text window to open. I changed the test to click on the left side and it passes now.

mehdi-zibout (commented):
@fozziethebeat  it was because cypress test clicks on the middle of the card (to get it focused so it can be dragged by keyboard), but that causes full text window to open. I changed the test to click on the left side and it passes now.

website/src/components/CollapsableText.tsx
```diff
@@ -1,16 +1,16 @@
...
-  maxLength?: number;
-  isDisabled?: boolean;
-}) => {
+export const CollapsableText = ({ text, maxLength = 220 }: { text: ReactNode; maxLength?: number }) => {
   if (typeof text !== "string" || text.length <= maxLength) {
     return <>{text}</>;
   } else {
-    return <span>{text.substring(0, maxLength - 3)}...</span>;
+    const visibleText = text.substring(0, maxLength - 3);
+    const dots = visibleText[visibleText.length - 1] === "." ? ".." : "...";
```
AbdBarho, line 8: There is no need for this check, it is not problem.

also, it does not handle all cases, what if your text ends with two dots? with three?

as said, not needed.

AbdBarho (commented):
One final comment

website/src/components/CollapsableText.tsx
mehdi-zibout: Okay removed it.
I have a question in `SortableItem.tsx` we're using `useMemo` to change the background when the card is active. Isn't this overkill? we could use Tailwindcss `hover:` for this and for the cursor as well.

AbdBarho (approved):
Thanks

website/src/components/CollapsableText.tsx
AbdBarho: Yes, our code needs simplification, feel free to clean up if you find anything worthy, preferably in another PR.

fozziethebeat (approved):

PR #1387, opened 2023-02-09
Repared Some errors in arabic localization

ParisNeo:
I have spot some errors in the arabic locale ui. So I proposed some upgrades


AbdBarho:
does this conflicts with #1385?

ParisNeo:
I first did a pull then upgraded the locale. I have solved all the conflicts before doing the pull request. Some of my updates were already taken care of by the pull request you mensioned.

fozziethebeat (changes requested):
I see a few merge conflicts that need to be addressed.  Will review and approve after that

ParisNeo:
can someone review my pull request? I have merged it with the incoming updates fron issue #1385

AbdBarho:
@ParisNeo I see a lot of python file changes, nothing related to arabic UI translation, can you explain what this PR is supposed to do?

ParisNeo:
OK,
Sorry, This pr was pending for quite a while now and i was also working on another pull request and somehow I merged both which created a mess.
This PR concerns only updates in the UI localization files. 

My other Pr (thre one that includes python files) concerns the addition of an auto translation module that is not built to translate the UI but to translate messages from the database. The idea is to solve the disparity between languages in terms of number of prompts and answers by automatically translating messages from one language to another using an automated github workflow for example or a cron or whatever.  The users can then rate those translations and enhance them. Thus people from different languages contribute once in their language and then we automatically enritch the other languages. I use facebook's MBart50 model from transformers library. It can support up to 50 languages which is sufficient in our case.

fozziethebeat (changes requested):
I see only python changes in this PR now.  Is this PR still relevant now or can we close it given that the referenced localization changes went it?

PR #1393, opened 2023-02-09
Admin user management

notmd:
part of #1022 
- Allow updating the show_on_leaderboard field.
- Add raw JSON of the user object.
- Add a new user message table.
- Also fixed style issue: hard to see reaction count when the user also reacted.
- Rename MessageTable to MessageConversation.

UI
image
image


website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -0,0 +1,110 @@
...
+    cell: ({ getValue }) => {
+      return <DateDiff>{getValue()}</DateDiff>;
+    },
+  }),
+  columnHelper.accessor((row) => row.id, {
+    header: "Actions",
+    cell: ({ getValue }) => (
+      <DataTableAction
+        as={NextLink}
+        href={ROUTES.ADMIN_MESSAGE_DETAIL(getValue())}
```
notmd, line 90: this page doesn't exists yet, I will add it in the next PR

website/src/components/DataTable/useCursorPagination.ts
```diff
@@ -0,0 +1,40 @@
...
+  };
+
+  const toNextPage = (data: undefined | { prev?: string; next?: string }) => {
+    setPagination({
+      cursor: data?.next || "",
+      direction: "forward",
+    });
+  };
+
+  const resetCursor = () => setPagination((old) => ({ ...old, cursor: "" }));
```
AbdBarho, line 32: `useCallback` for these 3 functions.

website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -0,0 +1,110 @@
+import { Avatar } from "@chakra-ui/avatar";
+import { Badge, Flex } from "@chakra-ui/layout";
+import { Tooltip } from "@chakra-ui/react";
+import { createColumnHelper } from "@tanstack/table-core";
+import { formatDistanceToNow, formatISO9075 } from "date-fns";
```
AbdBarho, line 5: I wonder if this is something that `next-i18n` already provides.

AbdBarho (commented):
Thanks! it would be great if you could make upcoming PRs smaller than this, it is easier to review.

website/src/components/Messages/AdminMessageTable.tsx
fozziethebeat: From my knowledge, the answer is no.  I'm pretty sure we'd have to use some new library to do this so date-fns seems good enough

fozziethebeat (commented):
Since @AbdBarho left some good comments, I have nothing else to add.  Looks fine after those changes.

fozziethebeat (approved):
By special request by Andreas I'm going to approve and merge this.  Maybe make the cleanups in a follow up PR?

website/src/components/DataTable/useCursorPagination.ts
notmd: These 3 function will need a wrapper function to pass data, so useCallback has no effect here.

PR #1398, opened 2023-02-09
Add tests and update docstring to tokenizer matching

jackapbutler:
This PR does performs some minor updates;

* Adds tests for the matching logic within `get_tokenizer`
* Updates docstring with an example to make the logic clear

### DIFF
--- a/model/supervised_finetuning/utils.py
+++ b/model/supervised_finetuning/utils.py
@@ -37,7 +37,7 @@ TOKENIZER_CONFIGS = {
 
 def match_tokenizer_name(model_name: str) -> TokenizerConfig:
     """Match a partial model name to a tokenizer configuration"""
-    tokenizer_config_matches = [config for name, config in TOKENIZER_CONFIGS.items() if name in model_name]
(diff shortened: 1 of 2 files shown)


model/supervised_finetuning/utils.py
```diff
@@ -37,7 +37,7 @@ class TokenizerConfig(NamedTuple):
 
 def match_tokenizer_name(model_name: str) -> TokenizerConfig:
     """Match a partial model name to a tokenizer configuration"""
-    tokenizer_config_matches = [config for name, config in TOKENIZER_CONFIGS.items() if name in model_name]
+    tokenizer_config_matches = [config for name, config in TOKENIZER_CONFIGS.items() if model_name in name]
```
sanagno, line 40: It was already correct, no? We call `match_tokenizer_name` based on the name of the model not the keys of TOKENIZER_CONFIGS

jackapbutler: Hey @sanagno, yes you're right sorry 😂. I've just updated it back and added some docstrings / tests to make it clearer.

sanagno (approved):
Looks great :)!

PR #1404, opened 2023-02-09
Basic chat UI

AbdBarho:
~~Blocked by #1418~~

This feature is currently hidden behind a flag, so deployment is not a problem, also, there are no inference servers!

https://user-images.githubusercontent.com/24505302/217936648-09e79be0-a04e-475d-82bd-50bfd11a8d21.mp4


aayojezuzchrezt:
(app)

website/src/pages/api/chat/index.ts
```diff
@@ -0,0 +1,12 @@
+import { post } from "src/lib/api";
+import { withoutRole } from "src/lib/auth";
+
+// TODO: move to env
+export const INFERENCE_URL = "http://localhost:8000";
+
+const handler = withoutRole("banned", async (req, res, token) => {
```
fozziethebeat, line 7: Do we want to try calling the inference service directly from the client side without forwarding it through the next api routes?

AbdBarho: I only proxy to make sure the user is authenticated & authorized, not sure what other alternatives we have.

fozziethebeat: that makes sense. i  need to get unlazy and setup JWT handling on the inference server.  Then we can rely on the JWT contents for this verification.

fozziethebeat (approved):
Web changes look solid.  Looks like something changed in `requirements.txt` tho.

AbdBarho:
> Web changes look solid. Looks like something changed in `requirements.txt` tho.

Fixed, sorry for the long delay, busy with work stuff!

andreaskoepf:
There are some minor conflicts now .. @AbdBarho could you check?

AbdBarho:
@andreaskoepf resolved!

yk (approved):
thank you!

PR #1399, opened 2023-02-09
I believe this is what is needed for Danish

lytjedk:
I believe that this will enable Danish as a language at the website.
And this would solve #1369 

I am new to Git. So I might have done something wrong though. :)

PR #1406, opened 2023-02-09
 Danish language - fixes #1399 - Trying again

lytjedk:


lytjedk:
And again it just doesn't see that I did the pre-commit locally.

PR #1413, opened 2023-02-10
Fix a bitwise operator

tomohideshibata:
I got the following error at [`model/supervised_finetuning/custom_datasets/summarization.py#L60`](https://github.com/LAION-AI/Open-Assistant/blob/main/model/supervised_finetuning/custom_datasets/summarization.py#L60):

> TypeError: unsupported operand type(s) for &: 'list' and 'str'

### DIFF
--- a/model/supervised_finetuning/custom_datasets/summarization.py
+++ b/model/supervised_finetuning/custom_datasets/summarization.py
@@ -57,7 +57,7 @@ def index_summary_merge(text, summary):
 class SummarizationDataset(Dataset):
     def __init__(self, dataset, cache_dir, split, max_words=512):
         self.name = dataset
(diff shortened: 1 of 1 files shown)


theblackcat102:
@tomohideshibata  not sure what's the difference between the two, can you extrapolate?

tomohideshibata:
@theblackcat102 Thanks for your response.
First, the current code does not work due to the operator precedence.
The following code would work, but I think it is normal to use `and` instead of `&`.
```python
if (summarization_config_mapping[dataset][0] in ["billsum", "tldr_news"]) & (split == "validation"):
    ...
```

tomohideshibata:
This is fixed via https://github.com/LAION-AI/Open-Assistant/pull/1455.

PR #1414, opened 2023-02-10
feat: romanian

MrlolDev:
Added full romanian translation, I am not native speaker so it can has some errors of expressions but the same happens with other languages.


MrlolDev:
Any help with the error pls?

AbdBarho:
@MrlolDev rebase your code on top of main, the error should be fixed.

website/public/locales/ro/side_menu.json
```diff
@@ -0,0 +1,12 @@
+{
+  "dashboard": "Panou de control",
```
fozziethebeat, line 2: I believe this file was removed.  Can you move these entries into `common.json`?

fozziethebeat (changes requested):
If possible, could up merge in the recent changes from main? That should fix the end2end test failure.

AbdBarho (approved):
Thanks

fozziethebeat (approved):

PR #1415, opened 2023-02-10
Fixing basque language name hydration issues on Google Chrome

fozziethebeat:
When we try to render the Basque language name on various browsers, we sometimes trigger a hydration issue since the pre-rendered name can be different.  This special cases "eu" to always return the same name regardless to ensure no hydration issues exist.  

Additionally, this refactors all code fetching language display names to use a shared method.


AbdBarho (approved):

igorlor:
> When we try to render the Basque language name on various browsers, we sometimes trigger a hydration issue since the pre-rendered name can be different. This special cases "eu" to always return the same name regardless to ensure no hydration issues exist.
> 
> Additionally, this refactors all code fetching language display names to use a shared method.

There is a mistake in the languages.ts, it should be "Euskara" instead of "Eurakasa". Thanks in advance.

PR #1417, opened 2023-02-10
Extract more strings for localization

GuilleHoardings:
Only English and Spanish translations added.


website/src/components/Buttons/Skip.tsx
```diff
@@ -11,13 +11,15 @@ import {
   Textarea,
   useDisclosure,
 } from "@chakra-ui/react";
+import { useTranslation } from "react-i18next";
```
AbdBarho, line 14: `"next-i18next"`?

AbdBarho (commented):
Thanks!

website/src/components/Buttons/Skip.tsx
GuilleHoardings: Oops, yes. Fixed!

website/src/components/Survey/LabelLikertGroup.tsx
```diff
@@ -18,160 +19,154 @@ interface LabelInfo {
   inverted: boolean;
 }
 
-const getLabelInfo = (label: string): LabelInfo => {
+const getLabelInfo = (label: string, t: (key: string) => string): LabelInfo => {
```
notmd, line 22: This is not typesafe? Could you look more into it? Mabe the following will work
`t: ReturnType<typeof useTranslation>["t"]`

GuilleHoardings: I've changed it. I don't know typescript, so I don't know if it's right. It seems to work.

GuilleHoardings: It looks like the E2E tests have failed. I've taken a look, but I don't understand why, sorry.

website/src/components/Buttons/Skip.tsx
```diff
@@ -11,13 +11,15 @@ import {
...
 } from "@chakra-ui/react";
+import { useTranslation } from "next-i18next";
 import { useState } from "react";
 
 interface SkipButtonProps extends ButtonProps {
   onSkip: (reason: string) => void;
 }
 
 export const SkipButton = ({ onSkip, ...props }: SkipButtonProps) => {
+  const { t } = useTranslation("common");
```
rjmacarthy, line 22: I'm not sure, but maybe you don't need to specify the namespace when the only one used is `common`?

website/src/components/Survey/TaskControls.tsx
```diff
@@ -1,4 +1,5 @@
 import { Box, Flex, IconButton, Progress, Tooltip, useColorModeValue } from "@chakra-ui/react";
+import { useTranslation } from "react-i18next";
```
rjmacarthy, line 2:
```suggestion
import { useTranslation } from "next-i18next";
```

website/src/components/TaskInfo/TaskInfo.tsx
```diff
@@ -1,4 +1,5 @@
 import { Box, Flex, Text, TextProps, useColorModeValue } from "@chakra-ui/react";
+import { useTranslation } from "react-i18next";
```
rjmacarthy, line 2:
```suggestion
import { useTranslation } from "next-i18next";
```

website/public/locales/en/common.json
```diff
@@ -2,6 +2,7 @@
   "about": "About",
   "account_settings": "Account",
   "admin_dashboard": "Admin Dashboard",
+  "any_feedback_on_this_task": "Any feedback on this task?",
```
rjmacarthy, line 5: Seems like not a common utterance, maybe better off in `tasks.json`?

website/src/components/Buttons/Skip.tsx
GuilleHoardings: You are right. I've changed it in all the files that only used `common`.

website/src/components/Survey/TaskControls.tsx
GuilleHoardings: Thank you. I've fixed all the `"react-i18next" ` that I added in my commit. However, there's still one in `src/pages/account/index.tsx` that was inserted by @AbdBarho and I don't know if it should be changed.

website/src/components/TaskInfo/TaskInfo.tsx
GuilleHoardings: See previous comment.

website/src/components/Survey/TaskControls.tsx
notmd: > Thank you. I've fixed all the `"react-i18next" ` that I added in my commit. However, there's still one in `src/pages/account/index.tsx` that was inserted by @AbdBarho and I don't know if it should be changed.

Yes it should be update too

website/public/locales/en/common.json
GuilleHoardings: Makes sense. I've updated it. Thank you again!

website/src/components/Survey/TaskControls.tsx
GuilleHoardings: Changed. Thank you!

website/src/components/Survey/LabelLikertGroup.tsx
GuilleHoardings: @notmd I've been working on this. I've found that the solution you proposed causes an infinite loop in the type checking and that was the cause of the E2E test failing.

I've gone back to my initial solution. I think that it is type safe. As I understand it, it's saying that parameter `t` is a function that receives a `string` and returns a `string`. Isn't that correct?

notmd: So I'm fine with the current behavior. Thanks

fozziethebeat (approved):

fozziethebeat:
It looks like all change requests have been applied or at least attempted.  Merging in.

PR #1418, opened 2023-02-10
Use docker compose profiles for all services

AbdBarho:
Similar to how it is done for the inference services.

I hope I didn't forget to update any file.

I don't think this affects deployment


fozziethebeat (approved):
Web changes approve!

andreaskoepf (approved):
let's give it a try.

PR #1419, opened 2023-02-10
add moderator role

notmd:
- Add a new moderator role (they have the same permission as the admin now except for viewing sensitive system parameters).
- Add useHasAnyRole hook.
- Add withAnyRole API handler helper.
- Show role in UserMenu when the user is admin or moderator.

image


fozziethebeat (approved):
We finally have more roles!

Issue #1364, opened 2023-02-08
Rank Assistant Replies lacks the ability to see line breaks.


phly95:
![image](https://user-images.githubusercontent.com/3526540/217560521-b52bfca5-185b-434e-a3ff-7f7f66fe861f.png)
See how if the response is supposed to have line breaks, you cannot see them. This means I cannot accurately perform this task.

Dryhb:
Also appears on things like ascii drawings which because even less visible (mostly a vertical thing) for ex : 43aac56c-fc2c-439a-848d-27422d5785b4

rjmacarthy:
I can fix this, please assign to me.

PR #1421, opened 2023-02-10
website: Fix SortableItem whitespace and improve cursor handling

rjmacarthy:
Hey, this pr fixes issue #1364 and also applies some refactoring of the code in `SortableItem`.

Here is a a summary of the changes:

- Replace `useState` hook to manage the mouse cursor.  Instead, we can use css.
- Added a background color indicator when the item is in active state.
- Added and updated the stories for `Sortable` to support these changes.

Any problems please let me know.

### DIFF
--- a/website/src/components/Sortable/SortableItem.tsx
+++ b/website/src/components/Sortable/SortableItem.tsx
@@ -2,7 +2,7 @@ import { Box, useColorModeValue } from "@chakra-ui/react";
 import { useSortable } from "@dnd-kit/sortable";
 import { CSS } from "@dnd-kit/utilities";
 import { GripVertical } from "lucide-react";
-import { PropsWithChildren, useState } from "react";
+import { PropsWithChildren, useMemo } from "react";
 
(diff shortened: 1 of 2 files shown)


website/src/components/Sortable/SortableItem.tsx
```diff
@@ -13,37 +13,36 @@ export const SortableItem = ({
...
   const disabledBackgroundColor = useColorModeValue("gray.400", "gray.700");
+  const activeBackgroundColor = useColorModeValue("gray.600", "gray.600");
   const textColor = useColorModeValue("white", "white");
 
   const { attributes, listeners, setNodeRef, transform, transition } = useSortable({ id, disabled: !isEditable });
 
-  const style = {
-    transform: CSS.Translate.toString(transform),
-    transition,
+  const sx = {
```
AbdBarho, line 21: can you `useMemo` here?

fozziethebeat (approved):
Thank you for using `useMemo`~

PR #1431, opened 2023-02-10
Update translation "view_user" in /de/message.json

Jakolo121:
I've updated the "view_user" from "user anzeigen" to "Benutzer anzeigen"


AbdBarho (approved):
Dankeschön!

Issue #397, opened 2023-01-05
Get model evaluation working on the reward model trainer


bth5032:
Based on #313 , we are having issues with model evaluation in the reward model trainer (code in model/ranking). It seems that the evaluation results are not being computed and logged for some reason. 

Preliminary research ~mentioned in this post https://github.com/LAION-AI/Open-Assistant/pull/313#issuecomment-1370430372, to summarize it may have something to do with the "evaluation_strategy" argument in TrainerArgs.~ was incorrect, see comment below.

It is extremely important that we have a normalized and robust test suite for our models in order to help select between them for large scale training. I've discussed this issue with @huyouare on discord and he seems to want to take it over. Will leave it to him to intro himself.

pinging @theblackcat102

bth5032:
Update, it seems like this issue is due to rankgen model not returning labels [here](https://github.com/LAION-AI/Open-Assistant/blob/main/model/reward/instructor/trainer.py#L32). 

To fix this issue simply return the labels (the label is [always 0](https://github.com/LAION-AI/Open-Assistant/blob/main/model/reward/instructor/trainer.py#L114), this is not an issue because we always run positive and negative batches through the model, then just construct the logits after the fact, this passes no information to the model) and run training with

`python trainer.py configs/rankgen-t5-base.yml`

to validate accuracy is being logged.

The code can be further refactored to make it a bit cleaner.

jackapbutler:
> Hey, I'm a research engineer currently working on a mix of ml research and software engineering with a focus on language modelling. I'd like to get involved and was wondering if this issue is free?

I've changed my username to this one.

bth5032:
@jackapbutler Yes this task is still open, but I'm not sure if the code in the repo is up to date. @theblackcat102 would know better. 

Also, I'll mention I did try to fix this issue using the fix I mentioned in the last post but it didn't seem to work, so just a heads up there.

jackapbutler:
Cool @bth5032 and @theblackcat102 I'm happy to take a look at this and flag up any issues I find in the process (assuming the problem still exists).

jackapbutler:
Hi, currently I'm spending most of my time on #1621 and don't think I'll be able to get to this soon, thought I would un-assign myself and others can chime in if they want to take a look 👍

maw501:
I can pick this up.

PR #1438, opened 2023-02-10
Update reward readme

### LINKED ISSUE #1008: reward/instructor/trainer.py needs to handle default values of config vars in training_conf (loaded from .yml file)
I have tried to run the training "python trainer.py configs/electra-base-dis-webgpt.yml" and ran into an issue of
"Keyerror: scheduler" at line 140
         lr_scheduler_type=training_conf["scheduler"]
Further inspection revealed that it is caused by missing "scheduler" from "configs/electra-base-dis-webgpt.yml". As a result, the key to index the training_conf data structure does not exist (the "scheduler" in this case). The current version of trainer.py simply throws an error in that case.  I suggest that we modify the behavior of trainer.py so that it uses a default value for a certain config var if it is not defined in the .yml file. This should apply to all such config vars, and "scheduler" happens to be the one of them I just spotted

jackapbutler:
* fixes #1008 by following the most recent advice in that thread

This small PR just adds a `dummy.yml` file containing all possible configuration options in the reward model training as well as updating the `README.md` so people are aware they need to set this up before they run the `trainer.py`.

### DIFF
--- a/model/reward/instructor/README.md
+++ b/model/reward/instructor/README.md
@@ -11,7 +11,8 @@ pip install -r requirements.txt
 Write or inherit a `configs/<config-name>.yml` file to store training
 configuration details.
 
-> The configuration file must have _at least_ all the keys present in dummy.yml
+> The configuration file must have _at least_ all the keys present in
+> [`configs/dummy.yml`](model/reward/instructor/configs/dummy.yml)
 
 Run training procedure
 
--- a/model/reward/instructor/README.md
+++ b/model/reward/instructor/README.md
@@ -2,20 +2,21 @@
 
 Trainer code based on huggingface. Compatible with deepspeed or accelerate
 
-Requirements
-
-```
-wandb
-evaluate
-datasets
-transformers
-torch==1.12
-```
-
-Start training reward model
+Install Python requirements
 
 ```bash
-python trainer.py configs/electra-base-dis-webgpt.yml
+pip install -r requirements.txt
+```
+
+Write or inherit a `configs/<config-name>.yml` file to store training
+configuration details.
+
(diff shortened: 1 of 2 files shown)


sanagno (approved):
Thanks!

henryj18:
Nice job and kudos to your guys, appreciate your hard work!

On retrospect, could we streamline the process by not opening a new issue, but instead placing the people working on 1438 into the team to fix 1008? That can help reduce the complexity of managing multiple github issues on the exact same issue, and help to save repeated work. I was assigned 1008 earlier, and got notified that 1438 fixed 1008 only when 1438 was done and merged. As a result, the work I have done on 1008 and was just about to push into github become repeated. It would be great if we can put into place some process to avoid such repeated work in the future.

sanagno:
That is a great point @henryj18, and apologies for the pure communication. Feel free to incorporate your work into the project now. We are in a stage where we need to move fast, and minor issues can be resolved faster. We can try to be more careful in the future!

henryj18:
@sanagno Thanks very for your update, appreciate that. Since you guys have fixed the issue, there is no need for me to submit further work unless I find something new. I understand that you guys were under the pressure to close issues fast. Cordially look forward to some potential future collaborations.

jackapbutler:
Hey @henryj18 apologies I just scanned through the thread and didn’t realise it was assigned to you. I was looking at similar files while investigating #397 and just made the change on the fly. I’ll be mindful in the future to check back on the issues for who’s assigned 🙂

henryj18:
@jackapbutler Thanks for your update. No need to apologize, as I said above "I understand that you guys were under the pressure to close issues fast. Cordially look forward to some potential future collaborations." 🙂

PR #1449, opened 2023-02-10
Add website translations to Latin American Spanish

SrWesly:
Add Latin American Spanish.


fozziethebeat (changes requested):
This change requires some thought, as do two other locales that I didn't notice the language code carefully enough.

Right now we the UI locale for both the UI presentation and task selection. This would split the data for "es" into two subsets: "es" and "es-419".  I'm not entirely sure that's the right decision so I want to make sure the web and backend team discuss this.

SrWesly:
There are differences between Latin American Spanish and Spanish from Spain, as well as a deeper difference than, for example, British English and U.S. English. Among these differences is the use of different verb tenses to express past actions. Similar to Portuguese and Brazilian Portuguese (see #1798), it should avoid being combined into a single language.

mguinhos:
Coff coff

PR #1453, opened 2023-02-11
implement admin message detail page

notmd:
- Add admin message detail page.
- Add new message tree UI and use/tree endpoint to receive data.
- New message tree only being used in the admin panel. Not in the user area yet.
- The current tree UI display is not really well in mobile when the depth > 4. I will address this in the next PR.

Message tree UI
image

### DIFF
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -1,5 +1,6 @@
 import {
   Avatar,
+  AvatarProps,
   Box,
   HStack,
   Menu,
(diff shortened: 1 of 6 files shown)


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -34,9 +47,17 @@ interface MessageTableEntryProps {
   message: Message;
   enabled?: boolean;
   highlight?: boolean;
+  avartarPosition?: "middle" | "top";
+  avartarProps?: AvatarProps;
```
notmd, line 51: These props are just to make the new UI compatible with the current UI. Will be cleanup in next PR

fozziethebeat (approved):

PR #1454, opened 2023-02-11
Example of dataset contribution

Vechtomov:
This is an example of dataset contribution. Comments are welcome.
Original dataset: OpenAI's [Grade School Math](https://github.com/openai/grade-school-math).


andreaskoepf (approved):
Thanks a lot .. ;-) .. and sorry for not merging it earlier, I wasn't sure from the title if it was a demo PR that should be kept open as an example..

PR #1455, opened 2023-02-11
Add more datasets and some fixes

theblackcat102:
1. Mainly added OA private datasets :

- [essay_instruction](https://huggingface.co/datasets/ChristophSchuhmann/essays-with-instructions) : essay writing
- private_tuning : new instruction generated dataset 
- translated instruction : translated instruction of the above
- [eli5](https://huggingface.co/datasets/eli5) : askh, eli5, asks dataset

2. Added collate function for mixing different tasks during in batch to make sure the attention learns to attention relevant previous context and question

3. Fix bugs from other PR


sanagno:
Thanks!
We can merge, I would propose some simple changes.
1. Add a flag to choose whether to use the new mixing in the collator
2. Add a script to split the OAPrivate data for sft, rm and rl
3. I would merge all the data utils from the sft, rm and rl, let me know if you agree I can do it today-tomorrow.

model/supervised_finetuning/custom_datasets/dialogue_collator.py
```diff
@@ -76,3 +77,108 @@ def __call__(self, features):
...
+    def __call__(self, features):
+        flatten_messages = []
+        label_masks = []
+        total_short_context = 0
+        for messages in features:
+            messages = list(messages)
+
+            # Add a way for the model to terminate generation
+            # When we predict the start of a new expected question, we want to be able to stop generation
+            messages.append(self.tokenizer.eos_token)
```
sanagno, line 104: For this collator to work, we need to replace this eos token with the <Question> tag. Let me know if I am wrong.

sanagno: I would change it anyway, for both this collator and the default collator

sanagno (approved):
We can merge for now, and I can make some changes now

PR #1482, opened 2023-02-11
use new message tree in user area

### LINKED ISSUE #1332: Remove refresh of public message tree view
After navigating to a longer tree by id, e.g. https://open-assistant.io/messages/8d5fdbdf-3c3b-487f-88c1-ddde1ae8bcb2
and scrolling down the page refreshes after a couple of seconds. This makes it impossible to read through the messages as the refresh resets the scroll position back to the top. In general I don't think that auto-refresh is helpful on this page-type, i.e. it could simply be removed without losing any important functionality.

Relevant code ref:
https://github.com/LAION-AI/Open-Assistant/blob/66b7ed2b9beba9a504792e8b7aab33494df78ff6/website/src/pages/messages/%5Bid%5D/index.tsx#L18-L20

notmd:
close #1332 
- Use the new message tree in the user area.
- Add author (only show in message tree), deleted and spam badge (only show when the user is mod or admin)
- The tree still looks not good on mobile when it's too deep. My idea is to limit the ancestor and descendant but let's see the feedback from users.
<details>
  <summary>Tree UI in mobile</summary>

  image
</details>
<details>
  <summary>It's still not perfect when the tree is too deep. Here is the tree that has 9 level.
</summary>

image
</details>


website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -281,9 +311,11 @@ const MessageActions = ({
...
               {t("view_user")}
             </MenuItem>
-            <MenuItem onClick={handleDelete} icon={<Trash />}>
-              {t("common:delete")}
-            </MenuItem>
+            {!message.deleted && (
+              <MenuItem onClick={handleDelete} icon={<Trash />}>
+                {t("common:delete")}
+              </MenuItem>
+            )}
```
notmd, line 318: Now it possible to show deleted message, so I hide the delete button when message is already deleted.

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -143,7 +153,7 @@ export function MessageTableEntry({
                   emoji={{ name: emoji, count }}
                   checked={emojiState.user_emojis.includes(emoji)}
                   userReacted={emojiState.user_emojis.length > 0}
-                  userIsAuthor={message.user_is_author}
+                  userIsAuthor={!!message.user_is_author}
```
AbdBarho, line 156: do we need this double negation?

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -158,6 +168,25 @@ export function MessageTableEntry({
           <LabelMessagePopup message={message} show={labelPopupOpen} onClose={closeLabelPopup} />
           <ReportPopup messageId={message.id} show={reportPopupOpen} onClose={closeReportPopup} />
         </HStack>
+        <Flex position="absolute" gap="2" top="-2.5" right="5">
```
AbdBarho, line 171: can you use `start` instead of `right`? for RTL languages?

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -158,6 +168,25 @@ export function MessageTableEntry({
...
+          {showAuthorBadge && message.user_is_author && (
+            <Tooltip label={t("message_author_explain")} placement="top">
+              <Badge size="sm" colorScheme="green" textTransform="capitalize">
+                {t("message_author")}
+              </Badge>
+            </Tooltip>
+          )}
+          {message.deleted && isAdminOrMod && (
+            <Badge colorScheme="red" textTransform="capitalize">
+              Deleted
```
AbdBarho, line 181: should this also be moved to localization? If not, add a comment to indicate why.

website/src/components/Messages/MessageTree.tsx
```diff
@@ -9,9 +9,10 @@ const messagePaddingTop = 16;
 const avatarSize = 32;
 const avartarMarginTop = 6;
 const maxDepth = 100; // this only used for debug UI in mobile
-const left = avatarSize / 2 - 1;
+const toPx = (val: number) => `${val}px`;
 
-export const MessageTree = ({ tree, messageId }: { tree: MessageWithChildren; messageId?: string }) => {
+// eslint-disable-next-line react/display-name
+export const MessageTree = memo(({ tree, messageId }: { tree: MessageWithChildren; messageId?: string }) => {
```
AbdBarho, line 15: why are we using memo here?

website/src/components/Messages/MessageTree.tsx
```diff
@@ -9,9 +9,10 @@ const messagePaddingTop = 16;
...
   const renderChildren = (children: MessageWithChildren[], depth = 1) => {
```
AbdBarho, line 16: Should this be its own component? otherwise maybe useMemo or similar? because this will be re-created and re-rendered every time.

website/src/components/Messages/MessageTree.tsx
```diff
@@ -53,46 +46,73 @@ export const MessageTree = ({ tree, messageId }: { tree: MessageWithChildren; me
...
+        md: `calc(100% - ${mdTop})`,
+      }}
       position="absolute"
       width="2px"
-      bg="gray.300"
-      left={`${left}px`}
-      top={top}
-      className={className}
+      bg={connectionColor}
+      left={{ base: toPx(-8), md: toPx(avatarSize / 2 - 1) }}
```
AbdBarho, line 97: start instead of right? RTL languages (again)

AbdBarho (commented):
Thanks, looks really cool

website/src/components/Messages/MessageTableEntry.tsx
notmd: Just to make TS happy cause backend mark it as optional (nullable)

website/src/components/Messages/MessageTableEntry.tsx
notmd: It's admin only feature

website/src/components/Messages/MessageTree.tsx
notmd: It only rerenders if tree and messageId changed because we already have memo

website/src/components/Messages/MessageTableEntry.tsx
notmd: I'm tried insetEnd but it didn't work? Could you help me with rtl stuff?

AbdBarho: yeah I guess it should be `inset-block-end` or maybe `inset-inline-end`, if it does not work you can leave like this, we need to fix a lot of RTL stuff anyway.

website/src/components/Messages/MessageTableEntry.tsx
AbdBarho: great, please add comment in the code, so no one tries to extract it out to the locale.

website/src/components/Messages/MessageTableEntry.tsx
notmd: Both of them didn't work. Charka only render right. Charka [docs ](https://chakra-ui.com/docs/styled-system/rtl-support) require to install `stylis` and `stylis-plugin-rtl`. I think we should ignore them for now

AbdBarho (approved):

mm12:
Its live, looks great!

PR #1486, opened 2023-02-11
A fix ukraine translate!

FastBoot-pixellab:
I'm not checking github if you want say something write me on discord U-Boot#2128 or wait.


AbdBarho (approved):
Thanks!

PR #1488, opened 2023-02-11
Adding new Dutch (NL) translations

tijsmaas:
Could you assign the label i18n, thanks!


notmd:
You need to add the language code to next-i18next.config.js too

fozziethebeat (changes requested):
As @notmd noted, you have to update `website/next-i18next.config.js` and add `nl` to the list.

fozziethebeat (changes requested):
I see a lot of merge conflcits.  Did these changes get submitted in a different PR? Can you check if this is still needed?

notmd:
I think this can be closed now. #1514 has cover this already

PR #1499, opened 2023-02-11
RL training

sanagno:
I know @theblackcat102 you dont like everything clamped together but I think it makes sense, just for now :))

1. fix some minor bugs
2. RL training using trlx (need to experiment more here)
3. OA dataset for RL training

We need to think how to use the reward model when training bigger models...


theblackcat102 (approved):
Overall seems fine, just confirm one question: So we are using the prefix part from SFT as prompt right?

theblackcat102:
How big of a model are we talking here which we can't fit it in a 8 A100 80G cluster?

sanagno:
Yes, we are feeding different prefixes and the task is to predict a new reply. 

Regarding the model size, I was referring to the rank model that is just running inference. Will have to check how others implement this. But in essence, we want the rank model to be able to rank generations of different prompts. Not sure if it makes sense to do it offline, or if it is something better done online.

theblackcat102:
Your implementation is very similar to my proof of concept code last year on TRLX. So I concern would be are we doing some fancy cross node offload work? like running the model on other node so we have the all the VRAM available for policy model? 

I am also interested on implementing some similar to sparrow which has multiple reward model scores multiplied into one final score ( page 11, eq 1 ). Not sure if this would cause a huge burden on RL training.

sanagno:
As long as reward models run on different nodes and there is a way to quickly query them, I don't see I reason why that would not work. Especially if it is done offline.

PR #1500, opened 2023-02-11
Minor Bug: Alert icon not appearing on 404 page

jacovzap:
The icon on the left of the button "Report a bug was not showed correctly". The color prop of the icon didn't have the correct format, i fixed it.


fozziethebeat (approved):

fozziethebeat:
I noticed that this and two other PRs you branch off from each other.  In the future I would like to request that each PR always branch off from `main` rather than other PRs.

Issue #1343, opened 2023-02-08
Main part of the task able to move horizontally on mobile website


Dryhb:
While on mobile website, the block with the messages and rating / box to type / multiple assistant answers  to sort and the bottom links (policy, terms of service, discord...) can be moved to the side (left) for no apparent reason even tho the top part (main logo, language selection user icon, dark mode icon, dashboard / message / leaderboard buttons) do not move and aren't moveable. This isn't a zoom problem too. This can be linked to  https://github.com/LAION-AI/Open-Assistant/issues/974 where it mentions the randking card issue (where most of the issue comes from).

jacovzap:
I have solved this problem! I made a pull request on the request #1501

AbdBarho:
#1501

PR #1501, opened 2023-02-12
Horizontally unwanted movement corrected Issue #1343

jacovzap:
The unwanted Horizontally movement was eliminated from the box container of the elements

### DIFF
--- a/website/public/locales/en/dashboard.json
+++ b/website/public/locales/en/dashboard.json
@@ -6,7 +6,7 @@
   "dashboard": "Dashboard",
   "go": "Go",
   "welcome_message": {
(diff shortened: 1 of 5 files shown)


fozziethebeat (approved):

notmd:
@jacovzap could you explain how it's unwanted? because your change makes the page not scrollable now.

jacovzap:
> @jacovzap could you explain how it's unwanted? because your change makes the page not scrollable now.

In mobile sizes the below part of the page was able to move to left, leaving a blank space at the right. You can see the issue #1343. I just deleted one property "overflow y" so It should not broke the scroll since there is no horizontal scrolling on the page

PR #1502, opened 2023-02-12
Set typescript type of TaskInfo.type correctly

johnflux:


johnflux:
Sorry for tiny commit with just one-liner.  I looked around for what else to add in the commit but couldn't see anything obvious.

notmd (approved):
Thanks

PR #1508, opened 2023-02-12
add CC BY-SA 4.0 to terms of use

wassname:
You can see a similar statement is used in [Wikipedia terms of use](https://foundation.wikimedia.org/wiki/Terms_of_Use/en) and it makes sure that the user contributed data is clear to be released under a CC with no disputes.

### DIFF
--- a/website/src/components/ToS.tsx
+++ b/website/src/components/ToS.tsx
@@ -72,6 +72,11 @@ const TermsData = [
         title: "",
         desc: "The use, distribution, storage, forwarding, editing and/or other use of images that violate these terms of use is prohibited.",
       },
+      {
+        number: "3.5",
+        title: "",
+        desc: "When you submit or modify text to which you hold the copyright, you agree to license it under: Creative Commons Attribution-ShareAlike 4.0 (“CC BY-SA 4.0”)"
+      }
     ],
   },
   {


wassname:
This follows a conversation with [Huu Nguyen on discord](https://discord.com/channels/1055935572465700980/1058348535612985394/1074213629764837489)

wannaphong:
Why? Open Assistant is CC BY 4.0. https://projects.laion.ai/Open-Assistant/docs/faq#can-i-download-the-data

wassname:
Sure that's the intention but the user hasn't agreed that thier
contribution is. For that we can include it in the terms of use. It's what
most websites do.

olliestanley:
The problem here is that you have written CC BY-SA in the ToS but this is not what has been discussed previously. CC BY is not the same as CC BY-SA.

wannaphong:
> The problem here is that you have written CC BY-SA in the ToS but this is not what has been discussed previously. CC BY is not the same as CC BY-SA.

Yes, CC BY-SA is not the same as CC BY.

ontocord:
Hey hey! I think it should be CC-BY-4.0 as this is consistent with other LAION datasets. Sorry my bad if I miscommunicated. And thank you for doing this PR!

andreaskoepf:
Short technical question: For >99% of our users we don't have a real name, only an e-mail address or a discord-id and of course a display name for the website (which is automatically generated during e-mail signup). What counts as "Attribution", i.e. where/how will we list the (currently) >22k users by name? I guess most users would not prefer not to have their e-mail address published...

bitplane:
> Short technical question: For >99% of our users we don't have a real name, only an e-mail address or a discord-id and of course a display name for the website (which is automatically generated during e-mail signup). What counts as "Attribution", i.e. where/how will we list the (currently) >22k users by name? I guess most users would prefer not to have their e-mail address published...

Oh this is a good point. I had a look on Wikipedia and they *don't* allow attribution to Wikipedia itself, you link to the article where the page history and list of contributors can be found. So this may not be viable. If you're gonna have attribution then you have to hold people's information perpetually. This will have GDPR implications; you'll have to honour "right to removal" requests as a data controller for as long as you're in control of it. Huggingface, as data processors will only be forced to do this if LAION don't/can't.

A more open license that doesn't require attribution would be preferable IMO. A few options:

* Users giving LAION the data under CC0 and the data being released the same way seems the fairest way to do it.
* Users giving the right for LAION to redistribute under a CC-BY with attribution to them. This seems a bit crappy - users have to attribute LAION, but they don't get the same back.
* The worst would be a generic copyright grant (web2.0 data highwayman approach) where the data could be closed up at any time. Even if data was released under CC0 it feels uneven.

wassname:
Right, I get the point now. Yeah it makes sense to make the output unencumbered by share alike. CC BY-SA is problematic and CC-BY is burdensome.

We could go CC-BY and just include our best effort list off all users on the website and in the dataset. Perhaps even saying that if you don't provide you name you waive right to attribution.

The [makehuman project](http://www.makehumancommunity.org/content/license_explanation.html) uses CC0 for it's output, so perhaps that would be the best.

wannaphong:
I think you see [Common Voice project](https//commonvoice.mozilla.org/). CommonVoice project is use CC-0 and It's a best project for speech dataset.

I don't sure about CC-0 with text corpus but I think if the corpus can be CC-0, It will a best corpus.

Sobsz:
Changing the license now would be problematic, since all past contributors would need to be asked for permission. Granted, it's also not very clear to contributors that the current license is CC BY, so we may be in a pickle regardless.

wassname:
Personally I'm convinced that CC-0 is better, but if we can't get consensus we should merge this CC BY change right now. It's noncontroversial and puts us on a better legal good footing.

Then we can make a new issue to debate CC-0, and change it get consensus.

So let's merge it!

website/src/components/ToS.tsx
```diff
@@ -72,6 +72,11 @@ const TermsData = [
         title: "",
         desc: "The use, distribution, storage, forwarding, editing and/or other use of images that violate these terms of use is prohibited.",
       },
+      {
+        number: "3.5",
+        title: "",
+        desc: "When you submit or modify text to which you hold the copyright, you agree to license it under: Creative Commons Attribution-ShareAlike 4.0 (“CC BY-SA ”)",
```
wannaphong, line 78: Change to CC BY

wannaphong (changes requested):
Chnage to CC BY

wassname:
Sorry I though I had done that. Now it's CC BY 4.0

andreaskoepf:
I think we effectively  ask users to provide information CC-0, maybe something like this should be added to the terms to make this clearer:

"If the user's input constitutes a work protected by copyright, the user grants LAION a simple, temporally, spatially and factually unrestricted right to use the input. In particular, LAION is authorized to use the user's inputs for the development and improvement of large language models."

PR #1512, opened 2023-02-12
Yet another one Russian translation update

0x22almostEvil:


0x22almostEvil:
Previous variant was bad in terms of mobile usage, "Предварительный просмотр" didn't fit in, but adding "\n" or check for these cases won't be that helpful, as it looks ugly in my honest opinion.
And one small fix for better context.

0x22almostEvil:
image

AbdBarho (approved):

PR #1515, opened 2023-02-12
Smoothing out dev setup process

bitplane:
1. Rename devcontainers so they have a nice name in vscode ui
2. Make the root devcontainer a generic container that includes backend and frontend development, plus the stuff for docs (if we're gonna fragment it like this there should be one for the bot too)
3. Add devcontainers and remote code plugins as recommended extensions, so first time the user opens the project they're prompted to install them, then they're prompted to restart using devcontainers once the extensions are installed.
4. Get rid of requirements-dev.txt in oasst-shared by using pyproject.toml, so we can do `pip install ./oasst-shared[dev]` to get dev deps. We should do this for the other packages too (I will if I get time).
5. Bump `pydantic` version so that we're using the same version in the library, backend and the bot. Stops `pip` errors when running multiple things locally.
6. Add `pytest` test discovery so the tests show up in the sidebar. This is a bit of a pig tbh and never works properly.
7. Add python debug and execution launch profiles so we can run and debug files inside or outside the devcontainers
8. Add a couple of trivial tests for the backend, as an example of what needs to be done to start with code coverage
9. A few fixes to the contract test stub script - make it quieter, make it fail when a step fails, make it wait a little longer to start. (IMO we should move the non-unit tests in here to an integration tests dir and have a fixture that spins up the stub using `testcontainers`. Then we can start to build out integration tests, run them alongside functional tests, and run the back-end components in coverage, combine the .cover files and get coverage reports for each type of test)


bitplane:
> x **pre-commit** failed. Please run `pre-commit run --all-files` locally and commit the changes. Find more information in the repository's CONTRIBUTING.md

Something funny going on here. `git commit` and `pre-commit` are acting differently, and fight over the files. Could be a devcontainers vs local issue.

andreaskoepf (approved):
lgtm!

PR #1507, opened 2023-02-12
Fix scroll on Layout

AbdBarho:
You can see on the dev environment that scrolling is broken:

https://web.dev.open-assistant.io/messages

PR #1522, opened 2023-02-12
fix layout issue bug

notmd:
close #1507

### DIFF
--- a/website/src/components/SideMenuLayout.tsx
+++ b/website/src/components/SideMenuLayout.tsx
@@ -11,12 +11,18 @@ export const SideMenuLayout = (props: SideMenuLayoutProps) => {
   const { colorMode } = useColorMode();
 
   return (
-    <Box backgroundColor={colorMode === "light" ? "gray.100" : colors.dark.bg}>
+    <Box backgroundColor={colorMode === "light" ? "gray.100" : colors.dark.bg} className="sm:overflow-hidden">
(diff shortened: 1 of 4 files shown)


website/src/components/SideMenuLayout.tsx
```diff
@@ -13,10 +13,16 @@ export const SideMenuLayout = (props: SideMenuLayoutProps) => {
...
       <Box display="flex" flexDirection={["column", "row"]} h="full" gap={["0", "0", "0", "6"]}>
-        <Box p={["3", "3", "3", "6"]} pr={["3", "3", "3", "0"]}>
+        <Box p={["3", "3", "3", "6"]} pe={["3", "3", "3", "0"]}>
           <SideMenu buttonOptions={props.menuButtonOptions} />
         </Box>
-        <Box className="p-3 lg:p-6 lg:pl-1 w-full">{props.children}</Box>
+        <Box
+          className="p-3 lg:p-6 ltr:lg:pl-0 rtl:lg:pr-0 w-full"
+          overflowX="hidden"
+          overflowY={{ base: "hidden", sm: "auto" }}
```
AbdBarho, line 22: it is only hidden on `xs`? it feels to me that you can remove the `base` item, or maybe remove the overflowY entirely?

notmd: It's hidden in xs to avoid the double scroll bar which is really annoying and auto on sm to allow scrolling

AbdBarho (approved):
I will close my PR

PR #1545, opened 2023-02-13
Chrome bot

heljakka:
A Chrome extension intended for testing the OpenAssistant.
Upon developer-mode installation, it shows a welcome page and installs the extension.
Once installed, the extension will show an input field for a question and a button a the top of every page. 
As the user writes a question and presses the button, the question and a cleaned up version of the page HTML contents is chunked into fixed-side array of parts and sent to the given (OpenAssistant) back-end endpoint for evaluation using JSON payload {"prompt": "Question"} and expecting a JSON response {"summ": "Answer"}. The response Answer is shown on the line below the input field.

Development notes:
* No data is sent from the extension nor from any website unless the user presses the Ask button.
* Safety of the content as such is not analyzed in any way. 
* The suitability of the extension to Chrome webstore has not been evaluated.
* Replace ASK_ENDPOINT with your own endpoint. There is no default working API endpoint as of this writing.
* Size of the chunks to be sent to the endpoint should be determined based on what the endpoint can process.
* The html contents are filtered only using very basic heuristics. This leaves occasional javascript code bits, button labels and similar irrelevant content to the json payload.
* UI layout is trivial and should be made nicer.
* The small OpenAssistant logo is used everywhere, whereas multiple sized logos would be appropriate.


andreaskoepf:
A chrome-plugin as proposed here would need a public API access. Due to hosting-costs it is unfortunately currently and for the foreseeable future not possible to offer this. At some point it might be interesting together with self-hosted LoRA models.

PR #1550, opened 2023-02-13
Update inference installation

### LINKED ISSUE #1549: Update inference documentation
I recently tried to setup the local inference server for testing different NLG models and found the installation documentation had some missing or incomplete information. 

We should update this so the inference server is a standalone service which doesn't require external context or documentation.

jackapbutler:
fixes #1549. 

I've verified that I can setup the inference-server locally using each method with a fresh Python environment. 

* Updates installation option 2. with necessary requirements.
* Pins `fastapi` version so that we can don't have a `RuntimeError` similar to [this](https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/7714).
* Adds missing `postgres` container to installation option 3.
* Adds missing API key to manual installation option 3

### DIFF
--- a/inference/README.md
+++ b/inference/README.md
@@ -92,7 +92,7 @@ Run the inference server:
 ```bash
 cd server
 pip install -r requirements.txt
-uvicorn main:app --reload
+DEBUG_API_KEYS='["0000"]' uvicorn main:app --reload
 ```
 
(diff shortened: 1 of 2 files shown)


yk (approved):
thank you so much!

PR #1552, opened 2023-02-13
website: Fix stories, add session decorator and button stories

rjmacarthy:
This PR fixes all the error preventing stories from loading.

- Added `SessionDecorator` as `useSession` requires a provider
- Added `RouterDecorator` for to support usage of `useRouter` hook.
- Added a story for Button components
- Fixed some linter warnings

Any problems please let me know.

### DIFF
--- a/website/src/components/LanguageSelector/LanguageSelector.tsx
+++ b/website/src/components/LanguageSelector/LanguageSelector.tsx
@@ -37,7 +37,7 @@ const LanguageSelector = () => {
       await router.push(path, path, { locale });
       router.reload();
     },
-    [router]
+    [router, setCookie]
   );
 
   const { language: currentLanguage } = i18n;
--- a/website/.storybook/decorators.js
+++ b/website/.storybook/decorators.js
@@ -1,0 +1,17 @@
+import { RouterContext } from "next/dist/shared/lib/router-context";
+import { SessionProvider } from "next-auth/react";
+import { createMockRouter } from 'src/test-utils/createMockRouter'
+
+export const SessionDecorator = (Story) => (
+  <SessionProvider>
+    <Story />
+  </SessionProvider>
+)
+export const RouterDecorator = (Story) => {
+  const router = createMockRouter()
+  return (
+    <RouterContext.Provider value={router}>
+      <Story />
+    </RouterContext.Provider>
+  )
+}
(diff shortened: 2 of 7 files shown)


website/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -23,7 +23,7 @@ const LanguageSelector = () => {
 
   // Memo the set of locales and their display names.
   const localesAndNames = useMemo(() => {
-    return router.locales.map((locale) => ({
+    return router?.locales?.map((locale) => ({
```
AbdBarho, line 26: why would the `router` be undefined?

website/src/components/LanguageSelector/LanguageSelector.tsx
```diff
@@ -37,13 +37,13 @@ const LanguageSelector = () => {
...
     },
-    [router]
+    [router, setCookie]
   );
 
   const { language: currentLanguage } = i18n;
   return (
     <Select onChange={languageChanged} defaultValue={currentLanguage}>
-      {localesAndNames.map(({ locale, name }) => (
+      {localesAndNames?.map(({ locale, name }) => (
```
AbdBarho, line 46: the previous `useMemo` should return an empty array in case there are no locales.

AbdBarho (commented):
Thanks

website/src/components/LanguageSelector/LanguageSelector.tsx
rjmacarthy: Since we are using this in a story it's easier to add this rather than try to mock the router and locale for storybook to render the Header component.

website/src/components/LanguageSelector/LanguageSelector.tsx
rjmacarthy: Storybook throws an error `Cannot read properties of undefined (reading 'map')`.  Adding this fixes that.

website/src/components/LanguageSelector/LanguageSelector.tsx
AbdBarho: Understandable, but this mixes testing logic with business logic.

we already have a router mock https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/test-utils/createMockRouter.ts feel free to extend / modify / copy as necessary.

you can use `RouterProvider` for providing the mock, similar to the session.

rjmacarthy: Ok, I'll take a look at using the mock router.

AbdBarho (approved):
Thank you!

PR #1555, opened 2023-02-14
[feature] Add script for interactive conversation with trained model and some bug fixes

theblackcat102:
**Bug fixes:**

* Mainly in __init__.py under custom_datasets, the original code didn't account into QA or summarization without validation/test split which cause the eval parameters missing ( assign **train** but should be **dataset** )

* Set a max cap on the numbers of preprocess pairs in translation, otherwise it was taking too long for translation to finish

**Features:**

Add interactive chat model tools

```
python -m tools.model_chat --model_path ../supervised_finetuning/EleutherAI/pythia-12b-deduped-1000 --max_new_tokens 256
```

Calling it via module method would remove the need to append path to sys

### DIFF
--- a/model/model_training/utils.py
+++ b/model/model_training/utils.py
@@ -214,7 +214,7 @@ def get_metrics(conf, tokenizer):
 
 
 def get_model(conf, tokenizer):
-    model = get_specific_model(conf.model_name, conf.cache_dir, conf.quantization, conf.seq2seqmodel)
+    model = get_specific_model(conf.model_name, cache_dir=conf.cache_dir, seq2seqmodel=conf.seq2seqmodel)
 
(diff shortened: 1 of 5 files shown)


model/model_training/utils.py
```diff
@@ -214,7 +214,7 @@ def get_metrics(conf, tokenizer):
 
 
 def get_model(conf, tokenizer):
-    model = get_specific_model(conf.model_name, conf.cache_dir, conf.quantization, conf.seq2seqmodel)
+    model = get_specific_model(conf.model_name, cache_dir=conf.cache_dir, seq2seqmodel=conf.seq2seqmodel)
```
sanagno, line 217: good catch, i missed this

sanagno (approved):
Looks great

PR #1562, opened 2023-02-14
Added feedback text to empty messages cards

### LINKED ISSUE #1547: Enhancing messages section with locales
I've seen that when you access to the messages page, if you recive no messages from the backend the cards remain empty with no feedback at all, so you don't know exactly what's going on.

I have a bit of code ready to create a PR to solve that issue and tell the user that there's no messages in that section. Also, I've added the locales for **_English_, _Spanish_, _Català_, _Euskera_ and _Galego_**.

I'll gladly create the PR 😄 

### Before

image

### After

image

image

ehcalabres:
## Fixed

As stated in #1547, when no messages were recieved in the `Messages` section, no feedback were given to the user (it was only showing an empty card), which could be misleading.

To fix this, I've added a `"No messages"` message that will appear in those cases. I've also added locale support for **English, Spanish, Català, Euskera and Galego.**

In addition, I've fixed an issue with the locales that was causing the `Messages` section to fail with a `Text content does not match server-rendered HTML` error when changing from one lenguage to another.

The view will look now as follows when some of the messages columns receives an empty list:

image

This PR closes #1547.


website/src/components/Messages/MessageConversation.tsx
```diff
@@ -9,16 +10,21 @@ interface MessageConversationProps {
...
     <Stack spacing="4">
-      {messages.map((message, idx) => (
-        <MessageTableEntry
-          enabled={enableLink}
-          message={message}
-          key={message.id + message.frontend_message_id}
-          highlight={highlightLastMessage && idx === messages.length - 1}
-        />
-      ))}
+      {!messages || messages.length === 0 ? (
```
notmd, line 16:
```suggestion
      {messages.length === 0 ? (
```
No need to check messages is fasly, because the type already expressed that

notmd (changes requested):

website/src/components/Messages/MessageConversation.tsx
ehcalabres: Perfect, I'll change that 👍

notmd (approved):
Thank you!

PR #1566, opened 2023-02-14
Translate the "Looks like no tasks were found." page

Fitli:
- Added the code for translation of the page that appears when there are no more tasks of given type available
- Updated en, es, and cs locale files


notmd (approved):

notmd:
Thanks

Issue #1528, opened 2023-02-12
Add new statistics view


andreaskoepf:
We currently share statistics about the dataset on the OA discord server but it should be directly integrated in the web app. 

I have added the `/api/v1/stats/cached` and `/api/v1/stats/cached/{name}` backend endpoints with PR https://github.com/LAION-AI/Open-Assistant/pull/1464 . The cached stats are currently recalculated every 60 minutes.

Available stats are currently:
```
    "human_messages_by_lang"
    "human_messages_by_role"
    "message_trees_by_state"
    "message_trees_states_by_lang"
    "users_accepted_tos"
```

We need now a nice front-end page to display the stats. A simple option would be to add a new "Statistics" entry to the command bar on the left side below the "Leaderboard" entry.

rjmacarthy:
I'm willing to cook something up for this.

olliestanley:
Closing as it looks like this was resolved by #1576

PR #1576, opened 2023-02-14
website: Add stats page and side menu item

rjmacarthy:
This PR uses the `/api/v1/stats/cached` to display a simple grid with the data displayed.  This PR is related to issue #1528.  It's assumed from the task description that this new page should be available for the public.

A summary of changes are as follows.

- Add a new route and page at `/stats`
- Add grid to display stats from `/api/v1/stats/cached` endpoint
- Add translations in `en` for the keys of the stats

Future TODOs:
Improve the styling / re-order the grid items, some may be longer height wise than others.

Any problems please reach out.

image


notmd (approved):
Really simple solution! In the future, we may want to use chart instead. Thank you!

rjmacarthy:
> Really simple solution! In the future, we may want to use chart instead. Thank you!

Thanks @notmd , I did consider a graph library but wasn't sure to decide on one. This code and be adapted easily in the future to support a graph for each component or we can keep it in draft/hidden by feature until we are sure.

PR #1587, opened 2023-02-15
Use eos token for end of generated sequence

theblackcat102:
Original method uses question token (<human>) as end of token (dialogue collator) and now change it to the original eos token (<|endoftoken|> in pythia )

Some other changes include adding [gsm8k coding variant - reasoning-machines/gsm-hard](https://huggingface.co/datasets/reasoning-machines/gsm-hard)

```
...
  datasets:
    - gsm8k_hard
...
```


sanagno:
Why not use the question tag to separate prompt-replies? This should help with longer conversations.

theblackcat102:
Oh I was referring to the end of newly generated answer. Joining different prompt-replies are still using question tag. 

```
<prefix>You are ..</prefix><question>Question 1<answer>Answer 1<question>Question 2<answer>(new generated sequence)<eos>
```

vs

```
<prefix>You are ..</prefix><question>Question 1<answer>Answer 1<question>Question 2<answer>(new generated sequence)<question>
```

I don't think it will have performance impact given more data, but since the model has been using eos token in pretraining, there's no reason why we should change it in SFT.

sanagno:
I agree that there should be no difference in performance. 

But it is more intuitive to expect the model to produce the question tag, I.e. that it is waiting for a new prompt, since it does not know when the end of conversation will be.

theblackcat102:
@sanagno the end of conversation is not for the model to decide though, its up to us to parse the result and add it to part of the history conversation and whether the user want to Ctrl+C. The model only focus in finishing the current tasks.

Just remember something, setting eos_token also won't break trlx? [generate argument in PPO rollout](https://github.com/CarperAI/trlx/blob/4390d3f778c403c061979f663adc3b080949014d/trlx/trainer/accelerate_ppo_trainer.py#L62)

sanagno:
I get it, but the model half of the time that it wants to stop generating learns to predict the question token and the other half learns to predict the eos token, which is a bad design choice. Setting the eos token to terminate dialogues only makes sense if there is a clear end to the dialogue itself. In most cases, we can expect this to not be the case. As I said it shouldn't affect performance, so we can move on!

Regarding trlx, that is a good point. But this problem will exist no matter what, since some of the sequence predictions end with the question tag instead of the eos.

I guess what you want to do is to always predict the eos token at the end of the generation of each reply. To do so, we will need to change the dialogue collator, as we will need to learn one reply at a time instead of learning the whole conversation. I don't think we want to do that though. We can also discuss in private about it.

sanagno (approved):

PR #1592, opened 2023-02-15
Add translation prompts in Ukrainian language

nmeln:
Added translation prompts for Ukrainian (similarly to how other languages were defined).


sanagno (approved):
Looks good!

PR #1600, opened 2023-02-15
Correct Info about team roles

AbdBarho:
Just to be correct


andrewm4894 (approved):

andrewm4894:
Makes sense, I kinda just put that stuff in as placeholders :)

PR #1603, opened 2023-02-15
website: Add missing translation keys for stats page

rjmacarthy:
This PR is self explanatory, there was some missing translations based on the keys returned from the server.


AbdBarho (approved):
Thank you!

PR #1607, opened 2023-02-15
Localize some labels that were missing in the tasks view

GuilleHoardings:
Translation to English and Spanish included

### DIFF
--- a/website/src/components/Tasks/Task/Task.tsx
+++ b/website/src/components/Tasks/Task/Task.tsx
@@ -186,7 +186,7 @@ export const Task = () => {
         show={taskStatus.mode === "DEFAULT_WARN"}
         title={t(getTypeSafei18nKey(`${taskInfo.id}.unchanged_title`)) || t("default.unchanged_title")}
         message={t(getTypeSafei18nKey(`${taskInfo.id}.unchanged_message`)) || t("default.unchanged_message")}
-        continueButtonText={"Continue anyway"}
+        continueButtonText={t(getTypeSafei18nKey(`${taskInfo.id}.continue_anyway`)) || t("default.continue_anyway")}
         onClose={() => taskEvent({ action: "RETURN_EDIT" })}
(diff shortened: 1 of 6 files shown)


website/src/components/Tasks/Task/Task.tsx
```diff
@@ -186,7 +186,7 @@ export const Task = () => {
         show={taskStatus.mode === "DEFAULT_WARN"}
         title={t(getTypeSafei18nKey(`${taskInfo.id}.unchanged_title`)) || t("default.unchanged_title")}
         message={t(getTypeSafei18nKey(`${taskInfo.id}.unchanged_message`)) || t("default.unchanged_message")}
-        continueButtonText={"Continue anyway"}
+        continueButtonText={t(getTypeSafei18nKey(`${taskInfo.id}.continue_anyway`)) || t("default.continue_anyway")}
```
notmd, line 189: Why do you need to fallback to "default.continue_anyway" here?

GuilleHoardings: I copied the schema of the two previous strings, `title` and `message`. If I understand the code correctly, the idea is that there might be new `taksInfo.id` that doesn't have a translation and, then, the default one is used.

notmd: ah make sense

notmd (approved):
Thanks

website/src/components/Tasks/Task/Task.tsx
rjmacarthy: Instead of the `||` syntax we can use the default option instead https://www.i18next.com/translation-function/essentials#passing-a-default-value @GuilleHoardings @notmd

GuilleHoardings: Good idea. I didn't know about that. I'll try it and make a pull request.

PR #1612, opened 2023-02-16
Fix the address for the schemas directory

totuta:
- minor fix: the address for the schemas directory


andreaskoepf:
@totuta @andrewm4894 I think schema docs are now a bit a bit outdated, since we added more properties to the message export. If someone wants to update it, would be really nice .. e.g. see [here ](https://github.com/LAION-AI/Open-Assistant/blob/ad05d3820957cd0b0300cd5bcbf634320516370c/backend/oasst_backend/utils/tree_export.py#L25-L62)

andreaskoepf (approved):

PR #1614, opened 2023-02-16
Read the signin env variables from server props

fozziethebeat:
This is to ensure we always use the runtime variable instead of the build time variable.


andreaskoepf (approved):
cool, thank for the changes!

notmd (approved):

PR #1618, opened 2023-02-16
Fix CodeT data notebooks to run out the box and display a sample output

olliestanley:
These notebooks previously required the "data" and "data/augmentation" directories to be manually created to run. This fixes that problem and introduces a cell to output a sample so that anyone browsing GitHub can see an example of the content of the data.


andreaskoepf (approved):
Thank you!

PR #1544, opened 2023-02-13
Markdown rendering

notmd:
close #1350 #837

- Use [Charka prose](https://chakra-ui.com/community/recipes/prose) to style markdown.
- Use [react-markdown](https://github.com/remarkjs/react-markdown) to render markdown.
- Use [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter) to highlight code.
- Remove tailwind base style in favor of Charka

Update: Ready to review now

<details>
  <summary>Lightmode</summary>
 
image
</details>
<details>
  <summary>Darkmode</summary>
 
image
</details>

PR #1620, opened 2023-02-16
Preview markdown when doing task

### LINKED ISSUE #1657: Weirdly formatted text in prompt

Screenshot_20230217_111553

Clearly this is a bad way to format text but I don't think anyone would intentionally format it like this. Is there perhaps something going wrong with the new Markdown functionality?

Link to the prompt: https://open-assistant.io/messages/4da581c2-5ec0-424d-934b-1d5fc3c6be58

notmd:
Follow up of #1544
Also fix #1657

- Add a preview tab just like Github when doing tasks.
- Display preview markdown when clicking the review button.
- Use [https://github.com/Andarist/react-textarea-autosize](https://github.com/Andarist/react-textarea-autosize) to make the textarea auto-grow when typing, so users don't need to scroll anymore.

image


website/src/components/Tasks/CreateTask.tsx
```diff
@@ -55,16 +57,37 @@ export const CreateTask = ({
...
-              textareaProps={{
-                placeholder: t(getTypeSafei18nKey(`tasks:${taskType.id}.response_placeholder`)),
-                isDisabled,
-                isReadOnly: !isEditable,
-              }}
-            />
+            {!isEditable ? (
+              <Suspense fallback={inputText}>
+                <RenderedMarkdown markdown={inputText}></RenderedMarkdown>
+              </Suspense>
```
AbdBarho, line 63: since these 3 lines are identical to the one below, can we hoist them to a `useMemo`?

website/src/components/Messages/RenderedMarkdown.tsx
```diff
@@ -48,7 +48,6 @@ const sx: SystemStyleObject = {
   },
   p: {
     whiteSpace: "pre-wrap",
-    display: "inline-block",
```
notmd, line 51: fixed #1657. I added inline-block because I want to keep the old behavior that text will be inlined with the avatar on mobile but that caused so many problems so I decide to drop it.

AbdBarho (approved):
Thanks

PR #1632, opened 2023-02-16
Pushed for Hindi locale

prasant-in:
Adding hindi local data for website.


AbdBarho (approved):
Thanks, the rest can be translated later!

Issue #1565, opened 2023-02-14
Trollboard: Users with empty display-names can't be banned by GUI


blankster:
When someone with the needed permissions tries to ban a user on the troll board and changes the role to "banned", this fails with the message: "User Role updated failed".

That means that this function is currently broken and changes need to be made in the database directly, what is suboptimal and should be fixed in the near future to enable moderators to do their job.

This seems to be connected with empty display-name- what should be prevented in general! 

A workaround could be a fallback mechanic back to the username.

markuschue:
I'm new to this but I'd like to give this issue a try!

andreaskoepf:
> I'm new to this but I'd like to give this issue a try!

Thanks a lot, I assigned the issue to you! :-) If you join our OA discord please ping me.

PR #1633, opened 2023-02-16
Display name validation

markuschue:
Adding some extra validation for input fields in the login and edit display name pages so that empty or whitespace names are no longer allowed. This should fix the #1565 issue.
It also changes all instances of the display name to the user's auth ID in the website only in case it doesn't match the validation.
I added a small file with the function that's used to validate these display names in `website/src/lib/display_name_validation.ts`.


website/src/pages/auth/signin.tsx
```diff
@@ -183,7 +195,9 @@ const EmailSignInForm = ({
           variant="outline"
           size="lg"
           placeholder="Email Address"
-          {...register("email")}
+          {...register("email", { required: true, pattern: /[^\s@]+@[^\s@]+\.[^\s@]+/g })}
+          isInvalid={errors.email ? true : false}
+          errorBorderColor="orange.600"
```
notmd, line 200: Can you use the `FormHelperText` some where around here and remove the `throwError` callback? see [https://chakra-ui.com/docs/components/form-control](https://chakra-ui.com/docs/components/form-control)

website/src/pages/auth/signin.tsx
```diff
@@ -226,13 +240,28 @@ interface DebugSigninFormData {
   role: Role;
 }
 
-const DebugSigninForm = ({ providerId, bgColorClass }: { providerId: string; bgColorClass: string }) => {
-  const { register, handleSubmit } = useForm<DebugSigninFormData>({
+const DebugSigninForm = ({
```
notmd, line 243: No need to update the debug form. It's only use in local.

notmd (commented):
Nice work overall. Also, can you add the validation to `/website/src/pages/api/username.ts` as well?

website/src/pages/auth/signin.tsx
```diff
@@ -176,32 +189,40 @@ const EmailSignInForm = ({
...
+            variant="outline"
+            size="lg"
+            placeholder="Email Address"
+            {...register("email", { required: true, pattern: /[^\s@]+@[^\s@]+\.[^\s@]+/g })}
+            isInvalid={errors.email ? true : false}
+            errorBorderColor="orange.600"
+          />
+          <FormErrorMessage>
+            {errors.email?.type === "required" && "Email is required"}
+            {errors.email?.type === "pattern" && "Email is invalid"}
```
notmd, line 206: Can you localize this? eg create signin.json file

website/src/pages/auth/signin.tsx
```diff
@@ -176,32 +189,40 @@ const EmailSignInForm = ({
...
+      <FormControl isInvalid={errors.email ? true : false}>
+        <Stack>
+          <Input
+            type="email"
+            data-cy="email-address"
+            variant="outline"
+            size="lg"
+            placeholder="Email Address"
+            {...register("email", { required: true, pattern: /[^\s@]+@[^\s@]+\.[^\s@]+/g })}
+            isInvalid={errors.email ? true : false}
```
notmd, line 201:
```suggestion
```
No need to add this prop. FormControl already handle this.

website/src/pages/auth/signin.tsx
```diff
@@ -176,32 +189,40 @@ const EmailSignInForm = ({
...
-        <SigninButton
-          data-cy="signin-email-button"
-          leftIcon={<Mail />}
-          mt="4"
-          disabled={!captchaSuccess && enableEmailSigninCaptcha}
-        >
-          Continue with Email
-        </SigninButton>
-      </Stack>
+      <FormControl isInvalid={errors.email ? true : false}>
```
notmd, line 192: Can you move this to wrap the Input and FormErrorMessage component only?

website/src/lib/display_name_validation.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Given a user's display name and its ID, returns a valid display name,
+ * checking if the original display name is invalid (e.g. empty or starts
+ * with whitespace).
+ *
+ * @param {string} displayName The user's display name.
+ * @param {string} id The user's ID.
+ * @returns {string} A valid display name.
+ */
+export const getValidDisplayName = (displayName: string, id: string): string => {
+  return !displayName || !displayName.match(/^\S+/g) ? id : displayName;
```
notmd, line 11: This regex is used in many places. Better to extract it to some variable.

website/src/pages/api/username.ts
```diff
@@ -1,19 +1,21 @@
...
-  const { name } = await prisma.user.update({
+  let { name } = await prisma.user.update({
     where: {
       id: token.sub,
     },
     data: {
       name: username,
     },
   });
+  name = getValidDisplayName(name, token.sub);
```
notmd, line 18: This is incorrect. You need to check if the username is valid or not and then perform an update.

notmd (changes requested):

website/src/pages/auth/signin.tsx
markuschue: Got it, I'm also adding FormControl and FormErrorMessage to the `pages/account/edit.tsx` file, if that's okay.

website/src/pages/api/username.ts
markuschue: Oh that's true, sorry. Then I should do the same for the other files where that function is used, right? (trollboard, leaderboard...)

website/src/pages/auth/signin.tsx
markuschue: Sorry I didn't understand this. What should this json file have, and where should I create it?

website/src/pages/api/username.ts
notmd: > Oh that's true, sorry. Then I should do the same for the other files where that function is used, right? (trollboard, leaderboard...)

No need to other files. They look good now.

website/src/pages/auth/signin.tsx
notmd: > Got it, I'm also adding FormControl and FormErrorMessage to the `pages/account/edit.tsx` file, if that's okay.

Yeah, please add to edit file too

website/src/pages/auth/signin.tsx
notmd: I mean create file `/website/public/locales/en/signin.json` with simirlar following content or something:
```
{
  "email_error_required": "Email is required",
  "email_error_pattern": "Email is invalid"
}
```

If you still not sure how to do that, just ignore it for now.

markuschue: I can create these files but I'm still not sure how localization really works. Plus, If I'm not wrong the signin and edit page need translations all around, right? Maybe I can work on it later, but I think I will ignore it for now.

notmd (approved):
Thank you!

PR #1638, opened 2023-02-16
Refactoring: use default value in translation function

GuilleHoardings:
As suggested [in this comment](https://github.com/LAION-AI/Open-Assistant/pull/1607#discussion_r1108367458).


notmd (approved):
Thanks, I don't the previous code work because `t` always has fallback by default

PR #1646, opened 2023-02-16
Update Ukrainian translation files

nmeln:
- Add `stats.json` translation for Ukrainian. 
- Add missing fields. 
- Update a few of previously missed or slightly incorrect translations.


AbdBarho (approved):
Thanks!

PR #1662, opened 2023-02-17
Sync locale across tabs

### LINKED ISSUE #1365: Wrong language tasks given to users, (Button seems to show the wrong language when labeling the inital prompt)

Screenshot_20230208_164020

Presumably the leftmost button should be saying "Not English" rather than "Not Russian".
I'm assuming this is a simple oversight where instead of the target language the detected language is accidentally being used.

AbdBarho:
Closes #1365


rjmacarthy (approved):
It's a shame that it doesn't work out of the box, but the code LGTM.

Issue #1543, opened 2023-02-13
text a bit hard to read on larger rank reply tasks


andrewm4894:
It's like maybe some of the formatting, newlines etc is maybe not in the pop up textbox which kinda just make this big wall of text a little harder to read. Its like maybe its using a tooltip or something - maybe being able to open it in a model with formatting just like the response boxes themselves might make it a bit easier to more easily and quickly read each reply so that they can be ranked.

It might just be me but seeing all the text like this slowed me down a bit as i was no longer able to just quickly scan/read it as i normally would. 

![image](https://user-images.githubusercontent.com/2178292/218455465-4492fcad-1006-48ce-8449-7ae1e313a792.png)

noobyogi0010:
@andrewm4894 I would suggest expanding the response box itself to display the entire text instead of opening a model because it will be more intuitive and will prevent shifting focus from the entire list of responses.
What do you think?

jwstolk:
On a phone, it is also tricky to scroll down to the 3rd answer, without moving answer 2 to position 1. I have to try and scroll by garbing the very right edge of the page. I don't know if this is possible, but maybe the right 1/3 of the answer texts could just scroll the page instead of swapping the answers, since the "grab" icons are on the left side of the answers anyway.

GuilleHoardings:
I like the proposal of using tabs that was brought up in #1378.

noobyogi0010:
@GuilleHoardings I also like the tabs idea too. Let me try and build something of some sort to see how it looks.

notmd:
I dont think change to tabs will resolve this. It's even more confusing with RTL language. And on mobile when having multiple replies we will have a bunch of really long columns. My idea is rank 1-1 reply instead, split the ranking to multiple task. And the backend will resolve the final result something like this A>B, B>C, D>A -> D>A>B>C. It will make ranking must faster. But will require signification backend change.

Jeremias-Roma:
I want to add a new problem, which happened to me in the same task as the first comment.

I really don't know if it is repeated in other web browsers. It should be added, I am using the development version of Microsoft Edge.

![image](https://user-images.githubusercontent.com/89226758/219504515-7f9f14f7-36f8-4746-993b-05cb966ebafb.png)

AbdBarho:
#1680

PR #1680, opened 2023-02-17
Make only leftside draggable for touch devices

mehdi-zibout:
Touch devices users weren't able to scroll easily #1543 
I implemented the solution proposed by [jwstolk](https://github.com/jwstolk) to make only the leftside draggable when using a touch screen.


AbdBarho:
Thanks! how does this look or work on right to left languages? does it switch positions?

mehdi-zibout:
> Thanks! how does this look or work on right to left languages? does it switch positions?

For now I don't think it does, I'll try to work on it today.

AbdBarho (approved):
Thanks

jwstolk:
Thanks! it works great on my phone.

PR #1691, opened 2023-02-17
Fix the path of mtsamples.csv

totuta:
minor fix of path


andreaskoepf (approved):
Thanks for the fix! :-)

Issue #1655, opened 2023-02-17
Styling & rearranging of information on stats page


andreaskoepf:
With the latest release of the OA website a stats page was added, see https://open-assistant.io/stats
The current page is functional and shows all the numbers but it could benefit from additional styling:

## Here are some ideas:
- show lang tags instead of or additionally to the native language names
- display data in tables (i.e. aligned numbers)
- use a different grouping for the section "Message trees states by language", e.g. by state
- (Charts would also be fantastic.)

The current code for the stats components can be found in [Stats.components.tsx](https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/components/Stats/Stats.components.tsx).

Before the stats page was added we regularly published stats as "text tables" on discord:
![image](https://user-images.githubusercontent.com/9976399/219611297-ab5eb729-752d-4f6d-bd6e-e4efb0caac9c.png)

rjmacarthy:
What do we think of using https://react-chartjs-2.js.org/? I've been messing around a little and have managed to render some pie and doughnut charts from the data, if acceptable I can make a PR for it.

![charts](https://user-images.githubusercontent.com/5537428/219617005-c442504d-75f2-48b5-8051-83a4f8d6866d.png)

GuilleHoardings:
Your proposal looks amazing, @rjmacarthy!

However, I think that showing also numbers could be useful. I have been working in a table for "Message trees states by language". It looks like this:

![image](https://user-images.githubusercontent.com/19565386/219619406-4c94afc0-27a4-43f6-b612-f74a4c6fd009.png)

I think it makes it easier to compare different languages.

Maybe we could mix the two approaches. Do you think is worthy?

rjmacarthy:
@GuilleHoardings yes I think your approach for the states by languages makes sense as a table.  If acceptable you could simply copy/rebase the code from my branch into yours and apply the table for the last stat. 
 Here is the code for the charts: https://github.com/rjmacarthy/Open-Assistant/tree/feature/stats-charts

GuilleHoardings:
I won't be able to work in it until this evening. If you want to go ahead and do the pull request, I can add the table once it is merged.

andreaskoepf:
@rjmacarthy @GuilleHoardings A combined approach of tabular data + charts would be fantastic!

horribleCodes:
While you're at it, sort the different tags per language in a fashion that makes sense chronologically. So maybe something like this:

[...]

PR #1696, opened 2023-02-17
Add charts and tables to the stats page

GuilleHoardings:
Related to #1655, it's a mix of the graphs, added by @rjmacarthy, and the table, added by me. It looks like this:

image

I'm sure that many things could be improved in my code. Suggestions are welcome!


website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,208 @@ type MessageTreeStateStatsProps = {
...
 };
 
+const findLangRow = (rows: string[][], lang: string): number => {
+  for (let i = 0; i < rows.length; i++) {
+    const row_lang = rows[i][0];
+    if (row_lang === lang) return i;
+  }
+
+  return -1;
+};
```
AbdBarho, line 49: maybe use this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,208 @@ type MessageTreeStateStatsProps = {
...
+    "aborted_low_grade",
+    "halted_by_moderator",
+  ];
+
+  const languages = [];
+  for (let i = 0; i < stat.stats.length; i++) {
+    if (!languages.includes(stat.stats[i].lang)) {
+      languages.push(stat.stats[i].lang);
+    }
+  }
```
AbdBarho, line 83: this is trying to get unique languages?
```js
const languages = [... new Set(stat.stats.map(s => s.lang))]
```

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,208 @@ type MessageTreeStateStatsProps = {
...
+    const threadCount = stat.stats[i].count;
+    rows[langRow][stateCol] = threadCount;
+  }
+
+  headers = headers.map((item) => t(getTypeSafei18nKey(item)));
+
+  for (let i = 0; i < rows.length; i++) {
+    const langCode = rows[i][0];
+    rows[i][0] = getLocaleDisplayName(langCode) + " [" + langCode + "]";
+  }
```
AbdBarho, line 114: `forEach`?

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,208 @@ type MessageTreeStateStatsProps = {
...
+  return { headers, rows };
+}
+
 export const MessageTreeStateStats = ({ stat, titleFn }: MessageTreeStateStatsProps) => {
   const { t } = useTranslation("stats");
+  const color = useColorModeValue(colors.light.text, colors.dark.text);
+  const langs = stat.stats.map((item) => item.lang);
+  const uniqueLangs = langs.filter((item, index) => langs.indexOf(item) === index);
+  const [selectedLang, setSelectedLang] = useState("en");
+  const stats = stat.stats.filter((item) => item.lang === selectedLang);
```
AbdBarho, line 124: can you `useMemo` on some of these values?

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,208 @@ type MessageTreeStateStatsProps = {
...
+    responsive: true,
+    scales: {
+      x: {
+        stacked: true,
+      },
+      y: {
+        stacked: true,
+      },
+    },
+  };
```
AbdBarho, line 171: This is constant so it could be moved out.

AbdBarho (commented):
Thank you, looks great.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: I was looking for someway of using set, as I would do in Python. I tried your approach, but I got an error about needing option `--downlevelIteration` for using it with a `Set`. I used `Array.from` instead.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Thank you! `findIndex` was what I was looking for.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Done!

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: I changed the code to use the `Set` approach to find the unique languages and states. In addition, I added useMemo. It's my first encounter with this hook, so I hope I did it ok!

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Sorry, I don't understand what you are proposing. Moving it out of where and to where?

notmd: @GuilleHoardings He mean move the bar_options to top-level scope (out side of MessageTreeStateStatsStacked component)

GuilleHoardings: Thanks, I see. However, as it only used in that function, isn't it better to leave inside? I see it as a non-mutable variable, which is defined where is used, not as a constant for all the program.

website/src/components/SimpleTable/SimpleTable.tsx
```diff
@@ -0,0 +1,38 @@
+import { Table, TableContainer, Tbody, Td, Th, Thead, Tr } from "@chakra-ui/react";
+
+interface TableProps {
+  headers: string[];
+  rows: string[][];
+}
+// Create a SimpleTable component that takes in a list of headers and a list of rows
+export const SimpleTable = ({ headers, rows }: TableProps) => {
```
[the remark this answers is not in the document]
GuilleHoardings: I saw it, but it looked very complicated, with many options that are not required in this case. But I understand that creating a whole new component just for this is not a good idea. I will try to use the Table form chakra directly.

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,191 @@ type MessageTreeStateStatsProps = {
...
+
+  const bar_datasets = [];
+  uniqueStates.forEach((state) => {
+    const data_state = stat.stats.filter((item) => item.state === state);
+
+    const lang_count = {};
```
rjmacarthy, line 160: Please could we use camelCase throughout? For example `barDatasets`

website/src/components/Stats/Stats.components.tsx
rjmacarthy: The main reason I would move this out is so that it is not declared on every render of the component.

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,191 @@ type MessageTreeStateStatsProps = {
...
+    const threadCount = stat.stats[i].count;
+    rows[langRow][stateCol] = threadCount;
+  }
+
+  headers = headers.map((item) => t(getTypeSafei18nKey(item)));
+
+  // Add the language code in brackets after the language name
+  rows.forEach((row) => {
+    const langCode = row[0];
+    row[0] = getLocaleDisplayName(langCode) + " [" + langCode + "]";
```
rjmacarthy, line 98: You could use a template literal here for clarity.

website/src/components/Stats/Stats.components.tsx
```diff
@@ -38,30 +39,191 @@ type MessageTreeStateStatsProps = {
...
+    }[];
+  },
+  t
+) {
+  // The headers come from the message tree State.
+  // These two states are not considered because they make
+  // the table too wide and they are not so interesting:
+  // - "ready_for_scoring",
+  // - "scoring_failed"
+  let headers = [
```
rjmacarthy, line 57: Maybe we could declare this outside of the function and rename to something like `headingsOrder` I assume that is why we declare it? Then you could map over this and declare `headings` as a `const` below.

website/src/components/SimpleTable/SimpleTable.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <TableContainer>
+      <Table>
+        <Thead>
+          <Tr>
+            {headers.map((header) => (
+              <Th key={header}>
+                {header.split(" ").map((word, index) => (
```
rjmacarthy, line 16: Is header not a unique value? We could use css to wrap the text in the heading if its too long.

website/src/components/SimpleTable/SimpleTable.tsx
```diff
@@ -0,0 +1,38 @@
...
+                ))}
+              </Th>
+            ))}
+          </Tr>
+        </Thead>
+        <Tbody>
+          {rows.map((row) => (
+            <Tr key={row[0]}>
+              {row.map((cell, index) => (
+                <Td key={row[0] + index}>{cell}</Td>
```
rjmacarthy, line 30: It's generally not recommended to use indexes for keys.

website/src/components/Stats/Stats.tsx
```diff
@@ -17,7 +17,19 @@ export const Stats = ({ data }: StatsProps) => {
     return null;
   }
 
-  const keys = Object.keys(data.stats_by_name);
+  const keys = Object.keys(data.stats_by_name).filter((key) => key !== "users_accepted_tos");
+
+  // Add two copies of the message_trees_states_by_lang for creating the
+  // stacked bar and the table in addition to the donough chart
+  if (keys.indexOf("message_trees_states_by_lang_table") == -1) {
```
rjmacarthy, line 24: I think it's cleaner to add a separate component for the table rather than adding duplicate entries for the stat map.

rjmacarthy:
Thanks for putting the work in to make the stats page more appealing it's looking great. I left some small comments for your consideration please let me know your thoughts on them.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: That makes sense. I was not aware of that kind of performance problems. Thank you!

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Absolutely! Sorry, lately I work in Python and I make this kind of mistakes.

website/src/components/SimpleTable/SimpleTable.tsx
GuilleHoardings: The text of a header can be many words (for instance, "Human messages by language") and needs to be broken in order to have a not so wide table. I don't know how to mix CSS here, so that's why I went with this approach. I'm going to reimplement the table, so maybe this is not needed.

website/src/components/SimpleTable/SimpleTable.tsx
GuilleHoardings: It was the only way I could think of to solve the error message I got when I didn't use them. Anyway, I'm going to try to get rid of this component, so this will be fixed.

website/src/components/Stats/Stats.tsx
GuilleHoardings: Sorry, but I don't understand exactly what you proposed, maybe because I'm not fluent with React components. Could you elaborate a little more?

GuilleHoardings:
> Thanks for putting the work in to make the stats page more appealing it's looking great. I left some small comments for your consideration please let me know your thoughts on them.

Thank you for your comments: they are very helpful. I will try to address all of them.

website/src/components/Stats/Stats.tsx
rjmacarthy: I mean simply add a new component for the table separately below the map for the graphs.

website/src/components/SimpleTable/SimpleTable.tsx
GuilleHoardings: I've reimplemented the table, but I still have the problem of the long headers. How do I break them with CSS? I see that there's `word-wrap: break-word;` in CSS, but I don't know how to change the style only for the headers of this table. Could you please give me a pointer?

website/src/components/SimpleTable/SimpleTable.tsx
GuilleHoardings: This was removed in the new implementation.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Done.

website/src/components/SimpleTable/SimpleTable.tsx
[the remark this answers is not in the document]
GuilleHoardings: In the end, I used the DataTable component.

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Done.

website/src/components/Stats/Stats.tsx
GuilleHoardings: Sorry, but I don't know if I understand the proposal completely. I tried to think another way of doing it, but everything I can think of would require much more code. My lack of knowledge about React is probably why I don't see a better solution. Maybe we could leave it like this and, later, you could send a pull request with a refactoring. What do you think?

website/src/components/SimpleTable/SimpleTable.tsx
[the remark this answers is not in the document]
GuilleHoardings: Although I was able to make it work, I think that the code I wrote in the `getDataTable()` function looks awful. I have the feeling that it could be done without having to repeat the name of the states in so many places (`interestingStates`, `LangRow` and in the column helpers). Do you see a better approach?

website/src/components/Stats/Stats.tsx
rjmacarthy: That is fine with me, thanks.

website/src/components/SimpleTable/SimpleTable.tsx
rjmacarthy: Maybe try this? https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

GuilleHoardings: Thank you, but the problem is not the CSS property to use, but that I don't know how to apply CSS to the headers of the DataTable component, which is what is used now for the table. Do you know how to do that?

rjmacarthy: If it's a chakra component you can use the `sx` prop.

GuilleHoardings: I've done that: I added the white space property to the DataTable component with the sx prop. It works! Thank you for your help!

AbdBarho (approved):
Unless @rjmacarthy has any comments, it all good from my side.

rjmacarthy (approved):

PR #1709, opened 2023-02-18
Updated task docs & help links

horribleCodes:
I've rewritten all of the task docs to give a bit more details for each job - including "Ranking assistant replies", which didn't have a task doc before. I also updated some help links that previously just pointed to the guidelines - they should point to a more apt document now.


AbdBarho (approved):
Looks good.

andrewm4894 (approved):

Issue #1710, opened 2023-02-18
Add "controversy" as an optional label for items


jerzydziewierz:
Some topics are controversial still beyond being not appropriate, hate speech, sexual content or rude. 

For example, the concept of "earth is flat" is not "inappropriate", not "hate speech" not "sexual", not "low quality", not "humorous", not "rude" and it is difficult to classify it as violent.

Please add me (@jerzydziewierz ) for this issue, as I have already prepared a PR for this.

ignas-g:
"the earth is flat" or "vaccines cause autism" are more fringe opinions than controversial. You should be able to have opinions on a topic, but not have your own facts. We know the earth is flat, and vaccines don't cause autism, so the label should be "fringe", but the "controversy" label can certainly be useful when it does not relate to facts. For example, if a movie is good or bad.
Also, can labels have a degree? Some opinions might be slightly controversial, due to the strength of the emotions and some might be slightly fringe due to the distance from the facts.

andreaskoepf:
The current label was already reduced over time, e.g. `political_content` and `moral_judgement` are defined by not part of the survey (and more options were deleted because they were too similar to the ones which are now in the set). Of course the current label set isn't perfect: Potentially there are MANY labels that could be added. But we also need to balance between UI clutter and relevance. If we discuss new labels that apply to <1% of the messages we should explain why it is important/necessary or how the label could be used (e.g. to filter the dataset or to train a model). Beside selecting spam we offer the option to report messages (which adds the flag emoji) which is technically free-text labeling.

> Also, can labels have a degree?

Yes, technically all labels are stored internally as float value between 0 and 1. Initial prototypes of the web app had sliders but early beta tests showed that this is not practical/necessary. Therefore some labels are now binary-flags - either Yes/No questions or toggle-buttons and others are 5-value likert scores with opposing ends. More options potentially lead to more noise and subjectivity.

ignas-g:
I am not sure what the proposition is, but it probably is quite important for the AI to know what is factual and what is not. And not to avoid topics just because there is a vocal fringe minority that disagrees with the facts. Sometimes the minorities can be quite large, for example, a lot of Republicans in the US don't believe in global warming or evolution.

bitplane:
I really like this idea, I think it's better than most of the other axes.

[...]

PR #1711, opened 2023-02-18
add "controversial" to labels

jerzydziewierz:
regarding the issue: https://github.com/LAION-AI/Open-Assistant/issues/1710

This PR:
* Adds "controversial" to labels
* Adds translations of the label to all languages
* Fixes a minor typo in the Polish translation. 

It was tested in local deployment to operate correctly and not throw any new warnings.


oasst-shared/oasst_shared/schemas/protocol.py
```diff
@@ -405,6 +405,7 @@ def __new__(cls, label: str, widget: LabelWidget, display_text: str = "", help_t
     humor = "humor", LabelWidget.likert, "Humorous content including sarcasm"
     helpfulness = "helpfulness", LabelWidget.likert, "Helpfulness of the message"
     creativity = "creativity", LabelWidget.likert, "Creativity"
+    controversial = "controversial", LabelWidget.likert, "Controversial"
```
andreaskoepf, line 408: Why likert? What would be the two extreme ends of the scale? IMO something is either controversial or not, it could be a flag option.

jerzydziewierz: I will not agree with this. Using my definition of `controversial`, there are levels of controversy. 

As per discussion at https://github.com/LAION-AI/Open-Assistant/issues/1710 , "controversial" does not mean `true` nor `untrue` nor `undecidable`. It only means that it has a high potential to cause anger, bias, disagreement, drama and heated discussion.

olliestanley:
Would it be possible for you to split this PR into separate PRs, for the translation fixes and the label? I think the label addition will need more discussion, whereas the translation fixes we can get merged quickly so that we avoid excessive merge conflicts

olliestanley:
Closing this as the author mentioned in #1710 they no longer wish to implement this. If you would still like to make the translation changes in the PL locale feel free to submit them separately :)

PR #1721, opened 2023-02-18
dont render image in markdown by default

notmd:
image

### DIFF
--- a/website/public/locales/en/message.json
+++ b/website/public/locales/en/message.json
@@ -18,6 +18,7 @@
   "report_placeholder": "Why should this message be reviewed?",
   "report_title": "Report",
   "send_report": "Send",
+  "show_image": "Click to show the image",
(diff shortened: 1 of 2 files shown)


website/public/locales/en/message.json
```diff
@@ -18,6 +18,7 @@
   "report_placeholder": "Why should this message be reviewed?",
   "report_title": "Report",
   "send_report": "Send",
+  "show_image": "Click to show the image",
```
notmd, line 21: I'm not sure about the message, feel free to suggest.

0x22almostEvil: That's fine, but maybe we should add some kind of warning: "Warning! Loading from external sources may cause damage."

olliestanley:
As I understand it this is made redundant by the now merged #1724

PR #1650, opened 2023-02-16
Message Tree Topic Modeling Pipeline

danielpatrickhug:
This is a PR to add topic modeling and k-hop message passing including a much faster sparse implementation and sentence transformer embedding aggregation. 

Using message passing and using the k_hop adj matrix to aggregate the embedding features into cluster features like a GCN seems to result in much better topic clusters.

I also added loading tools for the exported message trees, a new util requirements.txt, and refactored the cosine_similarity in similarity_functions.py to instead compute the cosine similarity kernel. cos_sim and embed_data functions were ported over from one of @kenhktsui filter/cluster notebook(Scalable Agglomerative Clustering.ipynb) 

BERTopic: https://maartengr.github.io/BERTopic/

still a WIP but I tested it locally and it works and wanted to get feedback. There are a couple more cleanup tasks, like typing, doc_strings, moving globals like ADJACENCY_THRESHOLD and MODEL_NAME to config, allowing for more customizability of the topic_model, etc. 

Please let me know if you notice any errors or have any suggestions. :)

PR #1726, opened 2023-02-18
A few more unit tests (trivial stuff)

bitplane:
Just doing this bit by bit. 

I've moved stuff about so it doesn't conflict with #605

Adds a bit more to #84 - it's some more examples and a bit more coverage, but doesn't actually run coverage or interfere with #605


backend/tests/unit/utils/test_similarity_functions.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+from oasst_backend.utils.similarity_functions import cosine_similarity, euclidean_distance
+
+
+def test_cosine_similarity():
```
andreaskoepf, line 5: :'( .. I think https://github.com/LAION-AI/Open-Assistant/pull/1650 wants to remove those functions...

bitplane: It's okay, `copilot` wrote the tests for these ones :)

Should I remove them, or keep and remove the tests afterwards?

danielpatrickhug: I can put them back in!

bitplane: I think that if they're code that isn't used then it's best to remove them. Think of code as "lines spent" rather than "lines produced"

bitplane: Cool I'll update this later on, to cover the new functions :)

bitplane:
Converted to draft for now, since some functions have been removed. I'll update in the next day or two

backend/tests/unit/utils/test_similarity_functions.py
bitplane: done :)

bitplane:
I changed the tests to do a simple test of the new cosine similarity functions, they're not great but still better than nothing.

PR #1730, opened 2023-02-19
Translation of missing terms (Spanish)

Jeremias-Roma:
The missing terms in the Spanish version were translated, taking into account the main English version.


Jeremias-Roma:
FAQ was added as such, to speed up future comparisons. In addition, this term is the most commonly used in our language, although it is not an abbreviation in Spanish.

website/public/locales/es/message.json
```diff
@@ -11,7 +13,7 @@
   "open_new_tab_action": "Abrir en una pestaña nueva",
   "parent": "Padre",
   "reactions": "Reacciones",
-  "recent_messages": "Mensajes recientes",
+  "recent_messages": "Mensajes recientes en {{language}}",
```
ArgiesDario, line 16: Why adding the language to the recent messages translation?

AbdBarho: because thats how it should be:
https://github.com/LAION-AI/Open-Assistant/blob/71984331ffcca5f7386d212e2599ca1fce377c5f/website/public/locales/en/message.json#L16

AbdBarho (approved):
thanks

website/public/locales/es/message.json
andreaskoepf: We have now filtering by language .. also the English string was change: `"recent_messages": "Recent Messages in {{language}}"`

PR #1738, opened 2023-02-19
Added Hindi and Japanese Translation 

shashankx86:


LarsEJonasson:
In common.json line 43 and 44 are duplicate translations of "trollboard", and the first lacks the final quotation mark and comma.

In index.json  line 4 lacks the final comma.

In labelling.json line 19 and 65 and lacks the final comma. Line 67 has quoted English terms, but JSON interprets the double quotation marks as a stop. Using escape sequences (a \ before each ") should work.

Line 5 in tos.json should not have a final comma.

shashankx86:
@LarsEJonasson all fixed

LarsEJonasson:
> @LarsEJonasson all fixed and i also can translate to japanese but there isn't any jp folder should i do it?

You still have some formatting issues with tasks.json. If you have Python you can run 'pre-commit run --all-files' on your local computer and it will correct them. For example the spaces between " and : are missing. You also have three initial spaces at each line, but it should be two.

As for more translations, the more the better.

shashankx86:
@LarsEJonasson Done.

shashankx86:
@LarsEJonasson Added jp translation 

image

but here 

image

website/public/locales/jp/labelling.json
```diff
@@ -0,0 +1,79 @@
+{
+  "accurate": "Точное",
```
AbdBarho, line 2: for the japanese files, if we don't have a translation now, then please use the english one, or remove entirely.

shashankx86:
@LarsEJonasson Done

GuilleHoardings:
Japanese files should be under the `ja` directory, not `jp`.

Thank you for your contribution!

shashankx86:
> Japanese files should be under the ja directory, not jp.

Do i need to fix it? @GuilleHoardings

GuilleHoardings:
I think that it doesn't work if you don't fix it. As far as I know, the framework looks in the directory that matches the locale, and the locale for Japan is `ja`. Have you tested it?

shashankx86:
> I think that it doesn't work if you don't fix it. As far as I know, the framework looks in the directory that matches the locale, and the locale for Japan is `ja`. Have you tested it?

nope

shashankx86:
@GuilleHoardings renamed to ja

GuilleHoardings:
Ok, thank you. I've received another proposal for the Japanase translation, and the submitter merged it with yours. I'll submit as an update when yours is merged.

andreaskoepf (approved):

PR #1747, opened 2023-02-19
Implement separate rate limit for message submission tasks

olliestanley:
This adds separate rate limits for tasks depending on their types. Specifically it implements a tough restriction on the rate of assistant replies and a fairly strict one on prompter replies.

The logic here is that users should not be spamming many replies in a short period of time as this is likely impossible without compromising quality.


backend/oasst_backend/api/deps.py
```diff
@@ -166,6 +166,50 @@ async def __call__(self, request: Request, response: Response, api_key: str = De
...
+        milliseconds: int = 0,
+        seconds: int = 0,
+        minutes: int = 1,
+        hours: int = 0,
+    ) -> None:
+        async def identifier(request: Request) -> str:
+            """Identify a request based on api_key and user.id"""
+            api_key = request.headers.get("X-API-Key") or request.query_params.get("api_key")
+            user = (await request.json()).get("user")
+            return f"{api_key}:{user.get('id')}"
```
andreaskoepf, line 187: To be guaranteed unique the key needs to include api-key, auth-method and user-id, beside of extracting the user from the request json it would probably be easier to get it from the oasst_user_header (`x-oasst-user`) .. which already contains auth-method+username delimited by ':'.

olliestanley: This code is just identical to the code which was in the existing UserRateLimiter here:

https://github.com/LAION-AI/Open-Assistant/blob/fa0a110afb7104ba8e6e419a354a2eeadb3d4d2f/backend/oasst_backend/api/deps.py#L145

Do we need to change that?

olliestanley: @andreaskoepf I have now made a change to this along with the original code it came from, let me know if this is what you had in mind

andreaskoepf: Yes, that looks fantastic. Thanks!

andreaskoepf:
I think for the well behaving web-app it is now right. Maybe for bots and other api users we either need to enforce that x-oasst-user is set (if not return an error) or we indeed still need a fallback to the old method you had + extracting auth-method prefix.

olliestanley:
Added something to handle this case

andreaskoepf (approved):

PR #1750, opened 2023-02-20
Fix graphs in documentation darkmode

someone13574:
Replaces the transparent background of the database graphs white to increase visibility of the arrows while using dark mode.


andrewm4894 (approved):
Thanks!

PR #1763, opened 2023-02-20
fix: added platform support for ghcr.io images to be run on Apple Sil…

melvinebenezer:
Added support for multiple platforms for ghcr.io images. 

When running this image on an x86_64 / amd64 machine, the amd64 variant is pulled and run.

Reference:
https://docs.docker.com/engine/reference/commandline/buildx_build/#platform
https://docs.docker.com/build/building/multi-platform/


yk (approved):
awesome, thank you!

Issue #1023, opened 2023-01-30
Display user avatar in leaderboard


lainisourgod:
I miss avatars next to nicknames. Avatars create a feeling that here on leaderboard are humans not just database records. It makes a process more engaging and it's more interesting to compete with real human than "anonymouses".

<img width="1156" alt="CleanShot 2023-01-31 at 00 30 41@2x" src="https://user-images.githubusercontent.com/24773652/215588262-f646d08b-382e-4758-90b7-fed5f9810df6.png">

fozziethebeat:
I don't think we have all the user avatar information readily available when we're rendering this.  But if anyone wants to take a stab at making that work (even partially) that'd be a nice UI bonus

andreaskoepf:
The frontend DB contains avatar urls in the "image" column of the "Users" table for all discord users. For E-Mail users a dummy avatar image could be shown.

MrlolDev:
Working on it, for email users working with https://dicebear.com/playground

PR #1767, opened 2023-02-20
Display user avatar in leaderboard 

MrlolDev:
Update to #1023
Using dicebear to generate initial avatar or saved avatar_url


website/next.config.js
```diff
@@ -10,6 +10,10 @@ const nextConfig = {
         protocol: "https",
         hostname: "**.discordapp.com",
       },
+      {
+        protocol: "https",
+        hostname: "api.dicebear.com",
```
AbdBarho, line 15: this is not necessary, you can leave the image blank if the user has no image.

AbdBarho (commented):
Thanks

website/src/components/LeaderboardTable/LeaderboardTable.tsx
```diff
@@ -61,14 +62,34 @@ export const LeaderboardTable = ({
...
-          isAdminOrMod ? (
-            <Link as={NextLink} href={`/admin/manage_user/${row.original.user_id}`}>
-              {getValue()}
-            </Link>
-          ) : (
-            getValue()
-          ),
+        cell: ({ getValue, row }) => (
+          <div className="flex flex-row items-center gap-2">
+            {row.original.image ? (
```
AbdBarho, line 67: is it possible to only ternary on the url?

website/src/components/LeaderboardTable/LeaderboardTable.tsx
```diff
@@ -61,14 +62,34 @@ export const LeaderboardTable = ({
...
+              <Image
+                src={`${row.original.image}`}
+                alt={`${getValue()}'s avatar`}
+                width={30}
+                height={30}
+                className="rounded-full"
+              />
+            ) : (
+              <Image
+                src={`https://api.dicebear.com/5.x/initials/png?seed=${getValue()}&radius=50&backgroundType=gradientLinear`}
```
rjmacarthy, line 77: I noticed that you are calling `getValue` more than once, maybe should we cache it?  I realise this prevents the explicit return.  Additionally, might it make sense to create a component for this i.e `<UserAvatar displayName={displayname} />` incase we want to use it again?

notmd (approved):
thanks!

PR #1771, opened 2023-02-20
Allow SortableElements to move out of the parent element

### LINKED ISSUE #1685: Ranking assistant replies bug - a reply is too big
Hello,
I was ranking some assistant reply when I came across a reply so wide it wouln't let me place it at the top if it was in the bottom before (and vice versa). 
image
This only happens if the reply is first or last, If it's in the middle I can move it anywhere.

owen-duncan-snobel:
Fix #1685 the Sortable component for "Ranking assistant replies bug - a reply is too big"


AbdBarho (approved):
Thanks

PR #1772, opened 2023-02-20
Introduce users and Discord auth to inference server

### LINKED ISSUE #1471: Introduce Users to Inference
Currently, the inference system lives under `/inference`. After installing the various dependencies, plus tmux, you can run it using the `full-dev-setup.sh` script (or use the docker compose inference profile).

The inference system currently does not consider any users, meaning it has a global collection of "chats", and anyone can interact with any chat by sending HTTP requests to the `/chat/...` endpoints.

Like the data-collection frontend does, we need to connect these chats to users. To solve this issue, see how the frontend currently hands off users to the backend and either replicate this, or come up with an improved version. Then, store the association of users to chats etc. in the database and modify the DB retrievals such that they consider chat ownership.

olliestanley:
Resolves #1471.

The idea here is to introduce a flexible system which allows arbitrarily many authentication methods to be added in future, starting with Discord. We use a Discord OAuth2 flow to login a user (or register a new user), then issue a JSON Web Token (JWT) which can be decoded to obtain a user ID directly. Therefore any new auth method can follow the same pattern and auth can be handled entirely on the backend.

This additionally includes a separate Redis instance for the inference stack, resolves the pre-commit failure in `main`, and adds missing dependencies for the inference server.

- Ideally we would use `fastapi-discord` which would handle the Discord side for us, however there is no version of it which is compatible with FastAPI versions above `0.84.0`, and we depend on FastAPI `0.88.0`

Some code relating to creating/decoding JWTs is duplicated from existing backend code and may be possible to move to `oasst_shared`. We also really need to clean up `main.py` in inference.

After this initial inclusion of users to the server is merged, we will require two follow-on issues:

- Modify DB schemas and code to associate client actions with DB users
- Lock relevant API calls behind authentication by validating provided JWTs

The new settings will also need configuring in the environments for the server to operate properly.


olliestanley:
I currently cannot test this or fix the pre-commit issues due to problems in the inference app in the main branch. However the code can be reviewed in the meantime.

AbdBarho:
@olliestanley is this in a state where we can start working on delegating auth to from the website to backend? (at least for inference)

olliestanley:
@AbdBarho 

Currently (in theory, as there are unrelated problems with the inference stack preventing me from running and testing this) you could send users through the Discord OAuth2 flow and acquire a JWT. The other endpoints have not yet been modified to actually use this JWT after acquiring it. Once the second of the two bullet points at the end of this PR description is implemented, the website will be able to switch over fully on the inference side.

I would like to replicate the same approach for the data collection backend but it will require a bit of backend-frontend coordination.

yk:
Hey @olliestanley it's likely the database scripts diverged. I suggest you delete your alembic upgrade script, then reset the database, then run the server once to get the db up to the previous state, and then do `alembic revision --autogenerate` (or something like this order)

olliestanley:
> Hey @olliestanley it's likely the database scripts diverged. I suggest you delete your alembic upgrade script, then reset the database, then run the server once to get the db up to the previous state, and then do `alembic revision --autogenerate` (or something like this order)

Hmm ok. The errors I've been getting seemed related to Redis but I will try this and report back

inference/server/main.py
```diff
@@ -1,20 +1,29 @@
...
+from cryptography.hazmat.primitives.kdf.hkdf import HKDF
+from fastapi import Depends, HTTPException, Request, Security
 from fastapi.middleware.cors import CORSMiddleware
+from fastapi.security import APIKeyCookie
+from jose import jwe, jwt
 from loguru import logger
-from oasst_inference_server import client_handler, deps, interface, models, worker_handler
+from oasst_inference_server import client_handler, deps, interface, worker_handler
 from oasst_inference_server.chat_repository import ChatRepository
+from oasst_inference_server.models import DbUser, DbWorker
```
yk, line 20: I think it's generally advised to import modules, not names directly. But it's more of a style choice than a mistake.

inference/server/main.py
```diff
@@ -106,6 +110,70 @@ def maybe_add_debug_api_keys():
         raise
 
 
+@app.get("/auth/login/discord")
+async def login_discord(request: Request):
+    domain = str(request.url).split("/auth/")[0]
```
yk, line 115: the domain part should probably come from the settings, not from the request, otherwise there might be some nasty security exploits possible

inference/server/main.py
```diff
@@ -106,6 +110,70 @@ def maybe_add_debug_api_keys():
...
+    raise HTTPException(status_code=302, headers={"location": auth_url})
+
+
+@app.get("/auth/callback/discord", response_model=protocol.Token)
+async def callback_discord(
+    code: str,
+    request: Request,
+    db: sqlmodel.Session = Depends(deps.create_session),
+):
+    domain = str(request.url).split("/auth/")[0]
```
yk, line 127: same

inference/server/main.py
```diff
@@ -106,6 +110,70 @@ def maybe_add_debug_api_keys():
...
+                "client_id": settings.auth_discord_client_id,
+                "client_secret": settings.auth_discord_client_secret,
+                "grant_type": "authorization_code",
+                "code": code,
+                "redirect_uri": redirect_uri,
+                "scope": "identify",
+            },
+        ) as token_response:
+            token_response_json = await token_response.json()
+            access_token = token_response_json["access_token"]
```
yk, line 144: raw dict access from outside source is generally dangerous :) maybe some error handling around the whole function?

inference/server/main.py
```diff
@@ -106,6 +110,70 @@ def maybe_add_debug_api_keys():
...
+        async with session.get(
+            "https://discord.com/api/users/@me", headers={"Authorization": f"Bearer {access_token}"}
+        ) as user_response:
+            user_response_json = await user_response.json()
+
+    discord_id = user_response_json["id"]
+    discord_username = user_response_json["username"]
+
+    # Get user in our DB linked to the Discord user
+    user: DbUser = get_user_from_provider_id(db, discord_id=discord_id)
```
yk, line 156: if this can return none, would call it something like `maybe_get_user_from_ ...`, but also just style

inference/server/main.py
```diff
@@ -157,22 +225,61 @@ def create_worker(
...
 ):
     """Deletes a worker."""
-    worker = session.get(models.DbWorker, worker_id)
+    worker = session.get(DbWorker, worker_id)
     session.delete(worker)
     session.commit()
     return fastapi.Response(status_code=200)
+
+
+def get_user_from_provider_id(db: sqlmodel.Session, discord_id: Optional[str] = None) -> Optional[DbUser]:
```
yk, line 249: `Optional[X]` can now be written as `X | None`

yk (approved):
looks good, thank you! I mostly left some style comments, nothing to take too seriously :)

inference/server/main.py
olliestanley: Good spot, changed this to a setting

inference/server/main.py
olliestanley: By dangerous do you just mean in terms of the dict maybe not containing the keys we expect? I have implemented quick error handling in that case but let me know if this should be extended further

inference/server/main.py
olliestanley: I have renamed it `query_user_by_provider_id` to match the naming conventions used for this sort of lookup function in the data collection backend repository classes

olliestanley:
@yk I think I have addressed all comments, but will wait for confirmation before merging

yk:
awesome, thank you!

PR #1776, opened 2023-02-20
Instructions to reproduce training

### LINKED ISSUE #1661: OA dataset consistent splits
We need to make sure that different splits of the dataset are used for sft, reward and rl training.

Basically [sft_dataset](https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/custom_datasets/prompt_dialogue.py#L14) and [reward](https://github.com/LAION-AI/Open-Assistant/blob/main/model/reward/instructor/rank_datasets.py#L304) datasets needs to use the same splits.

sanagno:
Still TODOs:
- Need to fix #1661
- @theblackcat102 please provide scripts on how you are preprocessing data for the RM

We also need:
- Simpler RM based on only our dataset
- Some refactoring on RM code
- More experiments with RL...


theblackcat102 (approved):
mostly just naming correction

PR #1781, opened 2023-02-21
Added note about GPT detectors

horribleCodes:
I included a link to a GPT detector so people might be more motivated to check if a reply has been generated by ChatGPT.


olliestanley (changes requested):
I like this change. Could you maybe include a qualification which states the detector is not always reliable and sometimes may be very confident that a result is AI-generated even if it is real, so some discretion is required? Users who are not familiar with the detectors may see this and assume it is foolproof

horribleCodes:
Good point. I considered it, but who knows why I ended up not including it right away.

olliestanley (approved):

PR #1783, opened 2023-02-21
Fix reward model evaluation

### LINKED ISSUE #397: Get model evaluation working on the reward model trainer

maw501:
Closes #397. A few other minor tidy-up pieces too.

### DIFF
--- a/model/reward/instructor/trainer.py
+++ b/model/reward/instructor/trainer.py
@@ -183,7 +183,7 @@ if __name__ == "__main__":
         args=args,
         loss_function=training_conf["loss"],
         train_dataset=train,
-        eval_dataset=evals,
+        eval_dataset=torch.utils.data.ConcatDataset(evals.values()),
         data_collator=collate_fn,
         tokenizer=tokenizer,
         compute_metrics=compute_metrics,
--- a/model/reward/instructor/trainer.py
+++ b/model/reward/instructor/trainer.py
@@ -64,7 +64,7 @@ class RankTrainer(Trainer):
                 loss = self.loss_fct(positive_outputs, negative_outputs)
             else:
                 raise NotImplementedError("Only ranking loss has been implemented for rankgen model")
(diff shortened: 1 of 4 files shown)


model/reward/instructor/trainer.py
```diff
@@ -178,11 +183,11 @@ def prediction_step(
         args=args,
         loss_function=training_conf["loss"],
         train_dataset=train,
-        eval_dataset=evals,
+        eval_dataset=torch.utils.data.ConcatDataset(evals.values()),
```
sanagno, line 186: do we actually want this?

maw501: A `KeyError` is raised if not, since `evals` is a `dict`, not a `Dataset`. I'm not sure why it was created like this in [`get_datasets`](https://github.com/LAION-AI/Open-Assistant/blob/4250425c94b550df8b9b90082d52923525cbcb7e/model/reward/instructor/utils.py#L121) - I could amend it there?

sanagno: Aha I see, it is created as a dict to follow the sft training I guess, but perhaps it needs some special handling to compute the losses. Maybe we can change this in the future.

sanagno (approved):
Looks good, @theblackcat102 if you want to take a look

PR #1784, opened 2023-02-21
Add basic load tests

### LINKED ISSUE #1622: Implement basic load tests for inference-server
Implement basic tests for starting conversations and sending messages to the inference server through the /chat and /chat/{chat_id}/message endpoints

jackapbutler:
* closes #1622

I’ve used locust to write a basic load test which will hit two endpoints sequentially that mimic the text-client in the inference server.
* `/chat` to start a new conversation with chat_id
* `/chat/{chat_id}/message` to send a message to the Assistant

An isolated load test user workflow is summarised by first spawning X users every Y seconds to a maximum amount of N concurrent users

1. A user starts a conversation with the Assistant
2. Then they enter a conversation loop
    a. Send a chat message to the Assistant
    b. Wait until Assistant responds
    c. Wait Z further seconds
Repeat

I've tested this locally on my MacBook and results show the Bot seems to be capable of scaling to 100's of users without any adverse errors (but response times do naturally increase). I'll be able to provide more quantitative results once I run it in a more realistic setup with GPUs.


kappa0x (approved):

inference/worker/requirements.txt
```diff
@@ -4,4 +4,5 @@ rel
 requests
 sseclient-py
 tokenizers
+tokenizers
```
olliestanley, line 7: Looks like a duplicate requirement?

inference/README.md
```diff
@@ -117,3 +122,16 @@ cd text-client
...
+## Distributed Testing
+
+We run distributed load tests using the
+[`locust`](https://github.com/locustio/locust) Python package.
+
+```bash
+pip install locust
+cd tests/locust
+locust
+```
```
olliestanley, line 135: Since 436dd9b478c238fc5eb6d3ca641c2525fb0b056d was merged the endpoints by default require authentication to call. To test without authenticating currently would need to set `use_auth=False` in the inference server settings, so we may want to document that here.

yk: small update, there's now a debug auth setting that lets you log in with arbitrary user names, see the text client for an example

jackapbutler: Thanks, I've copied over the new implementation into the locust test. I think for the time being that's okay and in the future we can think about making utility functions from the text client installable or something?

yk (approved):

olliestanley (approved):

PR #1785, opened 2023-02-21
implement language as state instead of cookie

### LINKED ISSUE #1365: Wrong language tasks given to users, (Button seems to show the wrong language when labeling the inital prompt)

### LINKED ISSUE #1530: Other languages thread becomes english
An thread is in Spanish, but the assistant reply is in english.

### LINKED ISSUE #1692: Unresponisive change of language
I usually work on Spanish. However, today I switched to English and then I was unable to switch back to Spanish.

notmd:
close #1365 close #1662 close #1692 close #1530
- Detect language from URL instead of cookie.

Steps to verify this.
- In tab 1, choose a task in language en.
- Open tab 2, change the language to a different one, and complete 1 task.
- Back to tab 1 to finish the pending task. Verify that the next task and the UI are still in English.

This change is really **critical**, so be careful when deploying to production, better to test it in staging with our community first

### DIFF
--- a/website/.eslintrc.json
+++ b/website/.eslintrc.json
@@ -14,8 +14,7 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ],
-    "@typescript-eslint/no-non-null-assertion": 0
+    ]
   },
   "plugins": ["simple-import-sort", "unused-imports"]
 }
--- a/website/.eslintrc.json
+++ b/website/.eslintrc.json
@@ -1,10 +1,6 @@
 {
   "root": true,
-  "extends": [
-    "eslint:recommended",
-    "plugin:@typescript-eslint/recommended",
-    "next/core-web-vitals"
-  ],
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "next/core-web-vitals"],
   "rules": {
     "unused-imports/no-unused-imports": "warn",
     "simple-import-sort/imports": "warn",
@@ -18,7 +14,8 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "@typescript-eslint/no-non-null-assertion": 0
   },
   "plugins": ["simple-import-sort", "unused-imports"]
 }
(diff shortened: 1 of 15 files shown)


notmd:
Pre-commit failure is not related to my change

website/.eslintrc.json
```diff
@@ -18,7 +14,8 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "@typescript-eslint/no-non-null-assertion": 0
```
AbdBarho, line 18: there is a comment to disable this rule in the corresponding file, is this necessary?

website/src/pages/dashboard.tsx
```diff
@@ -1,34 +1,24 @@
...
   });
 
-  useEffect(() => {
-    // re-fetch tasks if the language has changed
-    if (activeLang !== language) {
-      setLang(language);
-      fetchTasks();
-    }
-  }, [activeLang, setLang, language, fetchTasks]);
-
```
AbdBarho, line 31: I have seen that you updated the `useSWR`, will it auto fetch?

Otherwise the cards on the dashboard will be wrong.

AbdBarho (commented):
Thank you, hopefully we can close this problem once and for all.

website/.eslintrc.json
notmd: I found this rule is really annoying and doesn't have many advantages so I disable it globally, but I can revert if you want.

website/src/pages/dashboard.tsx
notmd: Yes, revalidateOnMount: true which is global config will fetch it by default

website/.eslintrc.json
AbdBarho: maybe we can make it a warning at least.

website/src/pages/dashboard.tsx
AbdBarho: great

AbdBarho (approved):
other than the eslin rule, it looks good!

I cannot merge because of the pre-commit, we are gonna have to wait for it to be fixed on main, or someone with "admin" rights to force merge.

Issue #1759, opened 2023-02-20
Trollboard: Option to Hide Banned Users


blankster:
The moderator team as well as the "spam" volume is getting bigger and bigger. To collaborate nicely together and to prevent spam/unwanted entries and inefficient overhead in collaboration, I would appreciate it if the already banned users could (optionally) be hidden on the troll board. 

This would allow a quick overview of critical users on the trollboard, instead of a view that is polluted by already banned users, and makes early detection of problematic users more difficult because this way all moderators have to go through all users again and again to evaluate whether someone is banned or not.

Early detection is important to avoid wasting the time of constructive community members and moderators.

ArgiesDario:
Hi, just to get a clear picture of this request, I created a "megatroll" user, which ends in the Trollboard cause negative votes on his post

![image](https://user-images.githubusercontent.com/7121037/220150791-700a1304-4502-4ccb-a326-41cea877cbe3.png)

So if this user gets its Role changed to "banned"

![image](https://user-images.githubusercontent.com/7121037/220151054-92f654da-0aba-47b4-8260-e0530c405a0a.png)

We should not longer see him in the Trollboard list

Is that correct?

andreaskoepf:
I added the necessary backend changes here: https://github.com/LAION-AI/Open-Assistant/pull/1765 You can add  `enabled=true` to the trollboards query to only retrieve enabled users, don't include send the query param to return all users and set it to false to only return disabled users. I think in the front-end it would be best to have a checkbox to toggle the filter option.

blankster:
@ArgiesDario I was also thinking about a checkbox or something like that. To have the option to choose this, but not being forced to never see the banned users again. Both views can have their advantages while moderation.

ArgiesDario:
> I added the necessary backend changes here: #1765 You can add `enabled=true` to the trollboards query to only retrieve enabled users, don't include send the query param to return all users and set it to false to only return disabled users. I think in the front-end it would be best to have a checkbox to toggle the filter option.

I was trying to make your endpoint work for a while, but it seems to be ignoring the enabled query param, still getting `"enabled": false` users as response

![image](https://user-images.githubusercontent.com/7121037/220247336-84fd8f66-ba3e-484b-92ab-b89cd7f9c44a.png)

If I hardcode your line 17 in trollboards.py from `enabled: Optional[bool] = None` to `enabled: Optional[bool] = True` it works as expected, so it seems to be some related to the get query params. Am I missing something?

[...]

PR #1786, opened 2023-02-21
Trollboard hide banned

ArgiesDario:
Fixes https://github.com/LAION-AI/Open-Assistant/issues/1759


website/public/locales/en/leaderboard.json
```diff
@@ -24,6 +24,8 @@
   "reply": "Replies",
   "reply_ranked_1": "Replies ranked first",
   "score": "Score",
+  "show_active_user": "Show active users",
+  "show_banned_user": "Show banned users",
```
AbdBarho, line 28: are those only visible for admins  & moderators? if yes, we don't need to bother the community with translating them, we can have them in the page directly.

AbdBarho (commented):
Thanks

website/public/locales/en/leaderboard.json
ArgiesDario: Yes, the trollboard is only visible for mods and admins, but since the trollboard table columns are translated, I added this radio too

notmd: @ArgiesDario Can you revert these translations and inline them? Translation for admin probably will be removed soon.

website/src/pages/admin/trollboard.tsx
```diff
@@ -22,6 +39,16 @@ const Leaderboard = () => {
...
+              <RadioGroup defaultValue="1" onChange={setEnabled.toggle}>
+                <Stack direction="row" spacing={5}>
+                  <Radio value="1" colorScheme="green">
+                    {t("show_active_user")}
+                  </Radio>
+                  <Radio value="2" colorScheme="red">
+                    {t("show_banned_user")}
+                  </Radio>
+                </Stack>
+              </RadioGroup>
```
notmd, line 51: Maybe a checkbox is enough herer?

ArgiesDario: It would be enough, but since the behavior of the trollboard is changing due this filter I wanted to make it crystal clear, now you only see active users or banned users. If I did it with a checkbox, the user might expect to see all users (active + banned, like it works right now), and check to only see active ones. But that's not the way it behaves with this new "enabled" flag

website/public/locales/en/leaderboard.json
ArgiesDario: Removed

notmd (approved):
lgtm

website/src/pages/admin/trollboard.tsx
notmd: This can be replaced by a select with 3 option which is more flexible
- Only active users (enabled = true)
- Only banned users (enabled = false)
- All users (enabled = null)

AbdBarho (approved):
Noice!

PR #1791, opened 2023-02-22
update pre-commit to use black 23.1.0

andreaskoepf:
Today we ran into an interesting pre-commit loop:
- `pre-commit run --files oasst-shared/oasst_shared/schemas/inference.py` said everything OK
- `pre-commit run -a` wanted to re-format the file, but after adding the file pre-commit changed it's mind and wanted the old format

In the hope to fix this I updated to [black 23.1.0](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html). The new black version wants to remove a couple of empty lines and parantheses.


yk (approved):
this is gonna create some sweet merge conflicts :D

PR #1796, opened 2023-02-22
fix language detection in `TrackedTextarea`

notmd:


andreaskoepf (approved):

andreaskoepf:
Thanks a lot for the quick fix! :-)

Issue #1138, opened 2023-02-04
Answers not submitted/ Prompts lost


jkminder:
It has happened to me twice now that an answer/prompt I wrote wasn't correctly submitted (not visible under messages). This happened both times after researching the answer for ~15min (which makes it especially annoying;)). Maybe the website timed out or the connection dropped. But I could still press the Submit button. Afterwards, I was forwarded the "No more tasks for you" page.

k-nearest-neighbor:
Thanks for posting. Is it possible to say when this happened? In the last hour?
It would be super awesome if you could estimate the time using this tool and paste the "output" if gives you that looks like `<t:NNN...>` here

jkminder:
Hmm I would say it happened between <t:1675504800:T> and <t:1675555200:T>. My username is "resoundingcarpet" if that helps.

PR #1802, opened 2023-02-22
handle error when submitting task

### LINKED ISSUE #1723: Show error message and retry when submitting tasks
https://github.com/LAION-AI/Open-Assistant/blob/4ab4821b05ba85a1d85bd6d94773cd0f0e5c452d/website/src/hooks/tasks/useGenericTaskAPI.tsx#L42

we should handle all kinds of errors instead of showing the NONE_AVAILABLE state. Maybe trigger a toast to let the user know or something.

notmd:
close #1723 

I also add loading state for submit and skip task


website/public/locales/en/error.json
```diff
@@ -0,0 +1,16 @@
...
+  "err_1003": "Invalid response type",
+  "err_1004": "Interaction request failed",
+  "err_1005": "Failed to generate task",
+  "err_1006": "No tasks of type {{task_type}} are currently available.",
+  "err_1007": "Task availability query failed",
+  "err_1008": "Message size too long",
+  "err_1009": "User recent messages have duplicates",
+  "err_1010": "Message text is empty",
+  "err_1011": "User cannot reply twice to the same message",
+  "err_1012": "User has too many pending tasks"
```
notmd, line 15: I just copied these messages from the backend. Please suggest more friendly messages

notmd: @andreaskoepf can you help on this?

AbdBarho: we could add what the user could do, for example "Current  has too many pending tasks, please wait for a couple of minutes and try again."

AbdBarho (approved):
LGTM

AbdBarho:
are there any retries here? as per https://github.com/LAION-AI/Open-Assistant/issues/1723

website/public/locales/en/error.json
andreaskoepf: I would show the error message that the backend sends in default case, e.g. I think this is better than "Something went wrong!". Regarding more explanatory error messages: Sure, could be done in a separate PR if someone is interested in analyzing when the error are generated and writing a more explanatory text, but IMO the strings as they are now are fine for v1.

andreaskoepf:
Closes  #1138

andreaskoepf:
> are there any automatic retries here? or do we just send the user back to editing state?

IMO we don't need automatic retry.. if the error is shown to the users and they can manually try to re-submit everything is good.

PR #1805, opened 2023-02-22
Inference: Require user authentication to call chat endpoints

### LINKED ISSUE #1789: Use JWT authentication on inference server API calls
As #1772 is merged and users and authentication exist in the inference server database, we need to enforce this authentication for the endpoints in the inference server which are designed for user consumption. The `decode_user_access_token` function in `main.py` can be used.

Relevant endpoint functions:

- `list_chats`
- `create_chat`
- `get_chat`

https://github.com/LAION-AI/Open-Assistant/blob/main/inference/server/main.py

olliestanley:
Closes #1789.

With this merged `use_auth` setting will have to be disabled for local testing without auth. We may want to have a dummy user for when auth is disabled, instead of just returning `None` for the current user ID.

### DIFF
--- a/inference/server/main.py
+++ b/inference/server/main.py
@@ -1,3 +1,4 @@
+import json
 import time
 from datetime import datetime, timedelta
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


yk (approved):
thank you! maybe post in the "breaking changes" channel on discord before merging

PR #1817, opened 2023-02-23
fix task error message

notmd:


website/src/pages/api/update_task.ts
```diff
@@ -33,7 +33,11 @@ const handler = withoutRole("banned", async (req, res, token) => {
 
   const taskId = (registeredTask.task as Prisma.JsonObject).id as string;
 
-  await oasstApiClient.ackTask(taskId, registeredTask.id);
+  try {
+    await oasstApiClient.ackTask(taskId, registeredTask.id);
+  } catch (e) {
+    console.error(e);
+  }
```
notmd, line 40: This try catch is required. When retrying the task, the backend will deny it because the task is already acked.

website/src/pages/api/update_task.ts
```diff
@@ -49,28 +53,13 @@ const handler = withoutRole("banned", async (req, res, token) => {
...
-      user: {
-        connect: {
-          id: token.sub,
-        },
-      },
-    },
-  });
-
-  // Send the next task in the sequence to the client.
-  res.status(200).json(newRegisteredTask);
```
notmd, line 73: The backend doesn't return a new task here (this looks like a legacy behavior) and the client doesn't consume this response so I removed it

website/src/pages/api/update_task.ts
andreaskoepf: I have made ACK idempotent if it is called with the same frontend_message_id ... https://github.com/LAION-AI/Open-Assistant/commit/3ba5bfe40c6f37b3e2e5e4a28082fb6661383bff

website/src/pages/api/update_task.ts
AbdBarho: it would be really great to know what was the intention behind this?
we are we tracking the registered tasks in the db? did we do (or wanted to do) any this with the information?

AbdBarho (approved):

website/src/pages/api/update_task.ts
notmd: Just remove legacy/unnecessary code. Here is what `newTask` variable contain
```
{
  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "type": "task_done"
}
```
We are storing unwanted records. I'm not sure why we store newTask in our DB. Maybe it is a legacy behavior. And in the client, it doesn't consume the response of this endpoint. I will use `/api/new_task` to receive a new task.

Issue #1659, opened 2023-02-17
Setting the Right Incentives: Dynamic Scoring


blankster:
The score you receive for a task can be a great motivator, but the current approach has led to situations where people only focus on the score and choose tasks that offer the highest points, which can have negative consequences.

In my opinion, the incentive system should be adjusted so that the highest points are given for completing tasks that lead to a "ready_for_export" conversation tree quickly and with good quality. Effort should also be considered, but the primary focus should be on the overall benefit to the project, and this should be rewarded with the most points.

Ideally, the scoring system should be dynamically adapting depending on where the most support is needed.

0x22almostEvil:
Yeah, we need more assistant's replies and labels rather than prompts.

horribleCodes:
Good idea. The stat page helps, but I can only assume a lot of people don't really know in what state the trees are and how they can speed things along.

andreaskoepf:
I have changed the weighting of initial prompts to 0.1: https://github.com/LAION-AI/Open-Assistant/commit/93d65bdbed73e154538453f8843274cbf416dce2 This change will become effective with the next deployment.

andreaskoepf:
> The stat page helps, but I can only assume a lot of people don't really know in what state the trees are and how they can speed things along.

All available tasks beside writing initial prompts bring trees closer to the "ready for export" state. The most valuable contributions (also somewhat reflected by the score) is to write good assistant replies. I write "somewhat" because there are obviously people write very long and well researched replies and others users who don't invest the time. A really fair valuation of assistant replies is hard and would require AGI level intelligence (e.g. grading of essays).

JohannesGaessler:
If we're talking purely about how maximizing score has the wrong incentives I think the bigger problem is that it disincentivizes thorough fact-checking. As far as I can tell you always get the same amount of score for labeling an assistant reply regardless of how much effort you put into determining the quality of the reply. So someone carefully checking every single claim in a reply will get a lower amount of score per hour than someone that just checks whether something _sounds_ correct and thus encourages the assistant to tell convincing lies if it can get away with it. I don't think there's an easy solution for this; I personally don't take the score that serious anyways.

There's also an issue in general where the amount of people that will be able to interact with your messages in a positive way will be much higher if the message is at the beginning of a tree rather than at the end. So this kind of disincentivizes putting in a lot of effort towards the end of a tree.

[...]

PR #1824, opened 2023-02-23
Scoring Adjustment

hzj5790:
To help clear the prompt backlog, as per https://github.com/LAION-AI/Open-Assistant/issues/1659, this cuts the score weighting for creating a prompt in half. It also increases the score weighting for penning a good reply (it getting ranked first) by about 20%.

### DIFF
--- a/backend/oasst_backend/models/user_stats.py
+++ b/backend/oasst_backend/models/user_stats.py
@@ -63,7 +63,7 @@ class UserStats(SQLModel, table=True):
             + int(self.accepted_prompts * 0.05)
             + self.accepted_replies_assistant * 4
             + self.accepted_replies_prompter
-            + int(self.reply_ranked_1 * 9 * 1.2)
-            + int(self.reply_ranked_2 * 3 * 1.2)
-            + int(self.reply_ranked_3 * 1.2)
+            + self.reply_ranked_1 * 11
+            + self.reply_ranked_2 * 3
+            + self.reply_ranked_3
         )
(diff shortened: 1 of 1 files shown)


hzj5790:
Just to confirm, self.rankings_total is the number of ranking tasks completed, right?

olliestanley:
We cut the reward for initial prompts by about 90% a few days ago. I'm not sure we can resolve the problem in this way, I think the remaining issues are mostly just that prompting is what people want to do.

hzj5790:
> We cut the reward for initial prompts by about 90% a few days ago. I'm not sure we can resolve the problem in this way, I think the remaining issues are mostly just that prompting is what people want to do, and actually communicating to them that many prompts are not the most helpful action.
> 
> [93d65bd](https://github.com/LAION-AI/Open-Assistant/commit/93d65bdbed73e154538453f8843274cbf416dce2)

Yeah, I saw that too, but my thinking was that any little nudge in the right direction would help, and I didn't take into account behavioral/communication factors like those you mentioned into account.

hzj5790:
If no one has anything more to add, I'll close the PR.

olliestanley:
> If no one has anything more to add, I'll close the PR.

We can keep this open. It might be useful to reduce the score further - we can wait and see what Andreas thinks

andreaskoepf:
I would suggest to add a new string env-variable with comma separated lang-codes of all languages for which the initial-prompt task should be disabled. A setting like "en,es" would then allow to deactivate the prompt for English and Spanish.

andreaskoepf:
I close this since we have two follow-up PRs:  #1849 & #1889

Issue #1788, opened 2023-02-21
Associate inference actions with DB users on the server


olliestanley:
As #1772 is merged and users exist in the inference server database, we want to link chats and actions stored in the database to individual users. This requires modifying the relevant database schemas to add users and recording these users when new actions are stored in the database.

This would include ensuring users can only request chats they created.

PR #1826, opened 2023-02-23
Inference: Associate chats with user IDs 

olliestanley:
Addresses #1788. I'm not sure if there are actions other than chats we want to associate with users also - @yk will need to confirm.

### DIFF
--- a/inference/server/alembic/versions/2023_02_23_2245-b74c66553643_add_user_id_to_chats_table.py
+++ b/inference/server/alembic/versions/2023_02_23_2245-b74c66553643_add_user_id_to_chats_table.py
@@ -1,0 +1,24 @@
+"""Add user id to chats table
+
+Revision ID: b74c66553643
+Revises: b365a18db6fd
+Create Date: 2023-02-23 22:45:19.946188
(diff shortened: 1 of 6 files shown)


yk (commented):
looks good. maybe as a follow-up thought: it might make sense to split the chat repository and extract all functions that are done from the user-side into a sort of user-chat-repository. then we could initialize that with the current user id, and then every function in there could check against that. it would separate the code that needs to be user-access protected from the code that doesn't (i.e. the code that's used from the worker side). but just a suggestion, maybe something for later

yk (approved):

olliestanley:
> looks good. maybe as a follow-up thought: it might make sense to split the chat repository and extract all functions that are done from the user-side into a sort of user-chat-repository. then we could initialize that with the current user id, and then every function in there could check against that. it would separate the code that needs to be user-access protected from the code that doesn't (i.e. the code that's used from the worker side). but just a suggestion, maybe something for later

I like this idea. I will add it as a point of #1843 as I think the inference server needs a bit of general cleanup too

PR #1829, opened 2023-02-24
Make  the colors of the "Message tree states" match

### LINKED ISSUE #1801: Have the colors of the "Message tree states" match
The new pie charts for the stats look great, but as of right now, both "Message tree states" charts use the same color palette, but assign different colors to the same value. This makes things a little hard to compare - it's especially unfortunate that "Ready to Export" in English is colored gray, which makes it look like an unfavorable state. It looks like the colors are randomly assigned, but can you use the same seed or something like that for all of them?

GuilleHoardings:
Fixes #1801

### DIFF
--- a/website/src/components/Stats/Stats.components.tsx
+++ b/website/src/components/Stats/Stats.components.tsx
@@ -53,6 +53,17 @@ const interestingStates = [
   "halted_by_moderator",
 ];
 
+const STATE_COLORS = {
+  initial_prompt_review: "rgba(54, 162, 235, 0.5)",
+  prompt_lottery_waiting: "rgba(255, 99, 132, 0.5)",
+  growing: "rgba(255, 159, 64, 0.5)",
+  backlog_ranking: "rgba(255, 205, 86, 0.5)",
+  ranking: "rgba(75, 192, 192, 0.5)",
+  ready_for_export: "rgba(153, 102, 255, 0.5)",
+  aborted_low_grade: "rgba(201, 203, 207, 0.5)",
+  halted_by_moderator: "rgba(100, 100, 100, 0.5)",
+};
+
 type LangRow = {
   language: string;
   initial_prompt_review: number;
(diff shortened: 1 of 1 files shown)


website/src/components/Stats/Stats.components.tsx
```diff
@@ -249,10 +267,30 @@ export const MessageTreeStateStatsStacked = ({ stat }: MessageTreeStateStatsProp
...
+
 export const Chart = ({ stat, titleFn, type: Component }: ChartProps) => {
-  const data = Object.keys(stat.stats);
+  let data = Object.keys(stat.stats);
   const { t } = useTranslation("stats");
   const color = useColorModeValue(colors.light.text, colors.dark.text);
+
+  if (stat.name === "message_trees_by_state") {
+    // Order data by interesting state
+    data = data.sort((a, b) => {
```
notmd, line 287: You dont need to reassign the `data` variable. `sort` mutate the original array

website/src/components/Stats/Stats.components.tsx
```diff
@@ -173,6 +184,11 @@ export const MessageTreeStateStats = ({ stat, titleFn }: MessageTreeStateStatsPr
   const uniqueLangs = Array.from(new Set(stat.stats.map((s) => s.lang)));
   const [selectedLang, setSelectedLang] = useState("en");
   const stats = stat.stats.filter((item) => item.lang === selectedLang);
+  const statsOrdered = stats.sort((a, b) => {
```
notmd, line 187: same here, just chain the `sort`

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: I didn't know that, thank you. Fixed!

website/src/components/Stats/Stats.components.tsx
GuilleHoardings: Also fixed.

website/src/components/Stats/Stats.components.tsx
```diff
@@ -53,6 +53,17 @@ const interestingStates = [
   "halted_by_moderator",
 ];
 
+const stateColors = {
```
AbdBarho, line 56: could you rename to something like `STATE_COLORS`? since this is a constant.

GuilleHoardings: Done!

AbdBarho (approved):
Thanks

PR #1830, opened 2023-02-24
rename openassistant/ -> data/

andreaskoepf:
This PR renames the `openassistant/` folder to `data/`. I think the original folder name caused a lot of confusion and this change was long overdue.


olliestanley (approved):
I agree with this change. Would it make sense to also add a `data/` section to the code owners file for this going forwards?

andreaskoepf:
> I agree with this change. Would it make sense to also add a `data/` section to the code owners file for this going forwards?

Yes, I also looked at CODEOWNERS and was a bit shocked to see that no-one was responsible for the old openassistant folder. We clearly have to revamp/reorganize the dataset collection efforts. I would nominate @Vechtomov, @bitplane & @ontocord as initial code-owners for `data/`.

olliestanley:
Looks like the code owners file won't recognise Vechtomov - does he have write access to the repo?

andreaskoepf:
> Looks like the code owners file won't recognise Vechtomov - does he have write access to the repo?

I have just sent him an invitation.

andrewm4894:
we could do with adding a README.md to /data maybe - will see if i can do this

PR #1833, opened 2023-02-24
Fix numbered list alignment

### LINKED ISSUE #1818: The Markdown numbering system does not display correctly when more than one digit is used.
image

Slowlife01:
Fixes #1818

image

### DIFF
--- a/website/src/components/Messages/RenderedMarkdown.tsx
+++ b/website/src/components/Messages/RenderedMarkdown.tsx
@@ -70,6 +70,9 @@ const sx: SystemStyleObject = {
       borderBottomColor: "gray.700",
     },
   },
+  "ol li::marker": {
+    content: "counter(list-item) '.'",
+  },
 };
 
 const plugins = [remarkGfm];


website/src/components/Messages/RenderedMarkdown.tsx
```diff
@@ -70,6 +70,9 @@ const sx: SystemStyleObject = {
       borderBottomColor: "gray.700",
     },
   },
+  "ol li::marker": {
+    content: "counter(list-item) '.'",
+  },
```
AbdBarho, line 75: could you please explain this fix to me?

notmd (approved):
Wow thank a lot, first time I see this magic

website/src/components/Messages/RenderedMarkdown.tsx
notmd: Whoop sorry, I didn't see your comment. Here is more information about this https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters

PR #1845, opened 2023-02-24
Add button to copy code in markdown

### LINKED ISSUE #1814: Allow to copy code in markdown
Add an icon to the top right code block as Github does
```
some text
```

Slowlife01:
Resolves #1814 

image
image

Not sure if using Toast is a bit too much but I think it fits the style

### DIFF
--- a/website/src/components/Messages/RenderedMarkdown.tsx
+++ b/website/src/components/Messages/RenderedMarkdown.tsx
@@ -17,9 +17,9 @@ import { useTranslation } from "next-i18next";
 import { memo, MouseEvent, useMemo, useState } from "react";
 import ReactMarkdown from "react-markdown";
 import type { ReactMarkdownOptions } from "react-markdown/lib/react-markdown";
-import { Prism as SyntaxHighlighter } from "react-syntax-highlighter";
-import { oneDark } from "react-syntax-highlighter/dist/cjs/styles/prism";
 import remarkGfm from "remark-gfm";
+import { RenderedCodeblock } from "src/components/Messages/RenderedCodeblock";
+
 interface RenderedMarkdownProps {
   markdown: string;
 }
@@ -27,6 +27,7 @@ interface RenderedMarkdownProps {
 const sx: SystemStyleObject = {
   overflowX: "auto",
   pre: {
+    width: "100%",
     bg: "transparent",
   },
   code: {
@@ -36,6 +37,12 @@ const sx: SystemStyleObject = {
     bg: "gray.300",
     p: 0.5,
     borderRadius: "2px",
+    _rtl: {
+      marginLeft: "3.2rem",
+    },
+    _ltr: {
+      marginRight: "3.2rem",
+    },
     _dark: {
       bg: "gray.700",
     },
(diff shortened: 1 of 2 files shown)


website/src/components/Messages/RenderedMarkdown.tsx
```diff
@@ -1,3 +1,4 @@
+import { CopyIcon, CheckIcon } from "@chakra-ui/icons";
```
notmd, line 1: Can you update to use https://lucide.dev/ icon? and remove the `@charka-ui/icons` deps

website/src/components/Messages/RenderedMarkdown.tsx
```diff
@@ -83,17 +86,64 @@ const disallowedElements = ["img"];
 const RenderedMarkdown = ({ markdown }: RenderedMarkdownProps) => {
   const { isOpen, onOpen, onClose } = useDisclosure();
   const [link, setLink] = useState<string | undefined>();
+  const [isCopied, setIsCopied] = useState(false);
```
notmd, line 89: This will cause a markdown containing multiple `code` blocks will show a copied state in the same time

website/src/components/Messages/RenderedMarkdown.tsx
```diff
@@ -36,6 +38,7 @@ const sx: SystemStyleObject = {
     bg: "gray.300",
     p: 0.5,
     borderRadius: "2px",
+    marginRight: "40px",
```
notmd, line 41: Also why you need marginRight here?

notmd (changes requested):

website/src/components/Messages/RenderedMarkdown.tsx
Slowlife01: I don't want the button to be hiding some text

website/src/components/Messages/RenderedMarkdown.tsx
Slowlife01: Should I use `clipboard-copy` or `clipboard`?

website/src/components/Messages/RenderedMarkdown.tsx
Slowlife01: Will do in the morning, it's pretty late now 😅

website/src/components/Messages/RenderedMarkdown.tsx
notmd: Maybe show the button when hover only like GitHub does, or maybe add margin to the code element only?

website/src/components/Messages/RenderedMarkdown.tsx
notmd: You can use copy which is like Github does

website/src/components/Messages/RenderedMarkdown.tsx
AbdBarho: also please try `inset-block-end` or `inset-inline-end` instead of `margin-right` for RTL languages.

Slowlife01: >or maybe add margin to the code element only?

Is this not the right place to put it?

image

---

>also please try inset-block-end or inset-inline-end instead of margin-right for RTL languages

image
Would this work instead? The inset-block-end and inset-inline-end don't seem to work for me

image
Also, is this how it's supposed to look like?

Slowlife01:
*Something gone wrong there. Super sorry*

notmd (approved):
Thank you so much

PR #1857, opened 2023-02-25
Inference: Server code clean up

### LINKED ISSUE #1843: Clean up inference server code
The inference server code is packed into a small number of files, and endpoints are currently just placed within `main.py`. Once #1826 is merged I want to clean this up to follow a structure more similar to the data collection backend.

From Yannic: One part of this is to split the chat repository and extract all functions that are done from the user-side into a sort of user-chat-repository. Then we could initialize that with the current user id instead of just initialising a regular chat repository, and every function in there could check against the user ID. It would separate code that needs to be user-access protected from the code that doesn't (i.e. the code used from worker side).

olliestanley:
Closes #1843. Not sure if one or two things in here are a bit overzealous, can modify if so.

This also contains a fix for an error in auth in some situations.

### DIFF
--- a/inference/server/main.py
+++ b/inference/server/main.py
@@ -9,11 +9,14 @@ import sqlmodel
 from fastapi import Depends, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from loguru import logger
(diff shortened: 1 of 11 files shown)


yk (approved):
very awesome, thank you!

Issue #1815, opened 2023-02-23
Allow to copy raw message content


notmd:
Add an item in `MessageActions` component to allow copy raw message content

CheckMC:
I'd like to work on this one

CheckMC:
Do I need to provide translations for each language? Or should I leave it in English for now?

notmd:
> Do I need to provide translations for each language? Or should I leave it in English for now?

English is enough for now

CheckMC:
@notmd, can you mark this as finished?

PR #1859, opened 2023-02-25
Copy message text button

CheckMC:
Adds a simple "Copy message text" button in the dropdown menu for messages. 

Issue: #1815

### DIFF
--- a/website/public/locales/en/message.json
+++ b/website/public/locales/en/message.json
@@ -2,7 +2,7 @@
   "confirm_open_link_body": "Do you want to open this link?",
   "confirm_open_link_header": "Confirm open link",
   "copy_message_id": "Copy message ID",
-  "copy_message_text": "Copy message text",
+  "copy_message_text": "Copy raw message",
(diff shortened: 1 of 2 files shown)


website/public/locales/en/message.json
```diff
@@ -2,6 +2,7 @@
   "confirm_open_link_body": "Do you want to open this link?",
   "confirm_open_link_header": "Confirm open link",
   "copy_message_id": "Copy message ID",
+  "copy_message_text": "Copy message text",
```
notmd, line 5: Maybe something like "Copy raw message" is better here?

notmd (approved):
Thanks, lgtm

website/public/locales/en/message.json
CheckMC: That also works. I used "Copy message text" to keep it more in line with "Copy message ID", but raw text may be more accurate. will update

CheckMC: changed

notmd:
Thank you!

Issue #1768, opened 2023-02-20
Ability to Delete Messages From the List View in User Detail View


blankster:
Currently, it's not possible to delete a message from the list view that shows the available messages of a user:

![image](https://user-images.githubusercontent.com/1443109/220166855-d404fc73-f293-4830-afff-01fc1a586329.png)

A direct action to delete the message would be useful to optimize the moderator's UX and efficiency. It could be placed next to the view action and should need a confirmation to prevent miss-clicks/unwanted deletion of messages.

ekp1k80:
Hello, i could do it. This would be my first issue so i would need to know the container or components path in the repo and the link to do the setup of the project in my enviroment

GuilleHoardings:
You can follow the instructions of the [README in the website directory](https://github.com/LAION-AI/Open-Assistant/blob/main/website/README.md).

blankster:
That's still something that would be very useful for content moderation when you have to delete stuff in bigger amounts. Whoever tackles this, best thanks in advance!

johnflux:
I can do this

johnflux:
https://github.com/LAION-AI/Open-Assistant/pull/1860

PR #1860, opened 2023-02-25
Add ability to Delete Messages From the List View in User Detail View…

johnflux:
… (#1768)

Also fix updating deleting messages from message view


website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -106,5 +119,17 @@ const DateDiff = ({ children }: { children: string | Date | number }) => {
...
+  const { trigger: deleteMessage } = useSWRMutation("/api/admin/user_messages", (path, args: { arg: string }) =>
+    del(`/api/admin/delete_message/${args.arg}`)
+  );
+
+  async function handleDeleteMessage(id: string) {
+    if (confirm("Are you sure you want to delete this message?")) {
+      deleteMessage(id);
+    }
+  }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
notmd, line 131: You should not ignore eslint here. Please add the deps to 2nd arg

website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -106,5 +119,17 @@ const DateDiff = ({ children }: { children: string | Date | number }) => {
 };
 
 export const AdminMessageTable = (props: StrictOmit<DataTableProps<Message>, "columns">) => {
+  const { trigger: deleteMessage } = useSWRMutation("/api/admin/user_messages", (path, args: { arg: string }) =>
```
notmd, line 122: Why you change the swr key here?

website/src/pages/admin/manage_user/[id].tsx
```diff
@@ -143,10 +143,13 @@ const ManageUser = ({ user }: InferGetServerSidePropsType<typeof getServerSidePr
 const UserMessageTable = ({ id }: { id: User["id"] }) => {
   const { pagination, toNextPage, toPreviousPage } = useCursorPagination();
   const { data, error, isLoading } = useSWRImmutable<FetchUserMessagesCursorResponse>(
-    `/api/admin/user_messages?user=${id}&cursor=${encodeURIComponent(pagination.cursor)}&direction=${
-      pagination.direction
-    }`,
-    get,
+    "/api/admin/user_messages",
```
Comment (notmd) [inline website/src/pages/admin/manage_user/[id].tsx:146]: Please revert these changes

website/src/pages/admin/messages/[id].tsx
```diff
@@ -17,7 +17,7 @@ const MessageDetail = () => {
   const { data, isLoading, error } = useSWRImmutable<{
     tree: MessageWithChildren | null;
     message?: Message;
-  }>(`/api/admin/messages/${messageId}/tree`, get);
+  }>("/api/admin/messages", () => get(`/api/admin/messages/${messageId}/tree`));
```
Comment (notmd) [inline website/src/pages/admin/messages/[id].tsx:20]: Revert this too

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -249,7 +250,6 @@ const MessageActions = ({
 
   const handleDelete = async () => {
     await deleteMessage();
-    mutate((key) => typeof key === "string" && key.startsWith("/api/messages"), undefined, { revalidate: true });
```
notmd, line 252: You need to call mutate to revalidate the data here. The key should be something like this `/api/messages/${id}/tree`

notmd (changes requested):
Thanks, look good overall. Can you extract the delete message to a custom hook. Also you may need to check swr docs how to revalidate the data

website/src/components/Messages/AdminMessageTable.tsx
johnflux: Because that's how swr is supposed to work.  We are modifying the key /api/admin/user_messages  (It can affect the child messages too). 
Without this, the UI does not correctly update.

website/src/pages/admin/manage_user/[id].tsx
Comment (johnflux) [reply website/src/pages/admin/manage_user/[id].tsx:146]: I will if you insist, but this is the correct way that it's supposed to work.  This is how swr is supposed to be used.  Can we can discuss on discord please?

website/src/components/Messages/MessageTableEntry.tsx
johnflux: This is not needed at all if we properly pass the key to swr

website/src/components/Messages/AdminMessageTable.tsx
johnflux: Done - I change it to use a proper dialog too.

johnflux:
> Thanks, look good overall. Can you extract the delete message to a custom hook. Also you may need to check swr docs how to revalidate the data

Thanks!
I will do!
 I have tested, and there's no need to revalidate the data because I pass the correct key.

website/src/pages/admin/manage_user/[id].tsx
Comment (johnflux) [reply website/src/pages/admin/manage_user/[id].tsx]: Sorry, here it needs to be reverted

website/src/pages/admin/messages/[id].tsx
Comment (johnflux) [reply website/src/pages/admin/messages/[id].tsx:20]: I think this is correct.  You are modifying /api/admin/messages

website/src/components/Messages/AdminMessageTable.tsx
johnflux: This is now extracted to a custom hook

website/src/pages/admin/manage_user/[id].tsx
Comment (johnflux) [reply website/src/pages/admin/manage_user/[id].tsx]: Done

website/src/pages/admin/messages/[id].tsx
Comment (johnflux) [reply website/src/pages/admin/messages/[id].tsx:20]: Done

website/src/components/Messages/MessageTableEntry.tsx
johnflux: I now invalidate /api/messages/* and /api/admin/messages/* and /api/admin/user_messages

notmd (approved):
Thank you!

Issue #1853, opened 2023-02-24
Sticky Navbar in App


cuevantn:
I'd like to make the navbar of the app sticky, just like in the docs:

Sticky:
![sticky-1](https://user-images.githubusercontent.com/83598208/221305109-decd6610-dc7a-4dba-9ecb-209eb25f2b01.gif)

Not sticky:
![sticky-0](https://user-images.githubusercontent.com/83598208/221305091-985afa06-95e9-40b3-b461-4ff97ef792ef.gif)

I would like to contribute with that.

olliestanley:
This seems like a useful change, I have assigned you to this ticket

notmd:
fixed in #1823

PR #1861, opened 2023-02-25
Solves sticky navbar for the app

cuevantn:
Solves #1853

It is solved in the main branch. The issue in production is that a `Grid` component is wrapping the `SideMenuLayout` unnecessarily.

Result:

sticky-5


notmd:
fixed in https://github.com/LAION-AI/Open-Assistant/pull/1823. Also, production branch is for production, please don't submit pr for it. You need to submit to main.

PR #1864, opened 2023-02-25
Show all my messages

### LINKED ISSUE #1306: Let Me See All of My Contributions
I'd love to be able to see all of the messages I've contributed, not just the most recent ones. It's very interesting to find how people have responded to prompts I've suggested.

Really, I want to see if the young seeds I've planted have grown into healthy trees.

johnflux:
Closes #1306
Lets you page through your own messages.

I also make it easier to show your own deleted messages (there's now an optional include_deleted=true  param).  But I don't turn this on, because we don't properly currently apply the badge correct to such deleted messages (not sure why)

### DIFF
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -351,6 +351,23 @@ export class OasstApiClient {
     return this.get<Message[]>(`/api/v1/messages?${params}`);
   }
 
+  fetch_my_messages_cursor(
+    user: BackendUserCore,
+    {
+      direction,
+      cursor,
+      ...rest
+    }: { include_deleted?: boolean; max_count?: number; cursor?: string; direction: "forward" | "back"; desc?: boolean }
+  ) {
+    return this.get<FetchUserMessagesCursorResponse>(`/api/v1/messages/cursor`, {
+      ...rest,
+      username: user.id,
+      auth_method: user.auth_method,
+      after: direction === "forward" ? cursor : undefined,
+      before: direction === "back" ? cursor : undefined,
+    });
+  }
+
   fetch_recent_messages(lang: string) {
     return this.get<Message[]>(`/api/v1/messages`, { lang });
   }
--- a/website/src/pages/api/messages/user.ts
+++ b/website/src/pages/api/messages/user.ts
@@ -7,20 +7,15 @@ const LIMIT = 10;
 const handler = withoutRole("banned", async (req, res, token) => {
   const user = await getBackendUserCore(token.sub);
   const client = createApiClientFromUser(user);
-  const { cursor, direction, include_deleted } = req.query;
+  const { cursor, direction } = req.query;
 
-  let messages;
-  if (typeof cursor === "string") {
-    messages = await client.fetch_my_messages_cursor(user, {
-      include_deleted: include_deleted === "true",
-      direction: direction as "back",
-      cursor: cursor as string,
-      max_count: LIMIT,
-      desc: true,
-    });
-  } else {
-    messages = await client.fetch_my_messages(user);
-  }
+  const messages = await client.fetch_my_messages_cursor(user, {
+    direction: direction as "back",
+    cursor: cursor as string,
+    max_count: LIMIT,
+    desc: true,
+  });
+
   res.status(200).json(messages);
 });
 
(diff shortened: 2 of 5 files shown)


johnflux:
Looks like this:
image

website/src/lib/oasst_api_client.ts
```diff
@@ -351,6 +351,23 @@ export class OasstApiClient {
...
+  ) {
+    return this.get<FetchUserMessagesCursorResponse>(`/api/v1/messages/cursor`, {
+      ...rest,
+      username: user.id,
+      auth_method: user.auth_method,
+      after: direction === "forward" ? cursor : undefined,
+      before: direction === "back" ? cursor : undefined,
+    });
+  }
+
```
notmd, line 370: There is `fetch_user_messages_cursor`, you dont need to implement this

website/src/pages/api/messages/user.ts
```diff
@@ -2,10 +2,25 @@ import { withoutRole } from "src/lib/auth";
...
+  if (typeof cursor === "string") {
+    messages = await client.fetch_my_messages_cursor(user, {
+      include_deleted: include_deleted === "true",
+      direction: direction as "back",
+      cursor: cursor as string,
+      max_count: LIMIT,
+      desc: true,
+    });
+  } else {
+    messages = await client.fetch_my_messages(user);
```
notmd, line 22: There is no use case for this, can you remove it?

notmd (commented):
Thanks

website/src/pages/api/messages/user.ts
```diff
@@ -2,10 +2,25 @@ import { withoutRole } from "src/lib/auth";
...
 const handler = withoutRole("banned", async (req, res, token) => {
   const user = await getBackendUserCore(token.sub);
   const client = createApiClientFromUser(user);
-  const messages = await client.fetch_my_messages(user);
+  const { cursor, direction, include_deleted } = req.query;
+
+  let messages;
+  if (typeof cursor === "string") {
+    messages = await client.fetch_my_messages_cursor(user, {
+      include_deleted: include_deleted === "true",
```
notmd, line 15: I believe we should not expose this until #1884 has been discussed

website/src/lib/oasst_api_client.ts
johnflux: @notmd   Could you help me understand please:

My understanding is have we have both:  `fetch_my_messages` and `fetch_user_messages` because the first one takes a username and auth_method and gets your own messages, authenticating with your own user, and the second takes the user_id and no auth_method only.

So I was following this pattern by adding a `fetch_my_messages_cursor` that also takes a username and auth_method.

I did actually try using `fetch_user_messages_cursor` but it returned a 500 server error, which I assumed was because I was trying to call it as non-admin.

website/src/pages/api/messages/user.ts
johnflux: Sure.  Do you want me to leave the end point in?

website/src/pages/api/messages/user.ts
johnflux: I can remove sure, but note that this is only in the api, and `fetch_message_tree` also has `include_deleted` in the api.  Was trying to be consistent.  Want me to remove anyway?  Thanks

website/src/lib/oasst_api_client.ts
notmd: ah sorry, those are different endpoints.

notmd (approved):
thanks!

website/src/pages/api/messages/user.ts
```diff
@@ -7,20 +7,15 @@ const LIMIT = 10;
...
-  if (typeof cursor === "string") {
-    messages = await client.fetch_my_messages_cursor(user, {
-      include_deleted: include_deleted === "true",
-      direction: direction as "back",
-      cursor: cursor as string,
-      max_count: LIMIT,
-      desc: true,
-    });
-  } else {
-    messages = await client.fetch_my_messages(user);
```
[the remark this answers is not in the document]
johnflux: Another dev confirmed that it's okay to break, so I withdraw my objection :-)

.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
   "eslint.workingDirectories": ["./website", "./docs"],
   "python.testing.pytestArgs": ["backend/tests", "oasst-shared/tests"],
   "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "editor.tabSize": 2
```
AbdBarho, line 8: if you could remove the changes in this file we can merge this directly

johnflux: Done

PR #1865, opened 2023-02-25
Various fixes in inference

yk:
- moved auth to headers instead of cookies
- some auth fixes
- added debug auth login
- added exception logging
- some db query fixes

### DIFF
--- a/inference/server/oasst_inference_server/user_chat_repository.py
+++ b/inference/server/oasst_inference_server/user_chat_repository.py
@@ -6,7 +6,7 @@ from oasst_inference_server import models
 from oasst_shared.schemas import inference
 
 
-class UserChatRepository:
+class UserChatRepository(pydantic.BaseModel):
     session: sqlmodel.Session
     user_id: str = pydantic.Field(..., min_length=1)
     do_commit: bool = True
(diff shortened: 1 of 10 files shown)


inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -1,33 +1,34 @@
...
 from oasst_inference_server import models
 from oasst_shared.schemas import inference
 
 
-class UserChatRepository:
-    def __init__(self, session: sqlmodel.Session, user_id: str | None, do_commit=True) -> None:
-        self.session = session
-        self.user_id = user_id
-        self.do_commit = do_commit
+class UserChatRepository(pydantic.BaseModel):
```
olliestanley, line 9: What's the rationale for making this a `BaseModel` here? Should we not have this be consistent with `ChatRepository` and either have both or neither as `BaseModel`s?

olliestanley (approved):
This looks good, just left one question. Thanks for setting up the proper debug auth login, I was meaning to raise an issue for that yesterday but didn't get around to it!

inference/server/oasst_inference_server/user_chat_repository.py
yk: the only rationale is to trigger pydantic validation, i.e. make sure `user_id` is really a string, but we could also do that ourselves

PR #1868, opened 2023-02-25
Add script for adding missing entries to translated json locales

d3nzil:
As a part of my updating Czech translation, I created this python script for adding missing entries to translated json locales. Requires Python 3.7+
Tested only on Windows, but doesn't use anything platform specific, so it should work as is on any other OS supported by Python.
Feel free to add it if you find it useful.

One thing I'm not sure is the --sort parameter being optional. This ensures that the target json entries follow the original (presumably English). This would hopefully making manual comparison to English easier. Let me know if I should make this automatic instead of optional parameter.

### DIFF
--- a/scripts/frontend-development/add_missing_json_locale_entries.py
+++ b/scripts/frontend-development/add_missing_json_locale_entries.py
@@ -1,0 +1,78 @@
+"""Tool for adding missing entries to json locales."""
+import argparse
+import json
+from pathlib import Path
+
+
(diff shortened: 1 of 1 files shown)


d3nzil:
To explain things a bit better, while there is already find-missing-locales.py script, it is for finding and showing what is missing or untranslated.
This complements that script by automatically copying the missing entries to selected locale.

In this latest update I removed the --sort option - after finding the appropriate documentation, it seems that the ordering should be always followed. And I also added short hint on how to use this tool to the documentation.

PR #1869, opened 2023-02-25
Distributed Sampler

sanagno:
Make the current sampler work correctly for distributed training

- split the dataset per epoch per device
- fix small error that cased the fractions/sizes to be ignored


model/model_training/trainer_sft.py
```diff
@@ -186,6 +188,12 @@ def argument_parsing(notebook=False, notebook_args=None):
     conf["local_rank"] = args.local_rank
     conf["deepspeed"] = args.deepspeed
 
+    # get the world size in deeepspeed
+    if conf["deepspeed"]:
+        conf["world_size"] = int(os.environ["WORLD_SIZE"])
```
andreaskoepf, line 193: Should it fail if not set or could we use `int(os.getenv("WORLD_SIZE", default="1"))`?

sanagno: Should always work with deepspeed but sure better to be safe

PR #1885, opened 2023-02-25
Add frontend test to CI

AbdBarho:


website/src/test-utils/TestProviders.tsx
```diff
@@ -0,0 +1,23 @@
...
+const mockSession = {};
+// the i18n is initialized in `jest.setup.js`, i18next does not have a clean way
+// of adding a provider context
+export const TestProviders = ({
+  router,
+  session,
+  children,
+}: PropsWithChildren<{ router?: NextRouter; session?: any }> = {}) => {
+  return (
+    <RouterContext.Provider value={router ?? mockRouter}>
```
notmd, line 19: Maybe we can add the `FlagsProvider` to this too?

notmd (approved):
lgtm

website/src/test-utils/TestProviders.tsx
AbdBarho: Yeah, I did not need it for the PR so I left it. I hope that is okay. maybe in an upcoming PR?

We should definitely push harder on testing.

notmd: Yeah, I think it's fine for now. We don't depend on feature flags so much.

PR #1886, opened 2023-02-25
Scroll to message when loaded

### LINKED ISSUE #1870: Allow jumping directly to hightlighted message in /messages/${id}
- Render additional id attribute. Maybe something like this is fine: `message_${id}`
- Update the URL to include the hashbang: `#message_${id}`
https://github.com/LAION-AI/Open-Assistant/blob/0a7cd410edb78773c60fd67d24edd53d8773ea76/website/src/components/Messages/MessageTableEntry.tsx#L112

johnflux:
Closes #1870

### DIFF
--- a/website/src/components/DataTable/DataTable.tsx
+++ b/website/src/components/DataTable/DataTable.tsx
@@ -124,7 +124,7 @@ export const DataTable = <T,>({
           <TableCaption pb={0}>{caption}</TableCaption>
           <Thead>
             {getHeaderGroups().map((headerGroup) => (
-              <Tr key={headerGroup.id} sx={{ "white-space": "normal" }}>
+              <Tr key={headerGroup.id} whiteSpace="normal">
                 {headerGroup.headers.map((header) => (
                   <Th key={header.id}>
                     <Box display="flex" alignItems="center">
--- a/website/src/components/Messages/MessageTableEntry.tsx
+++ b/website/src/components/Messages/MessageTableEntry.tsx
@@ -34,12 +34,13 @@ import {
 import NextLink from "next/link";
 import { useRouter } from "next/router";
 import { useTranslation } from "next-i18next";
-import { lazy, Suspense, useCallback, useEffect, useMemo, useState } from "react";
+import { lazy, Suspense, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { LabelMessagePopup } from "src/components/Messages/LabelPopup";
 import { MessageEmojiButton } from "src/components/Messages/MessageEmojiButton";
 import { ReportPopup } from "src/components/Messages/ReportPopup";
 import { useHasAnyRole } from "src/hooks/auth/useHasAnyRole";
 import { useDeleteMessage } from "src/hooks/message/useDeleteMessage";
+import { useScrollToElementOnMount } from "src/hooks/ui/useScrollToElementOnMount";
 import { post, put } from "src/lib/api";
 import { ROUTES } from "src/lib/routes";
 import { colors } from "src/styles/Theme/colors";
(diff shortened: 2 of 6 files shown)


website/src/components/DataTable/DataTable.tsx
```diff
@@ -124,7 +124,7 @@ export const DataTable = <T,>({
           <TableCaption pb={0}>{caption}</TableCaption>
           <Thead>
             {getHeaderGroups().map((headerGroup) => (
-              <Tr key={headerGroup.id} sx={{ "white-space": "normal" }}>
+              <Tr key={headerGroup.id} sx={{ whiteSpace: "normal" }}>
```
AbdBarho, line 127: is it possible to define the whitespace on the element directly? not `sx`?

AbdBarho (dismissed):
Thanks

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -124,6 +126,7 @@ export function MessageTableEntry({ message, enabled, highlight, showAuthorBadge
         outline={highlight ? "2px solid black" : undefined}
         outlineColor={highlightColor}
         cursor={enabled ? "pointer" : undefined}
+        ref={highlight ? focusElementRef : undefined}
         overflowX="auto"
```
AbdBarho, line 130: have you tried with `autoFocus` on the target element? does it solve the problem?

johnflux: I didn't try.  Worth trying.  I'll try to test it out.

website/src/components/DataTable/DataTable.tsx
johnflux: Not sure what you mean, sorry.  I was just fixing the obvious mistake here.

website/src/components/Messages/MessageTableEntry.tsx
johnflux: I investigated, and the answer seems to be no.  React itself doesn't support autoFocus on anything except for input

website/src/components/DataTable/DataTable.tsx
AbdBarho: I mean `<Tr key={headerGroup.id} whiteSpace="normal">`

website/src/hooks/ui/useUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ *  This is a custom hook that returns a ref that can be used to focus an element.
+ *
+ *   Use like:
+ *
+ *
+ *   const focusElementRef = useFocusElement();
+ *   return <div ref={focusElementRef} />
+ */
+export const useFocusElement = <T extends HTMLElement>() => {
```
notmd, line 12: Maybe this should be `useScrollToElementOnMount`? Also can name the file according to the hook's name?

website/src/components/Messages/MessageTableEntry.tsx
```diff
@@ -108,6 +109,7 @@ export function MessageTableEntry({ message, enabled, highlight, showAuthorBadge
   const { t } = useTranslation(["message"]);
 
   const router = useRouter();
+  const focusElementRef = useFocusElement<HTMLDivElement>();
```
notmd, line 112: We only want to scroll on /messages/${id} page. Can you add another prop `scrollToHightlighted` or something similar? It should be false by default.

notmd: Maybe better to forward the ref and pass it from MessageTree component

website/src/components/DataTable/DataTable.tsx
johnflux: done

website/src/hooks/ui/useUtils.ts
johnflux: done

website/src/components/Messages/MessageTableEntry.tsx
johnflux: done

johnflux: > can you add another prop scrollToHightlighted or something similar?

I added, but not actually currently needed because we don't pass the 'highlighted' prop except in /messages/[id] and /admin/messages/[id] pages

notmd (approved):
Thanks

PR #1887, opened 2023-02-25
Update Japanese translation

GuilleHoardings:
This files were sent to me by Discord user ミャー


AbdBarho (approved):
Thanks to you and ミャー

PR #1889, opened 2023-02-26
Add TreeManagerConfiguration.max_prompt_lottery_waiting parameter 

andreaskoepf:
- add lang column to message_tree_state, create multi-column index on `state`, `lang`
- stop generating initial prompt tasks for languages with more than `max_prompt_lottery_waiting ` prompts in `prompt_lottery_waiting` state
- add `MAX_PROMPT_LOTTERY_WAITING` workflow variable


LarsEJonasson:
Is 250 not a rather low value for max_prompt_lottery_waiting? It will not only block prompt generation for en and es, but also for de, fr, ru,ca,zh, pt-BR and th.

olliestanley (approved):

andreaskoepf:
> Is 250 not a rather low value for max_prompt_lottery_waiting? It will not only block prompt generation for en and es, but also for de, fr, ru,ca,zh, pt-BR and th.

The value 250 is a default value that is only used when no corresponding environment variable is set.

The question is: Do we need more prompts currently and why? Or asked differently: When and why do we want to disable initial prompts for some languages?

In the beginning we had always a maximum of N trees (e.g. 100) that included initial prompts. Then after N trees were generated initial prompts only appears very sparely after trees completed - people didn’t like it and demanded to be always able to submit initial prompts. We changes this… now we are at the point where the pendulum swung back and many people demand to stop initial prompts… and the question is 250 or 10k?

PR #1891, opened 2023-02-26
Use utf-8 in find missing locale script

d3nzil:
Windows don't use utf-8 locale, so running this script fails. Setting the encoding in open fixes that.


andreaskoepf (approved):
ok, thanks!

Issue #1881, opened 2023-02-25
Show all user chats


AbdBarho:
Show all the chat for the currently logged in user.

This will require the following (maybe some of it is already implemented)
* An endpoint in the inference server to get all user chats
* A page or sub page on the website to show the list of chats

Clicking on a item from the list should open the chat.

Questions:
* our chats are just Ids, what do we want to show to the user? maybe the first message?

olliestanley:
Since we added auth to the inference server this endpoint will list all chats for the currently authenticated user:

https://github.com/LAION-AI/Open-Assistant/blob/main/inference/server/main.py#L186

Issue #1880, opened 2023-02-25
Thumbs up / down for inference results


AbdBarho:
As follow up to #1477, add thumbs up / down buttons to the inference messages in the Website.

Unfortunately, our current dev setup for inference is a bit complicated, we might need to work on that first.

PR #1893, opened 2023-02-26
Add debug auth between Website & Inference

AbdBarho:
Refs #1880 #1881 

This PR does not add any new features to the UI, but enables debug authorization when talking to the inference api.

### DIFF
--- a/website/src/lib/oasst_inference_client.ts
+++ b/website/src/lib/oasst_inference_client.ts
@@ -1,0 +1,61 @@
+import axios, { AxiosRequestConfig } from "axios";
+import Cookies from "cookies";
+import type { NextApiRequest, NextApiResponse } from "next";
+import { JWT } from "next-auth/jwt";
+import { InferenceCreateChatResponse, InferenceDebugTokenResponse } from "src/types/Chat";
+
+// TODO: this class could be structured better
+export class OasstInferenceClient {
+  private readonly cookies: Cookies;
+  private inferenceToken: string;
+  private readonly userTokenSub: string;
+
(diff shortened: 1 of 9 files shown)


website/src/lib/oasst_inference_client.ts
```diff
@@ -0,0 +1,61 @@
...
+import { JWT } from "next-auth/jwt";
+import { InferenceCreateChatResponse, InferenceDebugTokenResponse } from "src/types/Chat";
+
+// TODO: this class could be structured better
+export class OasstInferenceClient {
+  private readonly cookies: Cookies;
+  private inferenceToken: string;
+  private readonly userTokenSub: string;
+
+  constructor(req: NextApiRequest, res: NextApiResponse, token: JWT) {
```
notmd, line 13: Depending on req and res in this class is really ugly IMO. We can extract to a custom factory or something. But I'm fine with it for now as it's just a prototype.

AbdBarho: Totally agree, I did not find a cleaner solution:

I wanted to save the user's auth token in the session of the user, I tried to inject into the JWT but it seems that next-auth only allows this in very restrictive location (i.e. only when creating a session).

The other option was cookies, the reason we depend on `req` and `res` is to read and write the cookie, which is handled by the `cookies` library, and it does not expose a cleaner API other than giving the req / res directly in the constructor.

We can definitely pass the `Cookies` instance as a constructor parameter, this will only partially solve the problem.

Another options we might consider: saving the token in memory since it is temporary anyway, or saving it in the database with the remaining user-related stuff (which will incur a db call overhead per-request)

notmd (approved):
lgtm

PR #1894, opened 2023-02-26
Store worker compliance check results, add initial scoring algorithm

### LINKED ISSUE #1892: Store inference worker compliance check results and implement scoring
Inference workers go through compliance checks issued by the inference server. As part of this ticket we want to:

- Store the results of compliance checks in the inference server database
- Implement a simple scoring algorithm to evaluate the overall compliance of workers

olliestanley:
Closes #1892,

This also refactors `models.py` into several model definition files by category.


inference/server/alembic/versions/2023_02_26_1328-a2865908a537_add_worker_compliance_checks_to_db.py
```diff
@@ -0,0 +1,52 @@
...
+        sa.ForeignKeyConstraint(
+            ["worker_id"],
+            ["worker.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
+        op.f("ix_worker_compliance_check_worker_id"), "worker_compliance_check", ["worker_id"], unique=False
+    )
+    op.alter_column("chat", "user_id", existing_type=sa.VARCHAR(), nullable=False)
```
olliestanley, line 39: I am not sure why Alembic generated this as I don't think this PR should have made any change affecting this. If anyone can confirm what might have caused this, we might want to remove it?

yk: if you look at the downgrade, then it looks like the type is kept exactly the same. I don't think it's detrimental to just leave it in

yk (commented):
thank you for this! my only real change request is that if such a compliance test fails, it's not necessarily the fault of the worker being tested, it could also be the fault of the worker of which the request is taken from. as we don't know which one made the mistake, both should get their scores dropped, and we'll rely on the law of large numbers to even out accidental misattributions

olliestanley:
> thank you for this! my only real change request is that if such a compliance test fails, it's not necessarily the fault of the worker being tested, it could also be the fault of the worker of which the request is taken from. as we don't know which one made the mistake, both should get their scores dropped, and we'll rely on the law of large numbers to even out accidental misattributions

Ok, I have modified the compliance check table to also store the ID of the "comparison" worker. This can then be used for the score. Honestly this score calculation could do with some improvement as well

yk (approved):
thank you, looks good. yea we can change the score computation later

PR #1895, opened 2023-02-26
Update HumanEval and MBPP notebooks

olliestanley:
This updates HumanEval and MBPP instruction dataset generation notebooks to correct the output format and push their results to HuggingFace Hub, while also moving them to the new `data` directory.

It also changes the name of the `grade-school-math-instructions` directory to `grade_school_math_instructions` for consistency with other directories in `data/datasets`.


Vechtomov (approved):
Later we can combine all these datasets in one and add SOURCE column. Also we can increase diversity of prompts.

olliestanley:
> Later we can combine all these datasets in one and add SOURCE column. Also we can increase diversity of prompts.

I will at least combine the two code gen datasets together and test gen datasets together for now, I can do it quickly. Prompt diversity may be a bit of a challenge given the nature of the instructions but I agree we can probably introduce some at least.

olliestanley:
Now updated to just produce 2 datasets (code gen and test gen) rather than 4.

andrewm4894 (approved):

PR #1904, opened 2023-02-27
Update README.md

beansofhell:
Hey I have generated something nice... It says 
we would like to have it itergrate with apis I think it sounds much better and nicer! please approve of my commit ! :)} 

https://sharegpt.com/c/seRjDvz

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,13 +118,9 @@ We can then take the resulting model and continue with completion sampling step
 
 ## The Vision
 
-We are not going to stop at replicating ChatGPT. We want to build the assistant
-of the future, able to not only write email and cover letters, but do meaningful
-work, use APIs, dynamically research information, and much more, with the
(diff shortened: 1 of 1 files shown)


olliestanley:
I think we should be careful here. We don't want to overpromise. Are there plans to achieve these goals?

andrewm4894:
New text seems a little marketing speak too and buzzy imo. Will defer to the project leads though but sort of does not seem to fit the character of the project as being open and transparent and minimizing or resisting the urge to get too wild with over promising and more abstract or flowery language. 

Totally just a personal opinion here though.

bitplane:
> New text seems a little marketing speak too and buzzy imo.

I agree. When it comes to writing docs I tend to think "it's not done when there's nothing left to add, it's when there's nothing left to remove"

Issue #1873, opened 2023-02-25
Copy dataset of factoid q/a pairs with difficulty ratings from Wikipedia articles


sedthh:
Copy the "Manually-generated factoid question/answer pairs with difficulty ratings from Wikipedia articles. Dataset includes articles, questions, and answers." dataset to HF. 

> Please cite this paper if you write any papers involving the use of the data above:
> [Question Generation as a Competitive Undergraduate Course Project] (http://www.cs.cmu.edu/~nasmith/papers/smith+heilman+hwa.nsf08.pdf)
[Noah A. Smith](http://www.cs.cmu.edu/~nasmith), [Michael Heilman](http://www.cs.cmu.edu/~mheilman), and [Rebecca Hwa](http://www.cs.pitt.edu/~hwa/)
In Proceedings of the NSF Workshop on the Question Generation Shared Task and Evaluation Challenge, Arlington, VA, September 2008.

Link: http://www.cs.cmu.edu/~ark/QA-data/

PR #1905, opened 2023-02-27
added CMU Wiki QA parser

sedthh:
- added notebook that loads the CMU Wiki QA dataset then cleans / filters it and saves it in
Assistant's prompt format
- added README.md based on HF data card
- added requirements.txt for notebook collab use
- updated __init__.py to include HF dataset

Solves https://github.com/LAION-AI/Open-Assistant/issues/1873


andreaskoepf:
@sedthh very nice, thanks a lot! Could you fix the minor formatting issues / pre-commit?

Vechtomov (changes requested):
If we approve this dataset can you add code for downloading and unpacking data?

sedthh:
There is a "quality" key in the metadata, when it's False, the answers are usually just yes or no asnwers, when it's True, then the answers are usually lengthier explanations. I included both because I couldn't get a clear answer on what to do in such cases and wanted to finally push this.

The data was downloaded manually from the link in the markup, the folder can be set at the beginning with the global variables.

Vechtomov (approved):

Issue #1874, opened 2023-02-25
Copy the Ubuntu Dialogue Corpus to HF


sedthh:
Copy the Ubuntu dialogue corpus to HF
https://www.kaggle.com/datasets/rtatman/ubuntu-dialogue-corpus

See if it can be further cleaned (some answers are low quality)

PR #1906, opened 2023-02-27
added Ubuntu QA dataset

sedthh:
- added notebook that loads the ubuntu dialogue dataset then filters Q&A-like chat messages
(about <1% of all data)
- added README.md based on HF datacard
- added requirements.txt to run notebook in collab
- updated __init__.py with HF dataset

solves https://github.com/LAION-AI/Open-Assistant/issues/1874


sedthh:
note: there are examples of the q&a chat messages at the end of the notebook, they vary in quality (some of them are not sure of their answers, some are just not answers at all, but they are mostly ok)

how it works:
- go through all chat messages that are 3,4,5 long
- check only question-like messages
- check answers, make sure they are long enough and aren't just telling the person it's "off-topic"
- if there is something like a "thank you" afterwards from the poster, then it's considered an answer and added to the dataset

sedthh:
Please check the example messages if it's ok to merge.

Vechtomov (changes requested):
Need to fix dataset preview on HF.
image

PR #1907, opened 2023-02-27
refactor inference message score

notmd:
- Use a column to store the message's score instead a table.
- Remove vote table


inference/server/alembic/versions/2023_02_28_0310-d18ddd431d71_added_score_to_message.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('vote')
+    op.alter_column('chat', 'user_id',
+               existing_type=sa.VARCHAR(),
+               nullable=False)
```
notmd, line 25: I'm not sure why alembic generates this. alembic seems so magic to me...

inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -108,17 +108,12 @@ def initiate_assistant_message(self, parent_id: str, work_parameters: inference.
         self.session.refresh(message)
         return message
 
-    def add_vote(self, message_id: str, score: int) -> models.DbVote:
-        logger.info(f"Adding vote to {message_id=}: {score=}")
+    def update_score(self, message_id: str, score: int):
```
olliestanley, line 111: If we are returning message can we add the type hint here as `models.DbMessage`

olliestanley (approved):
Looks good, just left a comment to add one type hint

inference/server/oasst_inference_server/user_chat_repository.py
notmd: Done!

inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -108,17 +108,15 @@ def initiate_assistant_message(self, parent_id: str, work_parameters: inference.
...
+        if score < -1 and score > 1:
+            raise fastapi.HTTPException(status_code=400, detail="Invalid score")
```
notmd, line 113: Is this the correct way to validate input in fastapi? Please correct me if I'm wrong

olliestanley: I think technically we could use [pydantic's built-in constraints](https://docs.pydantic.dev/usage/types/#arguments-to-conint) to constrain the `score` field of `VoteRequest` but this use case is really simple so I don't think this is a problem. Although I think you need to change `and` to `or`

notmd: ah thanks, it was a quick code.

inference/server/main.py
```diff
@@ -190,7 +190,7 @@ async def list_chats(
     logger.info("Listing all chats.")
     chats = ucr.get_chats()
     chats_list = [chat.to_list_read() for chat in chats]
-    return chats.ListChatsResponse(chats=chats_list)
+    return chat_schema.ListChatsResponse(chats=chats_list)
```
notmd, line 193: also fix error when trying get list of chat `/chat`

notmd:
E2E test failure is not related to my changes

olliestanley:
@notmd Sorry, the merge of #1894 has caused a merge conflict for your PR and also means your alembic script should be updated to revise the latest head. If you are not around or are busy with the web side I would be happy to make these changes for you, just let me know!

yk (approved):
Thank you!

PR #1913, opened 2023-02-28
Chat vote UI

notmd:
depend on #1907 
- abstract UI in `MessageTableEntry` into new `BaseMessageTableEntry` component
- refactor change in #1886

image


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -50,32 +61,156 @@ export const ChatConversation = ({ chatId }: ChatConversationProps) => {
...
+        await trigger({ chat_id: chatId, message_id: messageId, score: newScore });
+      } catch {
+        // TODO maybe we should trigger a toast or something
+        // revert on any error
+        setMessages((messages) =>
+          messages.map((message) => {
+            return message.id !== messageId ? message : { ...message, score: oldScore };
+          })
+        );
+      }
```
AbdBarho, line 97: I do understand why this is here, however, we don't handle any errors for almost all of our API requests.

I don't think this is necessary.

website/src/components/Messages/BaseMessageEntry.tsx
```diff
@@ -0,0 +1,60 @@
+import { Avatar, AvatarProps, Box, BoxProps, HStack, useBreakpointValue, useColorModeValue } from "@chakra-ui/react";
+import { forwardRef, lazy, PropsWithChildren, Suspense, useMemo } from "react";
+
+const RenderedMarkdown = lazy(() => import("./RenderedMarkdown"));
+
+export type BaseMessageEntryProps = PropsWithChildren<{
+  content: string;
+  avatarProps: Pick<AvatarProps, "name" | "src">;
+}> &
+  BoxProps;
```
AbdBarho, line 10: maybe this could be simplified to 
```tsx
BoxProps & {
  content: string;
  avatarProps: Pick<AvatarProps, "name" | "src">;
}
```

AbdBarho (commented):
Ok just for me to understand:

In this PR we also try to share as much code as possible between the "normal" messages and chat messages so we extract the shared part out to its  own component?

website/src/components/Chat/ChatConversation.tsx
notmd: But we are calculating the score base on the current score. It might be buggy without handling errors properly

website/src/components/Messages/BaseMessageEntry.tsx
notmd: done

AbdBarho (approved):

andreaskoepf (approved):
ok! :-)

PR #1915, opened 2023-02-28
View chat list

### LINKED ISSUE #1881: Show all user chats

AbdBarho:
Close #1881

Barebones chat list UI.

The chats we have only contain the ID, we need to expand them to show more stuff so that the users can recongnise which chat it is. Maybe showing the last message and the date of the message would be helpful.

I tried as much as possible to avoid potential conflicts with #1913. 

Known issues: Currently we don't fetch all messages for a given chat ID & therefore we cannot continue a previous chat (because we need the id of the last message), this could be fixed easily by adding a fetch call to `ChatConversion` or whichever equivalent component.

image

image


website/src/pages/chat/[id].tsx
```diff
@@ -0,0 +1,48 @@
...
+
+      <Flags authorizedFlags={["chat"]}>
+        <Flex direction="column" gap="2">
+          <Link href="/chat">
+            <Button leftIcon={<List />} size="lg">
+              {t("chat:back_to_chat_list")}
+            </Button>
+          </Link>
+
+          <ChatConversation chatId={id as string} />
```
Comment (notmd) [inline website/src/pages/chat/[id].tsx:29]: Isn't the id already string?

notmd (approved):
lgtm

website/src/pages/chat/[id].tsx
Comment (AbdBarho) [reply website/src/pages/chat/[id].tsx:29]: true, this is a remanent from and old idea I had where the id was part of the query parameters, and in that case, `id` was of type `string | string[]`, I will change it after work.

PR #1921, opened 2023-03-01
Add chat.json for Thai

wannaphong:
Add chat.json for Thai from #1915


AbdBarho (approved):
That was fast! thank you

Issue #1770, opened 2023-02-20
Configure pytest in SFT pipeline


maw501:
As mentioned in #1368, the SFT pipeline would benefit from moving some of the modules in the [`model_training`](https://github.com/LAION-AI/Open-Assistant/tree/main/model/model_training) directory around so it's easier to work with `pytest`, e.g. as described [here](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html).

bethanyconnolly:
Hi, I can take a look at this problem and set up pytest here.

theblackcat102:
Cool

bethanyconnolly:
closed by #1922

PR #1922, opened 2023-03-01
Configure pytest in sft pipeline #1770

bethanyconnolly:
Both the model_training and reward/instructor modules can now be installed and pytest can be run for each of them. I did this by adding a pyproject.toml file for each, downgrading pandas to prevent a conflict and  removing a faulty test. Some of the tests are very slow or require some more work so I have added skips to these and will open a new issue for those to be worked on in future.


model/model_training/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+    "evaluate==0.4.0",
+    "gdown",
+    "trlx @ git+https://github.com/CarperAI/trlx.git@b91da7b03d8e9fa0c0d6dce10a8f2611aca3013f",
+    "nltk==3.8.1",
+    "numpy>=1.22.4",
+    "py7zr",
+    "PyYAML>=6.0",
+    "scikit-learn==1.2.0",
+    "torch>=1.12.1",
+    "transformers==4.25.1",
```
andreaskoepf, line 26: please increase transformers to 4.26.1

theblackcat102:
@bethanyconnolly could you update the readme with instruction to run pytest?

theblackcat102 (approved):
Note, requirements.txt files are removed, so I am not sure if it would impact anyone else workflow. If we need it we could consider readding it.

PR #1610, opened 2023-02-15
Resolves #1286

BrianArbuckle:
I have added a cocktail dataset to Hugging Face as well as the loading script to the repo.

PR #1923, opened 2023-03-01
Updates based on #1610. Resolves #1286

### LINKED ISSUE #1286: Create cocktail recipes dataset
This would involve taking a table scraped cocktail recipes and converting them to a dataset. 

1. I'll need some pointers on how to format the data. 
2. I'll dig for the source of each recipe (they are integers, I am looking for the table that was associated with the source).

BrianArbuckle:
@dctanner noticed and error, which I fixed on the [Hugging Face Dataset](https://huggingface.co/datasets/brianarbuckle/cocktail_recipes) as well as replaced the `loading_script.py` with a notebook importer `loading_notebook.ipynb`.


olliestanley:
How come we are replacing a loading script with a notebook here? My understanding was that notebooks are for code which generates the data originally and uploads it to HF in the OA-compatible format, and then there should be no need for a specific script/notebook to load the dataset and instead it can just be added [here](https://github.com/LAION-AI/Open-Assistant/blob/main/data/datasets/__init__.py).

Issue #1943, opened 2023-03-03
Markdown inline code rendering broken


someone13574:
When putting inline code using back ticks there seems to be an indentation after it. Here are a screen shot of the problem.

Raw text
![image](https://user-images.githubusercontent.com/81528246/222616506-c29f4ace-a609-413b-955d-1ef4a0fd5e96.png "Raw text")

Rendered text
![image](https://user-images.githubusercontent.com/81528246/222616568-769bb964-2733-483c-9de1-2447fb45b9fe.png "Rendered text")

PR #1945, opened 2023-03-03
Markdown inline code rendering broken resolved

austraCalvin:
Issue to resolve: https://github.com/LAION-AI/Open-Assistant/issues/1943

I think those style props from 'Chakra UI' are unnecessary.


AbdBarho:
Thank you, how does this look like on an RTL language like arabic?

austraCalvin:
> Thank you, how does this look like on an RTL language like arabic?

notmd (approved):
lgtm, thank you!

PR #1946, opened 2023-03-03
[MRG] [model_training]📝 update model training doc

HuaizhengZhang:
- [x] move the .gitignore to model_training
- [x] code can not be run in python3.8&3.9, so add a reminder for 3.10
- [x] export shared util functions to avoid errors
- [x] fix the wandb issues


HuaizhengZhang:
better to use `squash` merge. 
forget to `pre-commit` at the beginning :)

model/README.md
```diff
@@ -30,6 +33,9 @@ files.
 
 ```bash
 cd model_training
+# export shared modules
+export PYTHONPATH=$PYTHONPATH:../../oasst-shared
```
andreaskoepf, line 37: This should not be necessary, normally you simply install the oasst-shared module (see readme there https://github.com/LAION-AI/Open-Assistant/tree/main/oasst-shared ). I will check the dependencies .. if mostly related to data this will be resolved as soon by a oasst-data is module which can be installed via pip ...

andreaskoepf:
> better to use `squash` merge.

Thanks for the reminder .. squash is the default mode for this repo.

sanagno (approved):
Looks good

model/README.md
HuaizhengZhang: may we provide this as an option? It seems `ossst-data` is still being updated frequently, so maybe just exporting it instead of installing it should also be ok

andreaskoepf (approved):

andreaskoepf:
@HuaizhengZhang could you resolve the .gitignore conflicts?

theblackcat102 (approved):
LGTM

PR #1954, opened 2023-03-03
Extended export script

yk:
- adds events (emojis, ratings, rankings) to export trees
- add anonymizer to anonymize user and message IDs in a stable way

### DIFF
--- a/backend/export.py
+++ b/backend/export.py
@@ -26,7 +26,7 @@ def fetch_tree_ids(
     db: Session,
     state_filter: Optional[TreeState] = None,
     lang: Optional[str] = None,
-    limit: int = 0,
+    limit: Optional[int] = None,
 ) -> list[tuple[UUID, TreeState]]:
     tree_qry = (
         db.query(MessageTreeState)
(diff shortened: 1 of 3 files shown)


backend/export.py
```diff
@@ -59,10 +72,45 @@ def fetch_tree_messages(
...
         qry = qry.filter(Message.review_result)
+    if limit is not None and limit > 0:
+        qry = qry.limit(limit)
 
     return qry.all()
 
 
+def get_events_for_messages(db: Session, message_ids: list[UUID]) -> dict[UUID, ExportMessageEvent]:
+    events = {}
+    emojis = db.query(MessageEmoji).filter(MessageEmoji.message_id.in_(message_ids)).all()
```
andreaskoepf, line 83: (We have to test this with the the production db size, I had similar code before for labels which I re-wrote after the query took > 0.5h to execute for 100k messages .. i.e. `fetch_tree_messages_and_avg_labels` function does the join directly .. but needs to be tried)

backend/export.py
```diff
@@ -71,6 +119,7 @@ def fetch_tree_messages_and_avg_labels(
     prompts_only: bool = False,
     lang: Optional[str] = None,
     review_result: Optional[bool] = None,
+    limit: int = 0,
```
andreaskoepf, line 122: Why zero instead of `Optional[int] = None` ? Since everything is none-able in python I always use None to indicate 'not set' instead of a special in-domain value (which I would use in C style code) ... but is a matter of taste of course.

andreaskoepf (approved):

backend/export.py
yk: fair point. let's test and see

PR #1956, opened 2023-03-04
Fix discord bot workflow

AbdBarho:
Commits on main are failing because of a new js discord bot without updating  the container. https://github.com/LAION-AI/Open-Assistant/commits/main

I left in the old py-bot in place for this fix, it might make sense to use the new bot at some point.


andreaskoepf (approved):
good catch! thanks for quick fix!

PR #1959, opened 2023-03-04
enable `strictBindCallApply`

AbdBarho:
Part of https://github.com/LAION-AI/Open-Assistant/issues/879

I chose to use `any` for some of the types, since these are all stories and  not high priority to type them correctly.


notmd (approved):
Thanks

PR #1961, opened 2023-03-04
added async postgres to inference

yk:


andreaskoepf (approved):
I hope no await was forgotten ;-) ...

(For actual concurrency inside the DB the isolation level is most important (default is Read Committed), if there are access patterns that deadlock async client code can make them more likely to occur (need to be addressed anyway))

olliestanley (approved):

PR #1963, opened 2023-03-04
Fix some null checks

AbdBarho:
null checks will take a while, I fixed like 15 errors, with roughly 80 remaining

Part of #879

### DIFF
--- a/website/.eslintrc.json
+++ b/website/.eslintrc.json
@@ -14,7 +14,8 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "@typescript-eslint/no-non-null-assertion": "off"
   },
   "plugins": ["simple-import-sort", "unused-imports"]
 }
--- a/website/src/lib/oasst_api_client.ts
+++ b/website/src/lib/oasst_api_client.ts
@@ -69,7 +69,7 @@ export class OasstApiClient {
     });
 
     if (resp.status === 204) {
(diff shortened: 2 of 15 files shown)


website/.eslintrc.json
```diff
@@ -14,7 +14,8 @@
         "varsIgnorePattern": "^_",
         "caughtErrorsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "@typescript-eslint/no-non-null-assertion": "off"
```
AbdBarho, line 18: I also gave up, sorry for being against the idea before.

website/src/lib/oasst_api_client.ts
```diff
@@ -69,7 +69,7 @@ export class OasstApiClient {
     });
 
     if (resp.status === 204) {
-      return null;
+      return null as T;
```
AbdBarho, line 72: I don't really like this, but otherwise we would need to re-write all api function return types

website/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "jsx": "preserve",
     "baseUrl": ".",
     "strictBindCallApply": true,
-    "strictFunctionTypes": true
+    "strictFunctionTypes": true,
+    "strictNullChecks": false
```
AbdBarho, line 20: I put it to false because there is still a lot of work remaining

notmd (approved):
Thanks, strictNullChecks and noImplicitAny probably take most of the work when migrating to strict mode.

PR #1966, opened 2023-03-04
Update hyperparameters to reflect InstructGPT

dvruette:
I noticed that there are some non-standard hyperparameter values (namely: Adam betas and weight decay), so I suggest considering experimenting with the values proposed by InstructGPT [1], which are quite standard for LLM training AFAIK:

> All models are trained with the Adam optimizer, with β1 = 0.9 and β2 = 0.95.

> We train our SFT models for 16 epochs with residual dropout of 0.2. We use a cosine LR schedule
down to 10% of the original learning rate, with no learning rate warmup. For our 1.3B and 6B
models, we use an LR of 9.65e-6 and a batch size of 32. For 175B, we use a LR of 5.03e-6 and
a batch size of 8.

Unfortunately, I don't have easy access to compute power, so I will have to let someone else validate whether or not these changes have any (desired) effect.

[1] https://arxiv.org/abs/2203.02155


andreaskoepf (approved):
Thanks a lot for looking into this. If you want to run experiments on our HW please ping me on OA discord.

HuaizhengZhang:
> Thanks a lot for looking into this. If you want to run experiments on our HW please ping me on OA discord.

I was wondering if your HW resources (GPUs) are still adequate.

Do you need some donations?

I really appreciate your work

model/model_training/trainer_sft.py
```diff
@@ -252,6 +252,9 @@ def argument_parsing(notebook=False, notebook_args=None):
         gradient_accumulation_steps=training_conf.gradient_accumulation_steps,
         per_device_train_batch_size=training_conf.per_device_train_batch_size,
         per_device_eval_batch_size=training_conf.per_device_eval_batch_size,
+        adam_beta1=training_conf.adam_beta1,
+        adam_beta2=training_conf.adam_beta2,
+        adam_epsilon=training_conf.adam_epsilon,
```
sanagno, line 257: float(training_conf.adam_epsilon)

andreaskoepf: Is this cast a suggested change? Only for adam_epsilon or for beta1/beta2 also.

sanagno: yaml loads floats in scientific notation as strings unless you use a dot in the mantissa. So either we change the yaml or add a cast here

sanagno (approved):
thanks!

PR #1969, opened 2023-03-05
Show user score in header

### LINKED ISSUE #564: Show logged in user's score in header
Making the users own score more prominent will probably increase gamification. If we include it in the header, maybe next to the account button, it will always be visible. This would b similar to what stackoverflow.com does.

AbdBarho:
Closes #564

I am going through old issues and trying to finish them.

Add user score to header and account menu

image

This is hidden on narrow screens

image

The score could also be seen in the user menu (a bit redundant for large screens, but ok).

image

### DIFF
--- a/website/src/hooks/ui/useUserScore.ts
+++ b/website/src/hooks/ui/useUserScore.ts
@@ -1,0 +1,15 @@
+import { useSession } from "next-auth/react";
+import { get } from "src/lib/api";
+import { LeaderboardEntity, LeaderboardTimeFrame } from "src/types/Leaderboard";
+import uswSWRImmutable from "swr/immutable";
+
+export const useUserScore = () => {
+  const { status } = useSession();
+  const isLoggedIn = status === "authenticated";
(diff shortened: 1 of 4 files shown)


website/src/hooks/ui/useUserScore.ts
```diff
@@ -0,0 +1,18 @@
...
+import uswSWRImmutable from "swr/immutable";
+
+export const useUserScore = () => {
+  const { status } = useSession();
+  const isLoggedIn = status === "authenticated";
+  const { data: entries } = uswSWRImmutable<Partial<{ [time in LeaderboardTimeFrame]: LeaderboardEntity }>>(
+    isLoggedIn && "/api/user_stats",
+    get,
+    {
+      dedupingInterval: 1000 * 60, // once per minute
```
notmd, line 13: This should be `refreshInterval` right? Also can you add `keepPreviousData` to avoid flashing UI?

AbdBarho: not really, I am using `useSWRImmutable` since the user scores are updated once every hour or so (if I am not mistaken), so there is no need to fetch new data.

notmd (approved):

Issue #1985, opened 2023-03-06
Invalid avatar images


someone13574:
Some users who account is linked to a discord account have an avatar image which is pointing to an invalid image. This results in the text "*User's* avatar" which overlaps their username on the leader board. Here's an image:

![image](https://user-images.githubusercontent.com/81528246/222994135-4a7ab23f-7def-422a-9e55-700a889abf06.png)

It can be viewed simply by going to the overall leader board where there are three examples of the broken avatars.

owen-duncan-snobel:
I can take care of this one

AbdBarho:
Ok after a bit of digging, these are the callbacks that are called when a user logs in (using discord), in this order:
```
signIn callback
jwt
signIn event
```

Whenever a user does a full page navigation, the following gets called:
```
jwt
session
jwt
session
```

I assume 2 times because SSR + browser.


It does not seem to be the case that we have a lot of options when we want to do a refresh of the users Image, our session is 30 days, and unless you don't open the website in 30 days, `login` will never be called.

At the same time, updating the user's image on every session would be too aggressive, we are going to have to find to find a middleground.

PR #1986, opened 2023-03-06
fix for invalid avatarUrl error

owen-duncan-snobel:
fix for https://github.com/LAION-AI/Open-Assistant/issues/1985

If the avatarUrl ends up being invalid, it will use the same default value as if the avatarUrl is missing. 

I tried to fix it without useState but it caused flickering as Next.js kept trying to fetch the invalid url 

Screenshot 2023-03-05 at 8 57 24 PM


notmd:
Thank you. But this only mitigates the problem. The root problem is we don't update the discord avatar (except for the first login). Can you address this?  You need to add the logic to the `signIn` event in `[...nextauth].ts` file

owen-duncan-snobel:
> Thank you. But this only mitigates the problem. The root problem is we don't update the discord avatar (except for the first login). Can you address this? You need to add the logic to the `signIn` event in `[...nextauth].ts` file

I can try and attempt this tonight. Would the behaviour you expect then be to update the url every sign in regardless of it changing? Or first get the user then check if the profile.image differs from the user.image?

AbdBarho:
I am not really sure how to go on about this, is `signin` called only when the user signs in with username and pass? or is an outdated session also considered a sign in? we need to update the images just the right amount of times.

notmd:
> > Thank you. But this only mitigates the problem. The root problem is we don't update the discord avatar (except for the first login). Can you address this? You need to add the logic to the `signIn` event in `[...nextauth].ts` file
> 
> I can try and attempt this tonight. Would the behaviour you expect then be to update the image url every sign in regardless of it changing? Or first get the user then check if the profile.image differs from the user.image?

Update when image_url change is good, but only update when user signin with discord.

> I am not really sure how to go on about this, is `signin` called only when the user signs in with username and pass? or is an outdated session also considered a sign in? we need to update the images just the right amount of times.

`signin` is called when the user sign in succeed.

AbdBarho:
> signin is called when the user sign in succeed, no matter what sign in method they use

My question was more on the behavior of `next-auth`, if it would call `signin` after a discord session has expired, or just `session`.

I signed in to our website with my discord only once since we created the app, it won't help us if `signin` is called only once.

I am going to check the behavior

owen-duncan-snobel:
> > signin is called when the user sign in succeed, no matter what sign in method they use
> 
> My question was more on the behavior of `next-auth`, if it would call `signin` after a discord session has expired, or just `session`.
> 
> I signed in to our website with my discord only once since we created the app, it won't help us if `signin` is called only once.
> 
> I am going to check the behavior

Let me know what you find/found

notmd (approved):
Thank you!

PR #1989, opened 2023-03-06
Added moon icon for when the light mode is active, removed oddly big gap between sign-in and color-mode button

theopfr:
Hey, I improved the color-mode button so that when the light mode is active a moon-icon is shown:
image

A small, second change:
When the user is not signed in the gap between the sign-in button and the color-mode toggler is oddly big:
image
The reason is that even when the user score is ``null`` the div around the ``<UserScore />`` component is still rendered:
```html
<div className="hidden md:block">
    <UserScore />
</div>
```
It seems that this can be solved by moving the outter div inside the ``<UserScore />`` component:
```html
<div className="hidden md:block">
    <HStack gap="1" title={t("score")}>
        <>{score}</>
        <Icon as={Star} fill="gold" w="5" h="5" color="gold" />
    </HStack>
</div>
```

### DIFF
--- a/website/src/components/Header/ColorModeToggler.jsx
+++ b/website/src/components/Header/ColorModeToggler.jsx
@@ -1,12 +1,12 @@
 import { Button, useColorMode } from "@chakra-ui/react";
-import { Sun } from "lucide-react";
+import { Sun, Moon } from "lucide-react";
 
 const ColorModeToggler = () => {
   const { colorMode, toggleColorMode } = useColorMode();
 
   return (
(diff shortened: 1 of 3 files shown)


AbdBarho:
The `UserScore` component is used in the drop down menu, what is the behavior there?

theopfr:
You are right, it doesn't work with the dropdown. I now used the "Show" chakra component instead and it looks like it works now also in the dropdown.

olliestanley:
I went ahead and fixed the pre-commit fail on this as it was a tiny change, hopefully ready for re-review now

AbdBarho (approved):
Thanks!

to @olliestanley as well.

PR #2005, opened 2023-03-07
Add SFT params random_offset_probability, label_masking

andreaskoepf:
Added SFT training parameters:
- `random_offset_probability` (float, default: 0.5): probability of random offset into conversations when conversation > max_length 
- `label_masking` (bool, default: true): if true only loss for tokens of assistant replies is calculated else for all tokens (including prompter)

### DIFF
--- a/model/model_training/custom_datasets/dialogue_collator.py
+++ b/model/model_training/custom_datasets/dialogue_collator.py
@@ -5,7 +5,7 @@ from typing import Optional, Union
 import numpy as np
 import torch
 from torch.nn import functional as F
-from transformers.tokenization_utils_base import PaddingStrategy, PreTrainedTokenizerBase
+from transformers.tokenization_utils_base import PaddingStrategy, PreTrainedTokenizerBase, TruncationStrategy
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


model/model_training/custom_datasets/dialogue_collator.py
```diff
@@ -54,13 +64,20 @@ def __call__(self, features):
                     list(map(lambda x: x[1], flatten_message["offset_mapping"])),
```
sanagno, line 64: this should not be required now

model/model_training/custom_datasets/dialogue_collator.py
```diff
@@ -29,18 +31,26 @@ def __call__(self, features):
...
+            if random.random() < self.random_offset_probability:
+                truncation = TruncationStrategy.DO_NOT_TRUNCATE
+                max_length = None
+            else:
+                truncation = TruncationStrategy.LONGEST_FIRST
+                max_length = self.max_length
+
+            # append eos token to each messages
+            assert self.tokenizer.eos_token
+            messages = [m + self.tokenizer.eos_token for m in messages]
```
sanagno, line 43: Did we converge to this? Using the end of prompt-reply the same as eos? Not totally against it if eveyone agrees

andreaskoepf: EOS at end of messages simplifies the decoding, TRLX also relies on eos tokens (i.e. would either have to be patched or eos could be set to `<human>`). The format is a compromise between framing format v3 and  old `<bot> <user>` format v2.

PR #2010, opened 2023-03-07
Residual dropout support

dvruette:
Adds functions for patching residual dropout into `GPTNeoX` models. Patching it like this should be fine since it won't be needed for inference, so we don't have to fork/subclass the model class. It should also be quite easy to add this to other models (Galactica, ...) if needed.

Residual dropout has been used during InstructGPT [1] training, presumably to help prevent overfitting:
> We train our SFT models for 16 epochs with residual dropout of 0.2.

I also include some config changes from the latest run

[1] https://arxiv.org/pdf/2203.02155.pdf

### DIFF
--- a/model/model_training/.gitignore
+++ b/model/model_training/.gitignore
@@ -1,3 +1,4 @@
 .cache
 wandb
 saved_model
+.saved_models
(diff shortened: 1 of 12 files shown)


andreaskoepf (approved):
Thanks a lot! Let's give it a try ;-) ..

Issue #1991, opened 2023-03-06
Wrong number of max steps calculated


sanagno:
The number of max steps calculated by [huggingface](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L1663) during training seem to be wrong. 

This is probably due to some inconsistency in our [sampler](https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/utils.py#L24). 

Need to fix this, as learning rate updates in the scheduler seem to be wrong

PR #2011, opened 2023-03-07
Group samples by sequence length

sanagno:
Should decrease the amount of padding by about ~40% with some potential trade-off on the randomness between the sample order in epochs.

Feel free to suggest changes


yk:
I like the minor shuffling. In addition to that, I'd also introduce before the minor shuffling that messages of the same length are shuffled within their length-group. i.e. all messages of length 50 are shuffled, all messages of length 51 are shuffled, etc. before the minor-shuffling step

model/model_training/configs/config.yaml
```diff
@@ -54,10 +54,11 @@ defaults:
   samples_mixing: false # uses collator that mixes samples in the batch to create a single sample with possible multiple tasks within
   verbose: false
   output_dir: saved_model
-  use_custom_sampler: false
+  use_custom_sampler: true
```
andreaskoepf, line 57: as long as the custom sampler is broken (e.g. doesn't use all samples if only single dataset is active) I would rather keep it disabled by default.

sanagno:
Also solves #1991

theblackcat102 (approved):
Looks good to me

PR #2019, opened 2023-03-08
Refresh active user avatar

AbdBarho:
Close https://github.com/LAION-AI/Open-Assistant/issues/1985
Close  #1986

Update user's avatar once a day, only for users who are actively using the website.


AbdBarho:
> #1986 will mitigate the issue in case we haven't updated the avatar yet. So it should not be closed by this

True! I updated it to "related"

notmd (approved):

PR #2021, opened 2023-03-08
add `include_user` params in /messages/cursor` endpoint

notmd:
I need this to display user information in message table

### DIFF
--- a/backend/oasst_backend/models/message.py
+++ b/backend/oasst_backend/models/message.py
@@ -5,8 +5,8 @@ from uuid import UUID, uuid4
 
 import sqlalchemy as sa
 import sqlalchemy.dialects.postgresql as pg
+from oasst_backend.models import User
 from oasst_backend.models.db_payload import MessagePayload
-from oasst_backend.models.user import User
 from oasst_shared.exceptions.oasst_api_error import OasstError, OasstErrorCode
 from pydantic import PrivateAttr
 from sqlalchemy import false
--- a/backend/oasst_backend/models/message.py
+++ b/backend/oasst_backend/models/message.py
@@ -5,11 +5,12 @@ from uuid import UUID, uuid4
 
 import sqlalchemy as sa
 import sqlalchemy.dialects.postgresql as pg
+from oasst_backend.models.user import User
 from oasst_backend.models.db_payload import MessagePayload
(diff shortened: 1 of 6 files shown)


backend/oasst_backend/prompt_repository.py
```diff
@@ -888,14 +888,27 @@ def fetch_message_with_max_children(self, message: Message | UUID) -> tuple[Mess
         max_message = max(tree, key=lambda m: m.children_count)
         return max_message, [m for m in tree if m.parent_id == max_message.id]
 
-    def _add_user_emojis_all(self, qry: Query) -> list[Message]:
+    def _add_user_emojis_all(self, qry: Query, include_user=False) -> list[Message]:
```
andreaskoepf, line 891: Could you add python type annotation?

backend/oasst_backend/models/message.py
```diff
@@ -6,6 +6,7 @@
 import sqlalchemy as sa
 import sqlalchemy.dialects.postgresql as pg
 from oasst_backend.models.db_payload import MessagePayload
+from oasst_backend.models.user import User
```
andreaskoepf, line 9: can also be imported from modes (omitting .user)

backend/oasst_backend/prompt_repository.py
```diff
@@ -888,14 +888,27 @@ def fetch_message_with_max_children(self, message: Message | UUID) -> tuple[Mess
         max_message = max(tree, key=lambda m: m.children_count)
         return max_message, [m for m in tree if m.parent_id == max_message.id]
 
-    def _add_user_emojis_all(self, qry: Query) -> list[Message]:
```
andreaskoepf, line 891: `_add_user_emojis_all` is a function that is used in several places as a wrapper around Message queries .. IMO this is a bit hacky to add an `include_user` param which is only needed at one place.

andreaskoepf (approved):

backend/oasst_backend/prompt_repository.py
notmd: Done

backend/oasst_backend/models/message.py
notmd: Unfortunately, it causes circular import error
image

andreaskoepf: ah, ok, we are in the Message model file ;-) ..

PR #2033, opened 2023-03-10
Add support for flash attention

dvruette:
Along with some miscellaneous changes, this PR adds support for patching FlashAttention [1] into GPTNeoX models. This not only makes training more efficient, but allows for almost arbitrarily long context sizes.

A few caveats apply:
- FlashAttention requires inputs to be in `fp16` and on CUDA.
- FlashAttention outputs slightly different results compared to the reference GPTNeoX implementation. For a single layer, the outputs are almost the same (less than `1e-3` absolute difference), but the error can accumulate during the forward pass. If and how this affects generated samples will have to be evaluated once a patched model has been trained. Worst case, flash attention will have to be used during inference.
  - For an alternative memory-efficient attention implementation, we try out xFormers [2].
- Non-contiguous attention masks are not supported by FlashAttention. This is not an issue as long as attention masking is only used to mask padding (which is usually the case).
- FlashAttention does not return attention scores, so even if `output_attentions=True` is passed to the model, the attention scores will just be `None`.

Misc. changes include:
- Exporting script for uploading models to HF (authored by @andreaskoepf, committed by me).
- Fixed a bug where only parts of the config was logged to WandB.
- Fixed a bug where WandB logging was activated even when disabled in the config.
- Updated training config

[1] https://github.com/HazyResearch/flash-attention
[2] https://github.com/facebookresearch/xformers


andreaskoepf:
Regarding training config I suggest that we create a new separate block for parameters used for training the current SFT-1 12B model (and also for all future models which we release) .. this will make it more transparent what we did and thereby simpler for others to replicate our results.

sanagno:
Yes that makes sense, we should also tag the commit/release with which the model was trained

sanagno (approved):

Issue #2022, opened 2023-03-08
Keep dark mode when refresh


alexgas:
There is a blink on the UI when refresh the page and the dark mode is activated. 

Maybe it's possible to keep that user preference in a cookie, so when the page is refreshed there won't be any blink.

jwstolk:
Same when switching languages.

PR #2034, opened 2023-03-10
Darkmode persists on refresh.

Anantha-Kandrapu:
Used cookies to store the user's choice.

### #2022 
https://github.com/LAION-AI/Open-Assistant/issues/2022


website/src/pages/_app.tsx
```diff
@@ -39,12 +42,19 @@ function MyApp({ Component, pageProps: { session, cookies, ...pageProps } }: App
...
   );
 }
-export { getServerSideProps };
+
+MyApp.getInitialProps = async (appContext) => {
+  const cookies = appContext["ctx"].req?.client?.parser?.incoming?.cookies ?? { "chakra-ui-color-mode": "light" };
+  return {
+    cookies,
+  };
+};
```
AbdBarho, line 58: Not merging with `App.getInitialProps`

website/src/styles/Chakra.tsx
```diff
@@ -1,24 +1,22 @@
...
+  const cookieColorModeManager = {
+    type: "cookie",
+    ssr: true,
+    get: () => {
+      return cookieColorValue;
+    },
+    set: () => {
+      console.log("Current colorMode set to :", cookieColorValue);
+    },
+  } as const;
```
AbdBarho, line 15: how it this different from the default `cookieStorageManagerSSR`?

website/src/styles/Chakra.tsx
```diff
@@ -1,24 +1,22 @@
-import { ChakraProvider, cookieStorageManagerSSR, localStorageManager } from "@chakra-ui/react";
-
+import { ChakraProvider } from "@chakra-ui/react";
 import { theme } from "./Theme";
 
 export function Chakra({ cookies, children }) {
-  const colorModeManager = typeof cookies === "string" ? cookieStorageManagerSSR(cookies) : localStorageManager;
+  const cookieColorValue = cookies["chakra-ui-color-mode"] ?? "light";
+  const cookieColorModeManager = {
```
AbdBarho, line 6: `useMemo`

website/src/styles/Chakra.tsx
```diff
@@ -1,24 +1,22 @@
-import { ChakraProvider, cookieStorageManagerSSR, localStorageManager } from "@chakra-ui/react";
-
+import { ChakraProvider } from "@chakra-ui/react";
 import { theme } from "./Theme";
 
 export function Chakra({ cookies, children }) {
-  const colorModeManager = typeof cookies === "string" ? cookieStorageManagerSSR(cookies) : localStorageManager;
+  const cookieColorValue = cookies["chakra-ui-color-mode"] ?? "light";
```
AbdBarho, line 5: it seems to me that it should be possible to use `document.cookie` as a parameter instead of having re-do everything ourselves.

website/src/components/Header/ColorModeToggler.jsx
```diff
@@ -1,9 +1,16 @@
...
+import { useEffect } from "react";
+import { useCookies } from "react-cookie";
 
 const ColorModeToggler = () => {
+  const [cookies, setCookie] = useCookies(["chakra-ui-color-mode"]);
   const { colorMode, toggleColorMode } = useColorMode();
 
+  useEffect(() => {
+    setCookie("chakra-ui-color-mode", colorMode, { path: "/" });
+  }, [toggleColorMode, colorMode]);
```
AbdBarho, line 12: `toggleColorMode` is not needed, right?

AbdBarho (changes requested):
Thank you.


The problem does not seem to be with our code, but more in the way next js / chakra works.

In the [official chakra ui docs](https://chakra-ui.com/docs/styled-system/color-mode#add-colormodemanager-optional-for-ssr), they show how to setup the color mode, which is identical to our use case, however, [nextjs custom app documentation](https://nextjs.org/docs/advanced-features/custom-app#caveats) mentions specifically that 
> App does not support Next.js [Data Fetching methods](https://nextjs.org/docs/basic-features/data-fetching/overview) like [getStaticProps](https://nextjs.org/docs/basic-features/data-fetching/get-static-props) or [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props).

So in this case, there is a mistake in chakra ui docs.

Some users reported a solution which would work but is not easy to implement: https://github.com/chakra-ui/chakra-ui/issues/3135#issuecomment-796886860

Using `getInitialProps` will [disable some of the features of nextjs](https://nextjs.org/docs/advanced-features/custom-app#caveats) and requires attention to be implemented correctly


In either case, I don't think it is necessary for us to re-invent the wheel.
If there is an easy solution to globally deal with the problem we can use that, otherwise this is an overkill for the edge of refreshing the page.

website/src/components/Header/ColorModeToggler.jsx
Anantha-Kandrapu: Yeah, it is not.

website/src/styles/Chakra.tsx
Anantha-Kandrapu: yeah.

Anantha-Kandrapu: I'll try doing that.

website/src/styles/Chakra.tsx
Anantha-Kandrapu: I tried passing the cookies to ```cookieStorageManagerSSR``` it didn't work. I don't know why.

website/src/pages/_app.tsx
Anantha-Kandrapu: I understand.

PR #2039, opened 2023-03-11
Show correct user avatars on leaderboard

AbdBarho:
Fixes a bug introduced in #1986 where going forwards or backwards in the in the users list on the leaderboard would not update the avatar.

### DIFF
--- a/website/src/components/LeaderboardTable/LeaderboardTable.tsx
+++ b/website/src/components/LeaderboardTable/LeaderboardTable.tsx
@@ -67,7 +67,7 @@ export const LeaderboardTable = ({
           const display_name = getValue();
           return (
             <div className="flex flex-row items-center gap-2">
-              <UserAvatar key={display_name} displayName={display_name} avatarUrl={row.original.image} />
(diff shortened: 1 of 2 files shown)


website/src/components/LeaderboardTable/LeaderboardTable.tsx
```diff
@@ -63,18 +63,21 @@ export const LeaderboardTable = ({
...
-            ) : (
-              getValue()
-            )}
-          </div>
-        ),
+        cell: ({ getValue, row }) => {
+          const display_name = getValue();
+          return (
+            <div className="flex flex-row items-center gap-2">
+              <UserAvatar key={display_name} displayName={display_name} avatarUrl={row.original.image} />
```
notmd, line 70: We already have an effect in the `UserAvatar` component, so why do we need key here?

notmd (approved):

website/src/components/LeaderboardTable/LeaderboardTable.tsx
AbdBarho: I wanted to be _really sure_ that it gets changed, I will remove this.

PR #2040, opened 2023-03-11
Trainer fixes

johnflux:
Fixes these two tools to work with a model just trained with trainer_sft

image
image

### DIFF
--- a/model/model_training/README.md
+++ b/model/model_training/README.md
@@ -89,17 +89,22 @@ python trainer_rl.py --configs defaults_rlhf
 
 ## Test your model
 
-You can itneractively test your model like this:
+You can interactively test your model like this:
 
 ```bash
-python tools/model_cli.py --model_path <saved_path/huggingface>
+python3 tools/model_cli.py --model_path <saved_path/huggingface>
+# For example, if you trained with the default config:
+python3 tools/model_cli.py --model_path saved_model
+# Add --8bit  if it is an 8bit model
 ```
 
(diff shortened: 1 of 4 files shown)


sanagno:
Great thanks!

Can you also change the pre/post-processing according to the new format?

Instead of `format_pairs` it should be something like
`
  pairs = format_pair(pairs)
  pairs = ["{}{}".format(p, eos_token) for p in pairs]
  pairs.append(QA_SPECIAL_TOKENS["Answer"])
`

and then in the end we should remove the `tokenizer.eos_token_id` from the response if generated. You can take a look https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/custom_datasets/dialogue_collator.py#L40

johnflux:
> Great thanks!
> 
> Can you also change the pre/post-processing according to the new format?
> 
> Instead of `format_pairs` it should be something like `pairs = format_pair(pairs) pairs = ["{}{}".format(p, eos_token) for p in pairs] pairs.append(QA_SPECIAL_TOKENS["Answer"])`
> 
> and then in the end we should remove the `tokenizer.eos_token_id` from the response if generated. You can take a look https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/custom_datasets/dialogue_collator.py#L40

Should we basically stop on the first eos_token_id but then stop on a Answer token as a fallback?

sanagno (approved):
Looks good!

Issue #2044, opened 2023-03-11
Synthesise prosocial data from reddit


shahules786:
The dataset here https://www.kaggle.com/datasets/pavellexyr/one-million-reddit-confessions contains 1 million posts from subreddits like r/confession. This can be used to generate data for training safety models if augmented using the prosocial-dialog dataset.

PR #2045, opened 2023-03-11
Reddit confessions dataset

shahules786:
As mentioned in #2044 Augmented dataset for training safety models. 
- The dataset is made in the same format as prosocial-dialog for ease of use. 
- safety label confidence is added to enable users to filter data according to a threshold of one's own preference.


shahules786:
@bitplane @Vechtomov Can you guys take a look at this PR? I am already using this data to train safety models.

shahules786:
@Vechtomov I see, that makes sense. I can add a `safety datasets` folder and document the format and datasets that are currently used for the purpose. What do you think?

Vechtomov:
Yes, please. I don't think we need separate safety datasets folder, for now changes in `__init.py__` and readme are enough.

shahules786:
@Vechtomov Done.

data/datasets/README.md
```diff
@@ -48,6 +48,18 @@ proposed. The text-only dataset must include the following columns:
 2. **SOURCE** (string)
 3. **METADATA** (JSON string, optional)
 
+### **Prosocial format**
```
Vechtomov, line 51: I prefer "Safety format" if it is common for all future safety datasets

data/datasets/README.md
```diff
@@ -48,6 +48,18 @@ proposed. The text-only dataset must include the following columns:
...
+
+For datasets that are intended to be used to train safety models, prosocial
+format is proposed. The format is given below
+
+1. **USER** (string)
+2. **RESPONSE** (string, optional)
+3. **ROTs** (List)
+4. **SAFETY_LABEL** (string - categorical)
+5. **EPISODE_DONE** (bool)
+6. **SOURCE** (string,optional)
```
Vechtomov, line 61: This needs more detailed description. I didn't realize it until I checked ProsocialDialog paper.

data/datasets/__init__.py
```diff
@@ -13,3 +13,8 @@
     "youtube_subs_howto100M": "totuta/youtube_subs_howto100M",
     "iapp_wiki_qa_squad": "wannaphong/iapp_wiki_qa_squad_oa",
 }
+
+SAFETY_DATASETS = {
+    "prosocial-dialog": "allenai/prosocial-dialog",
+    "prosocial-confessions": "shahules786/prosocial-confessions",
```
Vechtomov, line 19: How these two datasets can be fit in one training process if `RESPONSE` column isn't provided in `prosocial-confessions` dataset?

Vechtomov (changes requested):

data/datasets/__init__.py
shahules786: The current response for the user's prompt is not taken into consideration while modeling. Response is used to form contextual information about past conversations. In prosocial confessions, each example is a single conversation indicated by `episode_done` set to True.

data/datasets/README.md
shahules786: @Vechtomov  Done.

Vechtomov: Can you also add a link to the paper?

Vechtomov (approved):

PR #2053, opened 2023-03-12
Choice to add global system-prefix to the assistant during changes

sanagno:
Set the flag `use_system_prefix` to true and specify a `system_prefix` to be used during training.

It also refactors pre-processing for some datasets. Some of the datasets used are now deprecated and need to be reorganized.


model/model_training/configs/config.yaml
```diff
@@ -60,6 +60,11 @@ defaults:
   residual_dropout: 0.2
   use_flash_attention: true
   sort_by_length: false
+  use_system_prefix: false
+  system_prefix:
+    "You are Joi, a large language model trained by Open-Assistant. Answer as
+    concisely as possible.\nKnowledge cutoff: 2021-09-01\nCurrent date:
```
andreaskoepf, line 66: I am against adding this to the default config without any validation. Adding a constant prefix for training will it likely make harder to evaluate the model because the "oa model" will be conditioned on the existing of this exact prefix.

model/model_training/custom_datasets/formatting.py
```diff
@@ -7,12 +7,20 @@
...
+    )
+
 
+def format_pairs(pairs, eos_token, add_initial_reply_token=False):
+    conversations = [
+        "{}{}{}".format(QA_SPECIAL_TOKENS["Question" if i % 2 == 0 else "Answer"], pairs[i], eos_token)
+        for i in range(len(pairs))
+    ]
+    if add_initial_reply_token:
+        conversations.appned(QA_SPECIAL_TOKENS["Answer"])
```
theblackcat102, line 24: should be `append`?

model/model_training/custom_datasets/__init__.py
```diff
@@ -2,7 +2,7 @@
     High level functions for model training
 """
 from custom_datasets.oasst_dataset import load_oasst_export
-from custom_datasets.prompt_dialogue import OAPrivate, PrivateInstructionTuning
+from custom_datasets.prompt_dialogue import OAPrivate  # , PrivateInstructionTuning
```
theblackcat102, line 5: why don't just remove it entirely, instead of leaving a comment, since the class was removed already?

theblackcat102 (changes requested):
there's a typo in formating and necessary legacy comments (remove privatetuning?)

model/model_training/custom_datasets/formatting.py
sanagno: good catch

sanagno:
Most of the datasets are legacy and depend too much on the formatting to easily change. If we want to keep them we need to update them. I updated some of them but need to be a bit more precise about it.

theblackcat102 (approved):

PR #2058, opened 2023-03-13
TV and Movie dialogue corpus

sedthh:
A huge collection (2781 texts total) of TV and Movie dialgoues and transcripts either collected or scraped from transcript sites.
See datacard for more information: https://huggingface.co/datasets/sedthh/tv_dialogue

They are in a standardized format ([speaker] speech) and the idea is that these dialogues should help the Assistant to:
1) mimic the style of famous characters
2) generate new episodes of certain shows

NOTE: this is part 1, there will be a part 2 based entirely on ForeverDreaming's transcripts
NOTE2: the notebook only contains the crawler for IMSDb, the parsers for the rest of the sources were omited as they had a low code quality to save time


Vechtomov:
Can you add other source crawlers and parsers? And also converting to final format. Code quality is not so important as the ability to reproduce data mining.

sedthh:
I can add the converter for the outside sources that need conversion if that's enough.

Vechtomov:
Yes, please add everything you have 🙂

Vechtomov (approved):
Very cool, thanks!

PR #2071, opened 2023-03-14
Reward Model Training

dvruette:
This code is currently WIP and needs some additional work to get data ingested in the right way. Model and training loop are yet untested, but should be very close to working.


model/model_training/custom_datasets/ranking_collator.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def process_one(self, messages: Iterable[str], return_length: bool = False):
+        messages = list(messages)
+
+        if random.random() < self.random_offset_probability:
+            truncation = TruncationStrategy.DO_NOT_TRUNCATE
+            max_length = None
+        else:
+            truncation = TruncationStrategy.LONGEST_FIRST
+            max_length = self.max_length
```
sanagno, line 28: Should we always truncate the begging for RM training?

model/model_training/custom_datasets/ranking_collator.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # append eos token to each messages
+        assert self.tokenizer.eos_token
+        messages = [m + self.tokenizer.eos_token for m in messages]
```
sanagno, line 32: deprecated if you merge main

model/model_training/custom_datasets/ranking_collator.py
andreaskoepf: good point, IMO for RM we need to make sure to always include the last message (e.g. the assistant reply).

PR #2078, opened 2023-03-15
Add initial Ansible playbooks for inference server and worker

olliestanley:
We will likely need to include more environmental variables in these configs in future

### DIFF
--- a/ansible/inference/deploy-worker.yaml
+++ b/ansible/inference/deploy-worker.yaml
@@ -6,6 +6,7 @@
   vars:
     stack_name: "dev"
     image_tag: latest
+    worker_port: 8080
     backend_url: "ws://localhost:8000"
     api_key: "1234"
   tasks:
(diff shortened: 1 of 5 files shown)


ansible/inference/deploy-worker.yaml
```diff
@@ -0,0 +1,40 @@
...
+        state: started
+        recreate: true
+        pull: true
+        restart_policy: always
+        network_mode: "oasst-worker-{{ stack_name }}"
+        env:
+          BACKEND_URL: "{{ backend_url }}"
+          API_KEY: "{{ api_key }}"
+        ports:
+          - "{{ worker_port }}:8080"
```
yk, line 40: I don't think we need this. the worker connects to the backend, no port mapping necessary

ansible/inference/deploy-worker.yaml
```diff
@@ -0,0 +1,40 @@
+# Ansible playbook to setup inference worker
+
+- name: Deploy inference
+  hosts: dev
+  gather_facts: true
+  vars:
```
yk, line 6: in here we also need the "parallelism" variable defined in the worker-full dockerfile

yk (approved):
awesome, thank you

Issue #2077, opened 2023-03-15
Allow moderator to revert a message deletion


Shadowner:
Currently, if a mod deletes a message, it's permanent. If the mod accidentally clicks or misjudges a message, it could be great to have the ability to revert the deletion.

0x22almostEvil:
Sure, good suggestion!

PR #2079, opened 2023-03-15
[#2077] Possbility to revert message deletion

Shadowner:
Hello, here is the PR for that feature ! 

On the web side I've added a button when a message is deleted in the context menu : 
image

Also added that feature to the messages board :
image

With a warning message on reintroduction : 
image

Might need to still do the translation part, however I don't see any "admin" local :/ 

Original issue : https://github.com/LAION-AI/Open-Assistant/issues/2077


backend/oasst_backend/prompt_repository.py
```diff
@@ -1069,6 +1069,39 @@ def mark_messages_deleted(self, messages: Message | UUID | list[Message | UUID],
...
+        else:
+            raise OasstError("Server error", OasstErrorCode.SERVER_ERROR1, HTTPStatus.INTERNAL_SERVER_ERROR)
+
+        query = update(Message).where(Message.id == message_id).values(deleted=False)
+        self.db.execute(query)
+
+        if recursive:
+            parent_id = None
+            if isinstance(message, UUID):
+                parent_id = self.db.query(Message.parent_id).where(Message.id == message).first()[0]
```
andreaskoepf, line 1092: `Message.id == message` looks a bit strange, should it be `message_id`? Do you want to undelete all descendants? The easiest would be to use the `fetch_message_descendants()` method and use ORM to set deleted = False .. usually message trees have <100 messages.

backend/oasst_backend/api/v1/messages.py
```diff
@@ -325,7 +325,19 @@ def mark_message_deleted(
...
+@router.put("/{message_id}/reintroduce", status_code=HTTP_202_ACCEPTED)
+def reintroduce_message(
+    *,
+    message_id: UUID,
+    frontend_user: deps.FrontendUserId = Depends(deps.get_frontend_user_id),
+    api_client: ApiClient = Depends(deps.get_api_client),
+    db: Session = Depends(deps.get_db),
+):
+    pr = PromptRepository(db, api_client, frontend_user=frontend_user)
+    pr.reintroduce_deleted_message(message_id)
```
andreaskoepf, line 337: The recursive param should be exposed in the api vi a query param (maybe default true).

backend/oasst_backend/prompt_repository.py
```diff
@@ -1069,6 +1069,39 @@ def mark_messages_deleted(self, messages: Message | UUID | list[Message | UUID],
...
+                parent_id = message.parent_id
+
+            if parent_id is None:
+                return
+
+            # Fetching the entire parent_message so there is no parent_id query executed after
+            parent_message: Message = self.db.query(Message).where(Message.id == parent_id).first()
+
+            if parent_message is not None:
+                self.reintroduce_deleted_message(parent_message)
```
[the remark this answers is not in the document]
Shadowner: Totally, it recurse until a parent is not deleted or it's the root of the tree and marked those message as not deleted

backend/oasst_backend/prompt_repository.py
Shadowner: yes it could totally be messag_id, don't know why i did not do that.

I don't want to undelete every descendant only the parents like 
A (deleted)
|->A.b (deleted)
|-> A.c (deleted)

I undelete A.c so now the tree mus look like this : 

A (not deleted)
|->A.b (deleted)
|-> A.c (not deleted)

A.b stay deleted because it is not a parent of A.c

backend/oasst_backend/api/v1/messages.py
Shadowner: Will totally do that then !

backend/oasst_backend/api/v1/messages.py
```diff
@@ -325,7 +325,19 @@ def mark_message_deleted(
     pr.mark_messages_deleted(message_id)
 
 
-@router.post("/{message_id}/emoji", response_model=protocol.Message)
+@router.put("/{message_id}/reintroduce", status_code=HTTP_202_ACCEPTED)
```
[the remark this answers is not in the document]
Shadowner: let's go for 'undelete'

backend/oasst_backend/prompt_repository.py
andreaskoepf: oh, undeleting parents is indeed a VERY important and should not be optional - otherwise a broken tree is created which is bad - especially if they are still active (will show broken-conversation errors to users and will require manual fixes of the DB). If the recursive parameter was for the trace to root of the tree please do do not expose it to the REST API (and maybe could we rename it to undelete_parents)  ... it must not be possible to undelete an isolated child of a deleted parent via API.

Shadowner: That's what I thought, I will remove the recursive option and make that `undelete` always check if the parent is deleted or not, and if it's deleted proceed to `undelete` the parent (doing this recursively ofc)

I'm not sure to rename the function `undelete_parent` just keep `undelete` and change the description to say that it `undelete` parents recursively :)

notmd (approved):
web change lgtm

andreaskoepf (approved):

andreaskoepf:
@Shadowner could you please resolve the conflicts?

PR #2088, opened 2023-03-16
Unit Test Issue2006b

sir-braz:
npm run jest


website/src/components/Buttons/LikertButtons.tsx
```diff
@@ -25,7 +25,17 @@ export const LikertButtons = ({
...
+            key={value}
+            id={`radio-${value}`}
+            value={value}
+            borderColor="gray.400"
+            data-cy="radio-option"
+            size="md"
+            padding="0.6em"
+          >
+            <label htmlFor={`radio-${value}`}>{value}</label>
+          </Radio>
```
notmd, line 38: This change may not be correct. We may have multiple elements with the same id when rendering many `LikertButtons` component

website/src/components/Buttons/LikertButtons.tsx
notmd, line 1: This component is used in many places. Why did you delete it?

notmd:
Most of the tested components are just static components. I'm not sure if these tests will have many benefits or just create noise.

website/src/components/Buttons/LikertButtons.tsx
sir-braz: I'm going to reverse, Because what I did was wrong, I thought I had to delete what I did

sir-braz:
@notmd But are the tests working?

PR #2089, opened 2023-03-16
improvment_FAQ_docs

kamyardl:


olliestanley:
There is quite a bit of information added here which is incorrect (e.g. the frameworks in use, the license used for OA, describing the goals of OA are if they are already completed features). Tbh the existing content of the FAQ also needs changing, some of it is already quite outdated

PR #2092, opened 2023-03-16
Fix passing parallelism var to worker in Ansible playbook

olliestanley:
Corrects an error in #2078. The variable was added to the playbook but not passed as an environment variable to the worker container


yk (approved):
thank you

PR #2097, opened 2023-03-17
Add GitHub OAuth support to inference server

olliestanley:
As we are currently not allowing email authentication for the inference server it seems important to provide an alternative to Discord, which is not available in all countries. We will need to get a GitHub OAuth App client ID/secret for OA to fully enable this.


AbdBarho:
I am thinking if it makes sense to add github login to the website.

linking accounts might be a bit of a problem, but should be possible.

olliestanley:
> I am thinking if it makes sense to add github login to the website.
> 
> linking accounts might be a bit of a problem, but should be possible.

Yeah, account linking is the thing. Will make it a bit annoying to handle moving the data collection app to full backend auth. Inference is a lot easier as we're using separate accounts I think

AbdBarho:
Yeah, now that I think about it, even if we link the frontend accounts, we still need (a lot of) work to also link the accounts in the data backend.

Maybe linking won't be that easy....

In any case, if our users should be able to access the inference with their github account, and if this access should be available through our website, then we need to add github login support to the website.

Issue #2107, opened 2023-03-18
Add model & configuration options for chat


AbdBarho:
https://github.com/LAION-AI/Open-Assistant/blob/bad50cc2f521f679a6c21829a6a1912e282ff40f/inference/server/oasst_inference_server/routes/configs.py

PR #2115, opened 2023-03-19
Chat config UI

notmd:
There is a small bug in the chat settings button that doesn't display correctly on mobile. This is due to our layout which is really buggy lol. I'm going to fix it in the later PR. Also, this PR doesn't store the config in local storage yet.

**UPDATE**: I added preset, which will allow using pre-defined parameters in the server. Users could also select a custom preset to customize the paras as they want. I also remove the params translation as they really hard to correctly translate. We can add a tooltip for long descriptions like in the playground of Open in the future. 
Current params config:
| name               | min | max  |   |
|--------------------|-----|------|---|
| temperature        | 0   | 1.5  |   |
| max_new_tokens     | 1   | 1024 |   |
| top_p              | 0   | 1    |   |
| repetition_penalty | 1   | 3    |   |
| top_k              | 0   | 1    |  exponential  function: `Math.round(Math.pow(32000, x))`  |

UI
image
Custom preset
image


website/public/locales/en/chat.json
```diff
@@ -3,5 +3,14 @@
   "chat_date": "{{val, datetime}}",
   "empty": "(empty)",
   "queue_info": "Your message is queued, you are at position {{ queuePosition, number, integer }} in the queue.",
-  "your_chats": "Your chats"
+  "your_chats": "Your chats",
+  "chat_config": {
+    "title": "Chat configuration",
+    "model": "Model",
+    "top_k": "top_k",
```
notmd, line 10: I dont understand what this param is.

website/public/locales/en/chat.json
```diff
@@ -3,5 +3,14 @@
...
+    "top_p": "Top P",
+    "temperature": "Temperature",
+    "repetition_penalty": "repetition_penalty",
```
notmd, line 13: Same here

website/src/components/Chat/ChatSection.tsx
```diff
@@ -0,0 +1,212 @@
...
+  {
+    key: "repetition_penalty",
+    min: 1,
+    max: 2,
+  },
+  {
+    key: "top_k",
+    min: 1,
+    max: 2,
+  },
```
notmd, line 126: I'm not sure about the range of these parameters. But for `repetition_penalty` and `top_k` it should be greater than 1 otherwise the inference backend will crash

notmd: cc @yk

website/src/components/Chat/ChatSection.tsx
```diff
@@ -0,0 +1,212 @@
...
+type UseChatConfigForm = UseFormReturn<WorkParametersInput>;
+
+export const ChatSection = ({ chatId }: { chatId: string }) => {
+  const { control, getValues, register, setValue } = useForm<WorkParametersInput>({
+    defaultValues: {
+      temperature: 0.7,
+      top_p: 0.5,
+      top_k: 1.2,
+      repetition_penalty: 1.2,
+      max_new_tokens: 256,
```
notmd, line 48: Here are the default params. I just fill them randomly, please suggest

website/public/locales/en/chat.json
AbdBarho: This might be a bit of a long read, but It makes all of these parameters much clearer:
https://huggingface.co/blog/how-to-generate

You can also look up the documentation of huggingface

website/src/components/Chat/ChatSection.tsx
AbdBarho: I think the idea of #2107 was to make a request to `INFERENCE/config/models` which will give a pre-defined set of configurations that you could send with the requests. I hope I interpreted this correctly.

notmd: > I think the idea of #2107 was to make a request to `INFERENCE/config/models` which will give a pre-defined set of configurations that you could send with the requests. I hope I interpreted this correctly.

Let me ask Yannic about this. I'm not sure what behavior he wants.

website/src/components/Chat/ChatSection.tsx
andreaskoepf: `temperature`: Temperature of 1.0 is neutral, lowering this makes the sampling greedier, maybe range could be 0-4
`top_p`: is something in the interval 0-1, percentage of probability mass to consider
`top_k`: is a number of the top k logits to consider, can be 1 to number of vocabulary (usually something like 32k), especially smaller values like 50 are interesting, maybe use exponential transfer function of slider value x -> `Math.pow(32000, x)` for 0 <= x <= 1...
`repetition_penalty`: 1.0 means no penalty, paper that invents this recommends: "We find that using a greedy sampling and θ ≈ 1.2 yields a good balance between truthful generation and lack of repetition." .. maybe range could also be 1-3 (not sure if someone uses higher values)

website/src/components/Chat/ChatContext.tsx
```diff
@@ -0,0 +1,16 @@
...
+export type ChatContext = {
+  modelInfos: ModelInfo[];
+};
+
+const chatContext = createContext<ChatContext>({} as ChatContext);
+
+export const useChatContext = () => useContext(chatContext);
+
+export const ChatContextProvider = ({ children, ...props }: PropsWithChildren<ChatContext>) => {
+  const value = useMemo(() => props, [props]);
```
AbdBarho, line 13: does this make any difference?

website/src/components/Chat/ChatSection.tsx
```diff
@@ -0,0 +1,283 @@
...
+  );
+
+  const handleShowSliderChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const checked = e.target.checked;
+      setFormValue(name, checked ? customPresetDefaultValue[name] : null);
+      setShowSlider(e.target.checked);
+      if (checked && value === null) {
+        // setValue(customPresetDefaultValue[name]!);
+      }
```
AbdBarho, line 233: Is this still needed?

website/src/pages/chat/[id].tsx
```diff
@@ -47,9 +53,15 @@ export const getServerSideProps: GetServerSideProps<ChatProps, { id: string }> =
       notFound: true,
     };
   }
+
+  const { data: modelInfos } = await axios.get<ModelInfo[]>("/configs/models", {
+    baseURL: process.env.INFERENCE_SERVER_HOST,
+  });
+
```
Comment (AbdBarho) [inline website/src/pages/chat/[id].tsx:60]: I wonder why do we do this server side?

website/src/components/Chat/ChatContext.tsx
notmd: I update to explicit memo the `modelInfos`. Haven't tested the performance, but it's recommended to memo context value.

website/src/components/Chat/ChatSection.tsx
notmd: Deleted

website/src/pages/chat/[id].tsx
Comment (notmd) [reply website/src/pages/chat/[id].tsx:60]: SSR will make the code much cleaner. Fetching data in the client will require calling `setValue` on success to update the default form value. Also, we don't have to check where it's loading or not.

AbdBarho (approved):

PR #2137, opened 2023-03-20
Added ENABLE_CHAT to chat id page and docker-compose

yk:


website/src/pages/chat/[id].tsx
```diff
@@ -32,19 +36,17 @@ const Chat = ({ id }: { id: string }) => {
...
 Chat.getLayout = getDashboardLayout;
 
-export const getServerSideProps: GetServerSideProps<{ id: string }, { id: string }> = async ({
-  locale = "en",
-  params,
-}) => ({
+export const getServerSideProps: GetServerSideProps<ChatProps, { id: string }> = async ({ locale = "en", params }) => ({
   props: {
     id: params.id,
+    enabled: process.env.NODE_ENV === "development" || process.env.ENABLE_CHAT === "true",
```
Comment (notmd) [inline website/src/pages/chat/[id].tsx:49]: Maybe we can through 404 directly here?

notmd (approved):
We might need a more general solution for receiving env variables in future. Pretty pain we have to pass them manually on every page

yk:
Thanks for the hint! I've now used this to directly throw a 404 if chat is not enabled.
https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#notfound

I anticipate once chat is a "basic" feature, we can remove this code again

andreaskoepf (approved):

PR #2138, opened 2023-03-20
Fix stacking of pos/neg logits for RM compute_metrics

andreaskoepf:
Node positive & negative examples are stacked along last dimension so that compute_metrics receives predictions of size (val_examles, 2). The old code did not work with deepspeed if the number of validation examples was not divisible by the total batch size without remainder.


sanagno (approved):

yk (approved):
Thank you

PR #2141, opened 2023-03-20
Cancelling pending chat messages of user on submission of a new message

yk:


inference/server/oasst_inference_server/chat_repository.py
```diff
@@ -30,6 +31,9 @@ async def start_work(
         logger.debug(f"Starting work on message {message_id}")
         message = await self.get_assistant_message_by_id(message_id)
 
+        if message.state == inference.MessageState.cancelled:
+            raise chat_schema.MessageCancelled(message_id=message_id)
```
andreaskoepf, line 35: name of exception classes in python usually have Error/Exception suffix

inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -110,6 +111,29 @@ async def initiate_assistant_message(
...
+                models.DbChat.user_id == self.user_id,
+            )
+        )
+
+        pending_msgs: list[models.DbMessage] = (await self.session.exec(pending_msg_query)).all()
+        for pending_msg in pending_msgs:
+            logger.warning(
+                f"User {self.user_id} has a pending message {pending_msg.id} in chat {pending_msg.chat_id}. Cancelling..."
+            )
+            pending_msg.state = inference.MessageState.cancelled
```
andreaskoepf, line 133: does this work without explicit session.add() .. asking because the sqlmodel examples always have this extra `session.add(modified_obj)` call .. I never tested if it is actually necessary.

andreaskoepf (approved):

inference/server/oasst_inference_server/user_chat_repository.py
yk: yes, session.add is only necessary if the object is detached from the session previously, but probably doesn't hurt to have it
https://docs.sqlalchemy.org/en/20/orm/session_basics.html#adding-new-or-existing-items

PR #2143, opened 2023-03-20
add tool to encrypt & decrypt weights

iurimatias:
This script is to create XOR diff between fine-tuned weights and the original weights.
The llama weights fine tuned weights on OA HF friendly and sharded by 10GB, so first the original llama weights should be loaded and saved using `python model_enc.py convert_llama_to_hf...` Then a XOR diff can be created between the original and the finetuned weights.

HOWEVER unfortunately when encrypting and decrypting it back the weights file is not exactly the same and wouldn't load. Creating this as a draft hoping to get a second pair of eyes here on what the issue might be.

### DIFF
--- a/model/model_training/tools/convert_llama_weights_to_hf.py
+++ b/model/model_training/tools/convert_llama_weights_to_hf.py
@@ -1,0 +1,282 @@
+# from https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py
+# Copyright 2022 EleutherAI and The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


bitplane:
I don't know much about the llama stuff, but-

If you take random bits and xor them together in different ways, do you get the original output?

Are there range constraints on the things you're xoring? Like, negative numbers etc? I'm guessing binary format is important here but I've not run the code.

I highly recommend throwing in some pytest unit tests, try hard to make them fail, then fix the failures. This usually works for me and let's me write systems that are far more complex than I can understand.

iurimatias:
@bitplane yes as in (A XOR B) XOR A == B , the files are supose to be the same format.

**However**. I believe I found the issue, the files don't necessary have the same size, so the code needs to take this into account. I'll try to fix it tomorrow if no one gets to it first.

bitplane:
Yep I get for, I just wanted to point out that the test driven approach is a great way to force you to understand and communicate each unit's intent.

> **However**. I believe I found the issue, the files don't necessary have the same size, so the code needs to take this into account. I'll try to fix it tomorrow if no one gets to it first.

Cool. Can you do hmac from a seed? That works with arbitrary sized files. Chaining SHA1 is reasonably fast and resistant enough to collisions.

yk:
sorry to get a bit late to this. did you fix the issue?

iurimatias:
Yes should all be working

andreaskoepf:
@iurimatias thanks a lot for your work on this ..

PR #2147, opened 2023-03-21
Added a timeout for inference messages

yk:


andreaskoepf (approved):
ok, seems the session.add() is really not necessary .. probably can remove this at several places in the backend as well...

PR #2158, opened 2023-03-21
Derive RewardModel from PreTrainedModel

andreaskoepf:
First simple way to derive our RewardModel from PreTrainedModel so simplify loading .. still requires full download of base model which could probably be simplified in the future.

### DIFF
--- a/model/model_training/configs/config.yaml
+++ b/model/model_training/configs/config.yaml
@@ -102,8 +102,7 @@ oasst_export_eu:
   datasets:
     - oasst_export:
         lang: "en,es,de,fr"
-        #top_k: 2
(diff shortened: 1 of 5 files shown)


sanagno (approved):
LGTM

dvruette:
Looks good! Does this work with AutoModel? If so, how exactly? I would expect it to work with `AutoModelForSequenceClassification.from_pretrained`.

Also, why not inherit from `GPTNeoXModel` could save us the trouble of loading the base model. We could basically just copy `GPTNeoXForCausalLM` and add pooling between hidden states and head.

theblackcat102 (approved):
LGTM

PR #2159, opened 2023-03-21
Data analysis suite

MattAlexMiracle:
This adds in two scripts to compare rankings and votes.
The rankings is relatively simple and computes the expected kendall-tau correlation of a user to the consensus vote.
This is mostly for spotting bad actors. Currently it does not find statistically significant outliers that are also bad actors

This doesn't mean that individual bad actors don't exist, but it does mean that either
1. there are so many of them, the ranking results are so far beyond repair that the merged rankings themselves make no sense
2. there are so few of them, that they overall don't affect the results sufficiently negatively to be noticable

The truth is probably somewhere in the middle: since we only have ~3 rankers per message, if we had one that is consistently screwing with things, he would be hard to detect since by sheer luck he would agree enough with another ranker to overall vanish in the correlations.

The vote aggregation script is much more sophisticated and computes optimal weightings for each power-user based on aggregations of non-power users:
I now have an initial "smart" selection of messages based on a hierarchical consensus finding scheme:
The idea is to split the voters into an "aggregation" and a "re-rating" group:
The "aggregation" group is comprised of the set of users with less than X number of ratings.
These users get aggregated into one mean (i.e. weighting 1:1) and act as the anchor.
Users with more than X ratings are used to compute weightings, specifically we attempt to reweight them such that
min ∑ (A u - (y + agg)/2)²
subject to 
∑ u = 1

with "A" being the matrix of votes done by people in the "re-rating" group and "agg" being the aggregated voter vector.
y is simply the current weighting result y = A u_prev.
I.e. "y" is the current target, assuming that the current  u weightings are correct.
After finding "better" uweightings in the least-squares optimization, I recompute the new target y and compute the next u until a fixed point is reached.

The end result is a weighting of the "re-rating" group in u and a priority we can read from y.

However, the choice between what people we put into the "aggregation" and "re-rating" group is pretty arbitrary based on a split with who has more than X ratings:
- if we have low X then we have much much more people which can get re-weighted, but our "agg" anchor might be a lot more inaccurate
- if we have high X we have a much more statistically stable "agg" anchor, but less weighting potential.

The way I decided to solve this is by simply running many iterations 50≤X≤500 which gives us all (reasonable) configurations, and then giving a confidence score based on how often a message is in the top P% of messages.

The method outputs a CSV file with 
- "message_id"
- "message_tree_id"
- "fracs" (what fraction of the ensemble the message_id is in)
- "in_naive" (whether it would be in the naive "1 user; 1 vote" model)

### DIFF
--- a/scripts/postprocessing/importance_selection.py
+++ b/scripts/postprocessing/importance_selection.py
@@ -1,0 +1,261 @@
+import logging
+import warnings
+
+import numpy as np
+import pandas as pd
+import psycopg2
+import shap
+from scipy.special import softmax
+from scipy.stats import dirichlet
+from tqdm import trange
+from tqdm.contrib.logging import logging_redirect_tqdm
+from xgboost import XGBRegressor
+
+
+def get_df(study_label):
+    conn = psycopg2.connect("host=0.0.0.0 port=5432 user=postgres password=postgres dbname=postgres")
+    # Define the SQL query
+    query = "SELECT message_id, labels, message.user_id FROM text_labels JOIN message ON message_id = message.id;"
+
+    # Read the query results into a Pandas dataframe
+    df = pd.read_sql(query, con=conn)
+    print(df.head())
+    # Close the database connection
+    conn.close()
+    users = set()
+    for row in df.itertuples(index=False):
+        row = row._asdict()
+        users.add(str(row["user_id"]))
+    users = list(users)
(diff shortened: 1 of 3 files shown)


yk:
fyi I ran pre-commit, so if you want to change something, pull first

yk (approved):

PR #2163, opened 2023-03-21
Added param configs from eval to pass to the frontend

yk:


oasst-shared/oasst_shared/schemas/inference.py
```diff
@@ -8,20 +8,15 @@
...
-DEFAULT_MODEL_NAME = "distilgpt2"
 
 DEFAULT_MODEL_LENGTHS = {
     "_lorem": 1024,
-    "gpt2": 1024,
-    "distilgpt2": 1024,
-    "OpenAssistant/oasst-sft-1-pythia6b": 2048,
     "OpenAssistant/oasst-sft-1-pythia-12b": 2048,
     "OpenAssistant/oasst_sft_llama_7b_mask_1000": 2048,
     "OpenAssistant/oasst_sft_llama_13b_mask_1500": 2048,
```
olliestanley, line 19: We can't offer these LLaMa models now either, right?

yk: yes we can, as long as we don't profit monetarily. we can even provide the outputs without providing the actual weights until we figure out the license stuff

olliestanley: Ok - is the reason for removing 30B just that it would require too much compute then?

yk: yes exactly

PR #2166, opened 2023-03-22
check `per_digit_tokens` exists

HuangLK:
While running pytest, got "AttributeError: 'Namespace' object has no attribute 'per_digit_tokens'". So I added code to check whether the variable per_digit_tokens exists.


sanagno (approved):
thanks!

sanagno:
We should also add it to the defaults in the config.yaml

PR #2169, opened 2023-03-22
Fix: Correctly sending error to waiting client on message abort

yk:


yk:
sorry the github diff here is quite off, compare in vscode, it handles the diff better

yk:
I also added a setting to configure allowed models. could do it in a separate PR, but it's quite a small addition

andreaskoepf (approved):

olliestanley (approved):

PR #2172, opened 2023-03-22
rework layout

notmd:
Our layout now uses `position: fixed` instead of tricky `display: grid` and `over-flow: auto` to make content scroll. Manually verified all pages that there is no regression. The UI still looks the same as before.

### DIFF
--- a/website/src/components/SideMenu.tsx
+++ b/website/src/components/SideMenu.tsx
@@ -1,11 +1,13 @@
-import { Button, Card, Text, Tooltip } from "@chakra-ui/react";
+import { Box, Button, Card, Text, Tooltip } from "@chakra-ui/react";
 import { LucideIcon } from "lucide-react";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import { useTranslation } from "next-i18next";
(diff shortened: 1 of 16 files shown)


website/src/components/SideMenu.tsx
```diff
@@ -1,75 +1,84 @@
...
+              className="hidden lg:hidden sm:block"
+            >
+              <Link
                 key={item.labelID}
-                fontFamily="inter"
-                label={label}
-                placement="right"
-                className="hidden lg:hidden sm:block"
+                href={item.pathname}
+                target={!item.target ? "_self" : item.target}
```
AbdBarho, line 57: `item.target ?? "_self"`

AbdBarho (commented):
Thank you, looks good on mobile?

notmd:
> Thank you, looks good on mobile?

Yes. On mobile the navbar is now fixed too

website/src/components/SideMenu.tsx
notmd: > `item.target ?? "_self"`

That's is the old code, I didn't want to touch it.

AbdBarho (approved):

PR #2175, opened 2023-03-22
Add ability for inference backend to revoke auth refresh tokens

olliestanley:
An admin endpoint is provided to revoke all refresh tokens for a given user ID. This is achieved by storing hashed refresh tokens in a backend DB table.


olliestanley:
> quick question, where would would we need this functionality? for malicious actors? when the user logs out?

I don't know if we strictly need it, but it's a common practice. As refresh tokens are much longer-lived compared to access tokens, if one were compromised it potentially gives the attacker a long window, so it might be useful to have this ability in place in case it's ever required

AbdBarho (approved):

olliestanley:
I think this is ready to merge now but will wait for @yk review

yk (approved):
thank you

PR #2179, opened 2023-03-23
RL training

sanagno:
First changes to make RL work with the latest configuration.

Important issues to be resolved:
1. Make sure RewardModel is loaded correctly.
2. TRLX modifies our returned strings here https://github.com/CarperAI/trlx/blob/main/trlx/trainer/accelerate_base_trainer.py#L197, so input to the RM does not have our special tokens.
3. A lot of testing


theblackcat102 (approved):
Mostly just code from other branch, but changes for RL training dataset is good

andreaskoepf:
It is a bit hard to see what is new here. Would it be possible to generate a diff between this branch and main and apply it to a fresh branch that branched off of current head of main?

PR #2182, opened 2023-03-23
Add GPTNeoXRewardModel

andreaskoepf:
- add specific reward model based on GPTNeoX: GPTNeoXRewardModel
- support HF auto loading after import of `import models.reward_model`
- make model_training a python module

Note: Running `pip install -e .` in `model/` (parent directory of model_training) is required after this PR has been merged.

### DIFF
--- a/model/model_training/models/reward_model.py
+++ b/model/model_training/models/reward_model.py
@@ -42,8 +42,8 @@ class GPTNeoXRewardModel(GPTNeoXPreTrainedModel):
         if type(config) == GPTNeoXConfig:
             # When a normal GPTNeoX was loaded it will be converted into a reward model.
             # The direct `type(config) == GPTNeoXConfig` comparison is used (instead of
-            # `isinstance()``) since the configuration class of the reward model is also
-            # derived form `GPTNeoXConfig``.
+            # `isinstance()`) since the configuration class of the reward model is also
+            # derived form `GPTNeoXConfig`.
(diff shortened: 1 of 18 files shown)


model/model_training/models/reward_model.py
```diff
@@ -1,68 +1,94 @@
...
-            cache_dir=cache_dir,
-        )
 
-        self.transformer = transformer
-        self.out_proj = nn.Linear(transformer.config.hidden_size, 1)
+class GPTNeoXRewardModel(GPTNeoXPreTrainedModel):
+    config_class = GPTNeoXRewardModelConfig
+
+    def __init__(self, config):
+        if type(config) == GPTNeoXConfig:
```
dvruette, line 42: I personally prefer `isinstance(config, GPTNeoXConfig)`

dvruette (approved):
Very nice, looks good! One thing worth considering: It would be straight forward to abstract the model to a general `GPTNeoXForSequenceClassification` by adding `num_labels` as a config option. Just an idea, not necessary.

model/model_training/models/reward_model.py
andreaskoepf: The direct test was done intentionally, since `GPTNeoXRewardModelConfig` is also derived from `GPTNeoXConfig` ... e.g. the line was added to support the case when a normal non-rm model name is specified. I will add a comment.

PR #2194, opened 2023-03-24
Introduce model configs to abstract pairings of models and hardware

yk:
- introduces the concept of "model config" defined in `model_configs.py` that can be referenced by the frontend by name
- adjusted dockerfiles to automatically determine model id and quantization from a given config name
- introduced an OOM test to find out sequence length and batch size limits on different GPUs, results in notion

### DIFF
--- a/inference/server/oasst_inference_server/routes/configs.py
+++ b/inference/server/oasst_inference_server/routes/configs.py
@@ -65,12 +65,11 @@ DEFAULT_PARAMETER_CONFIGS = [
 
 @router.get("/model_configs")
 async def get_model_configs() -> list[ModelConfigInfo]:
-    allowed_models_list = settings.allowed_models_list
     return [
         ModelConfigInfo(
             name=model_config_name,
             parameter_configs=DEFAULT_PARAMETER_CONFIGS,
         )
         for model_config_name in model_configs.MODEL_CONFIGS
-        if (settings.allowed_models == "*" or model_config_name in allowed_models_list)
+        if (settings.allowed_model_config_names == "*" or model_config_name in settings.allowed_model_config_names_list)
     ]
--- a/inference/server/oasst_inference_server/routes/configs.py
+++ b/inference/server/oasst_inference_server/routes/configs.py
@@ -1,6 +1,7 @@
 import fastapi
 import pydantic
 from oasst_inference_server.settings import settings
+from oasst_shared import model_configs
 from oasst_shared.schemas import inference
 
 router = fastapi.APIRouter(
@@ -12,10 +13,10 @@ router = fastapi.APIRouter(
 class ParameterConfig(pydantic.BaseModel):
     name: str
     description: str = ""
-    work_parameters: inference.WorkParametersInput
+    sampling_parameters: inference.SamplingParameters
 
 
-class ModelInfo(pydantic.BaseModel):
+class ModelConfigInfo(pydantic.BaseModel):
     name: str
     description: str = ""
     parameter_configs: list[ParameterConfig] = []
@@ -25,8 +26,7 @@ DEFAULT_PARAMETER_CONFIGS = [
     ParameterConfig(
         name="k50",
         description="Top-k sampling with k=50",
-        work_parameters=inference.WorkParametersInput(
-            model_name="_model_",
+        sampling_parameters=inference.SamplingParameters(
             top_k=50,
             top_p=0.95,
             temperature=1.0,
@@ -36,8 +36,7 @@ DEFAULT_PARAMETER_CONFIGS = [
     ParameterConfig(
         name="nucleus9",
         description="Nucleus sampling with p=0.9",
-        work_parameters=inference.WorkParametersInput(
-            model_name="_model_",
+        sampling_parameters=inference.SamplingParameters(
             top_p=0.9,
             temperature=0.8,
             repetition_penalty=1.2,
@@ -46,8 +45,7 @@ DEFAULT_PARAMETER_CONFIGS = [
     ParameterConfig(
         name="typical2",
         description="Typical sampling with p=0.2",
-        work_parameters=inference.WorkParametersInput(
-            model_name="_model_",
+        sampling_parameters=inference.SamplingParameters(
             temperature=0.8,
             typical_p=0.2,
             repetition_penalty=1.2,
@@ -56,8 +54,7 @@ DEFAULT_PARAMETER_CONFIGS = [
     ParameterConfig(
         name="typical3",
         description="Typical sampling with p=0.3",
-        work_parameters=inference.WorkParametersInput(
-            model_name="_model_",
+        sampling_parameters=inference.SamplingParameters(
             temperature=0.8,
             typical_p=0.3,
             repetition_penalty=1.2,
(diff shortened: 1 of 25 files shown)


inference/server/oasst_inference_server/routes/configs.py
```diff
@@ -36,8 +36,7 @@ class ModelInfo(pydantic.BaseModel):
     ParameterConfig(
         name="nucleus9",
         description="Nucleus sampling with p=0.9",
-        work_parameters=inference.WorkParametersInput(
-            model_name="_model_",
+        sampling_parameters=inference.SamplingParameters(
             top_p=0.9,
             temperature=0.8,
             repetition_penalty=1.2,
```
andreaskoepf, line 42: I have seen `1/0.83` as magic-value for repetition_penalty a couple of times which is also ~1.2 .. do you know where it comes from?

inference/server/oasst_inference_server/routes/configs.py
```diff
@@ -66,17 +63,13 @@ class ModelInfo(pydantic.BaseModel):
 ]
 
 
-@router.get("/models")
-async def get_models() -> list[ModelInfo]:
-    allowed_models_list = settings.allowed_models_list
+@router.get("/model_configs")
```
andreaskoepf, line 66: could also be `/models/configurations` .. but matter of taste/style

inference/server/oasst_inference_server/routes/configs.py
```diff
@@ -66,17 +63,13 @@ class ModelInfo(pydantic.BaseModel):
...
-                for config in DEFAULT_PARAMETER_CONFIGS
-            ],
+        ModelConfigInfo(
+            name=model_config_name,
+            parameter_configs=DEFAULT_PARAMETER_CONFIGS,
         )
-        for model_name in inference.DEFAULT_MODEL_LENGTHS.keys()
-        if (settings.allowed_models == "*" or model_name in allowed_models_list)
+        for model_config_name in model_configs.MODEL_CONFIGS
+        if (settings.allowed_model_config_names == "*" or model_config_name in settings.allowed_model_config_names_list)
```
andreaskoepf, line 74: maybe add regex match feature later?

inference/worker/__main__.py
```diff
@@ -23,19 +24,37 @@ def main():
...
-    if settings.model_id != "_lorem":
-        if "llama" in settings.model_id:
-            tokenizer: transformers.PreTrainedTokenizer = transformers.LlamaTokenizer.from_pretrained(settings.model_id)
+    model_config = model_configs.MODEL_CONFIGS.get(settings.model_config_name)
+    if model_config is None:
+        logger.error(f"Unknown model config name: {settings.model_config_name}")
+        sys.exit(2)
+
+    if not model_config.is_lorem:
+        if model_config.is_llama:
```
olliestanley, line 33: Minor but we could avoid the nesting here

inference/worker/__main__.py
```diff
@@ -23,19 +24,37 @@ def main():
...
+            tokenizer: transformers.PreTrainedTokenizer = transformers.LlamaTokenizer.from_pretrained(
```
andreaskoepf, line 34: is this special handling still necessary with latest transformers version? (e.g. maybe wee need to update our json config files?)

olliestanley (approved):

inference/worker/basic_hf_server.py
```diff
@@ -47,14 +48,20 @@ def terminate_server(signum, frame):
...
-        tokenizer = transformers.LlamaTokenizer.from_pretrained(settings.model_id)
-        model = transformers.LlamaForCausalLM.from_pretrained(settings.model_id, torch_dtype=config.torch_dtype)
+
+    model_config = model_configs.MODEL_CONFIGS.get(settings.model_config_name)
+    if model_config is None:
+        logger.error(f"Unknown model config name: {settings.model_config_name}")
+        sys.exit(2)
+
+    logger.warning(f"Loading model {model_config.model_id}...")
+    if "llama" in model_config.model_id:
```
andreaskoepf, line 58: `model_config.is_llama`?

inference/worker/get_model_config_prop.py
```diff
@@ -0,0 +1,18 @@
...
+
+from oasst_shared import model_configs
+from settings import settings
+
+if __name__ == "__main__":
+    model_config = model_configs.MODEL_CONFIGS.get(settings.model_config_name)
+    if model_config is None:
+        print(f"Unknown model config name: {settings.model_config_name}")
+        sys.exit(2)
+    prop = sys.argv[1]
```
andreaskoepf, line 11: maybe add `len(sys.argv) > 1` check (since everything else is here handled gracefully)

inference/worker/interface.py
```diff
@@ -13,23 +13,19 @@ class GenerateStreamParameters(pydantic.BaseModel):
     temperature: float | None
     repetition_penalty: float | None
     seed: int | None
-    stop: list[str] = [
-        "<|endoftext|>",
-        "<|prompter|>",
-        "<|assistant|>",
-    ]  # TODO: make this a bit more workable because it's mutliple tokens
+    stop: list[str] = []
```
andreaskoepf, line 16: normal stop should always the eos token of the tokenizer ..

inference/worker/run_worker_container.sh
```diff
@@ -32,11 +31,10 @@ while true; do
     -e CUDA_VISIBLE_DEVICES=$gpus \
     -e LOGURU_LEVEL=$loguru_level \
     -e API_KEY=$api_key \
-    -e MODEL_ID=$model_id \
+    -e MODEL_CONFIG_NAME=$model_config_name \
     -e BACKEND_URL=$backend_url \
     -e HF_TOKEN=$HF_TOKEN \
     -e MAX_PARALLEL_REQUESTS=$max_parallel_requests \
-    -e QUANTIZE=$quantize \
     -e HF_HUB_ENABLE_HF_TRANSFER= \
```
andreaskoepf, line 38: just declare, no value like `true`?

oasst-shared/oasst_shared/model_configs.py
```diff
@@ -0,0 +1,83 @@
+import pydantic
+
+
+class ModelConfig(pydantic.BaseModel):
+    model_id: str
+    max_input_length: int = 512
+    max_total_length: int = 1024
+    quantized: bool = False
```
andreaskoepf, line 8: Quantize currently means float16 always? We may have to update this is in the future in case we want to support int8/int4 ..

andreaskoepf (approved):

inference/server/oasst_inference_server/routes/configs.py
yk: I called it "model_config" everywhere else, I felt this was ok

inference/server/oasst_inference_server/routes/configs.py
yk: no idea, I got this from you

inference/server/oasst_inference_server/routes/configs.py
yk: good idea

inference/worker/__main__.py
yk: with the latest verison, probably none, but the version I use right now is a very unholy merge of two forks...

yk: very likely we can remove this in the future once the llama models have arrived at the version used in text-generation-inference

inference/worker/interface.py
yk: that is happening anyway, independent of the tokens here. Moreover, putting eos token here confuses the stopping criteria, making it e.g. impossible to generate html

inference/worker/run_worker_container.sh
yk: correct. empty value meaning false.

oasst-shared/oasst_shared/model_configs.py
yk: Quantize means int8. it's a built-in of the HF server. by default our models are float16. int4 is going to be difficult...

PR #2206, opened 2023-03-25
Rl training

sanagno:
- Code to run RL
- Custom PPO trainer with correct formatting for RM model
- Change imports until they work
- Some refactoring


model/model_training/utils/ppo_utils.py
theblackcat102, line 1: Any reason why there's a patch of used transformer code not removed?

sanagno: yes, that is on me, will be removed

Aayog (approved):

sanagno:
@theblackcat102 @dvruette if everything looks good we can merge

model/model_training/tests/test_datasets.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 from model_training.custom_datasets import QA_DATASETS, SUMMARIZATION_DATASETS, get_one_dataset
 from model_training.custom_datasets.dialogue_collator import DialogueDataCollator
-from model_training.utils import get_tokenizer
+from model_training.utils.utils import get_tokenizer
```
andreaskoepf, line 6: Personally I don't like utils.utils .. could be solved via __init__.py  in utils dir ... but no reason not to merge this now.

andreaskoepf (approved):
GOOOOO! :-) .. please let's try to create smaller PRs in the future.

PR #2208, opened 2023-03-25
Use new inference model config / API

AbdBarho:
Blocked by #2207

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -9,7 +9,13 @@ import { useMessageVote } from "src/hooks/chat/useMessageVote";
 import { get, post } from "src/lib/api";
 import { handleChatEventStream, QueueInfo } from "src/lib/chat_stream";
 import { API_ROUTES } from "src/lib/routes";
-import { ChatItem, InferenceMessage, InferencePostMessageResponse, WorkParametersInput } from "src/types/Chat";
+import {
+  ChatConfigForm,
+  ChatItem,
+  InferenceMessage,
+  InferencePostMessageParams,
+  InferencePostMessageResponse,
+} from "src/types/Chat";
 import useSWR from "swr";
 
 import { BaseMessageEntry } from "../Messages/BaseMessageEntry";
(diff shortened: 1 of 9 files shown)


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -29,35 +35,34 @@ export const ChatConversation = ({ chatId }: ChatConversationProps) => {
...
-    const params = getValues();
-
-    const response: InferencePostMessageResponse = await post(API_ROUTES.CREATE_CHAT_MESSAGE, {
-      arg: {
-        chat_id: chatId,
-        content,
-        parent_id,
-        work_parameters: {
-          ...params,
-          top_k: params.top_k === null ? null : Math.floor(Math.pow(32000, params.top_k)),
```
notmd, line 57: Have you tested the top_k here? Andreas suggest an exponential function for this.

website/src/components/Chat/ChatSection.tsx
```diff
@@ -1,20 +1,25 @@
...
 import { ChatConfigDrawer } from "./ChatConfigDrawer";
 import { useChatContext } from "./ChatContext";
 import { ChatConversation } from "./ChatConversation";
 
 export const ChatSection = ({ chatId }: { chatId: string }) => {
   const { modelInfos } = useChatContext();
-  const form = useForm<WorkParametersInput>({
-    defaultValues: modelInfos[0].parameter_configs[0].work_parameters,
+
+  console.assert(modelInfos.length > 0, "No model config was found");
```
notmd, line 11: I think we can ignore this and assume the backend always returns at least 1 config.

website/src/components/Chat/ChatConversation.tsx
notmd: ah nvm I see the config below

AbdBarho: Yeah just for info: this works for the custom config where the slider goes between 0 and 1, but for all the presets it computes `pow(32000, 50)` or similar.

I did not find a clean way of changing this so I changed min and max. Arguably, no more than 1000 is needed, because the sampling probability is already so low it might as well be zero.

website/src/components/Chat/ChatSection.tsx
AbdBarho: Yes, this `console.assert` is there to inform other developers that we expect the array to have at least one. We know that, any new person who might look over this might not know this.

notmd (approved):

PR #2209, opened 2023-03-25
OA-261.tell.a.joke.dataset

mikegarts:


mikegarts:
Based on the https://huggingface.co/datasets/SocialGrep/one-million-reddit-jokes dataset and augmented into instruction form using keybert.

mikegarts:
> 1. There are edited jokes, where the author speaks with the auditory. They are usually have "Edit:". We need to filter out these jokes or remove the part that starts with "Edit:".
> 2. There are jokes where selftext is "[deleted]". They also should be filtered out.
> 3. Column "**index_level_0**" need to be removed from the parquet file. You can do it by passing `index=False` to `to_parquet` method
> 4. If you can you may add more then 10000 rows, because we will later filter datasets with our safety models
> 5. Please add some information into HF readme

Hi @Vechtomov, and thanks for reviewing :)
1. Done
2. Done
3. Done
4. It's 20000 entries now 
5. I've added a brief explanation about the DS to the HF readme.

Vechtomov (approved):

PR #2211, opened 2023-03-25
Fix horizontal scrolling on mobile

AbdBarho:
Problem:

Screenshot_20230325-183147_Bromite


notmd:
I also notice the bottom margin of the slim footer is not showing anymore. Could you check it?
image

notmd (approved):

PR #2212, opened 2023-03-25
fixes text client to work with new debug login workflow

yk:


AbdBarho (approved):
Sorry for breaking it!

PR #2221, opened 2023-03-25
Add Google OAuth support to inference backend

### LINKED ISSUE #1790: Add additional authentication methods to generate JWT on inference server
As #1772 is merged and users exist in the inference server database and can be authenticated via Discord OAuth2, we need to add email authentication. We should also discuss other authentication providers (e.g. GitHub)

Individual additional providers can have their own issues made to split this up further

olliestanley:
This will round out our auth options alongside Discord and GitHub. I think these should be sufficient for us to close #1790.


olliestanley:
I haven't tested this locally yet as I need to setup a Google OAuth app but marking ready for review for any code comments

AbdBarho:
> I haven't tested this locally yet as I need to setup a Google OAuth app but marking ready for review for any code comments

is this ready now?

olliestanley:
> is this ready now?

I think so. I think it could be cleaner but I won't have time to make changes for a couple of days so I will address it as part of #2286 with the other auth code

AbdBarho (approved):
The google api seems a bit weird to me. "People"?

olliestanley:
> The google api seems a bit weird to me. "People"?

Yeah it's odd. Maybe comes from the Google+ days. They also make the OAuth process a lot less clear than other providers!

Issue #2192, opened 2023-03-24
RM filtering SFT replies


sanagno:
Now that we have a good RM, we should use it to filter predictions from the SFT model.

Process:
- Get samples from the SFT model.
- Filter them based on the RM model.
- Compare the quality of the filtered vs unfiltered replies.

shahules786:
@sanagno I'll look into this.

sanagno:
thanks a lot! Basically what you have been doing already

shahules786:
Should this be one pipeline? @sanagno

sanagno:
Does not have to be, but it is up to you. Whatever works in the end and is expected to be more user-friendly in terms of use but also in terms of evaluating the results in the end.

PR #2225, opened 2023-03-26
SFT Rejection Sampling using RM

shahules786:
### Changes
- added functionality to do rejection sampling, saves rejected samples and selected samples to separate files
- Calculates mean, min, and max reward for each set for comparison
- Updated the README

fixes: #2192


andreaskoepf (approved):
very nice! :-) Thank you!

PR #2226, opened 2023-03-26
Style updates to chat UI

AbdBarho:
* limit width
* move chat date to the end

image

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -240,8 +240,8 @@ type PendingMessageEntryProps = {
 };
 
 const PendingMessageEntry = ({ content, isAssistant, children }: PendingMessageEntryProps) => {
-  const bgAssistant = useColorModeValue("gray.100", "gray.800");
-  const bgUser = useColorModeValue("#f1f8ff", "#42536B");
+  const bgUser = useColorModeValue("gray.100", "gray.700");
+  const bgAssistant = useColorModeValue("#DFE8F1", "#42536B");
   const { data: session } = useSession();
   const image = session?.user?.image;
 
@@ -254,10 +254,6 @@ const PendingMessageEntry = ({ content, isAssistant, children }: PendingMessageE
     <BaseMessageEntry
       avatarProps={avatarProps}
       bg={isAssistant ? bgAssistant : bgUser}
-      stackProps={{
-        alignSelf: isAssistant ? "start" : "end",
-        flexDirection: isAssistant ? "row" : "row-reverse",
-      }}
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       content={content!}
     >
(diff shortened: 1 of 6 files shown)


notmd (approved):
I'm not sure the new UI is better. IMO it's more ugly than the current (I'm sorry). And changing the position of the message base on role is quite weird compared to other chat UI. Most chat UI out there align the message to the left and hight light the different base on color and avatar (user messages have the same color as the background and only highlight the assistant's message, cause most of users want to read the assistant's instead of their own message). And about the chat list, I'm planning to have a secondary sidebar (something like this https://minimals.cc/dashboard/chat) when I have time. So the user can quickly switch between chats. The main sidebar will be collapsed. IMO, OpenAI UI is really good, just missing a few features for better UX. But anyway, it's up to you.

AbdBarho:
> 'm not sure the new UI is better. IMO it's more ugly than the current (I'm sorry). And changing the position of the message base on role is quite weird compared to other chat UI. Most chat UI out there align the message to the left and hight light the different base on color and avatar (user messages have the same color as the background and only highlight the assistant's message, cause most of users want to read the assistant's instead of their own message). And about the chat list, I'm planning to have a secondary sidebar (something like this https://minimals.cc/dashboard/chat) when I have time. So the user can quickly switch between chats. The main sidebar will be collapsed. IMO, OpenAI UI is really good, just missing a few features for better UX. But anyway, it's up to you.


Thank you for your feedback! no need to apologize, it is the reason why we create pull requests is to get feedback.

I can revert all the stuff to the chat conversation, I was also not 100% convinced, but for the dashboard it fixed a bug where the text would overflow past the container, which got me to fix the whole width thingy. Unless you are currently working on branch where this would create many conflicts?

What do you say? revert chat, keep dashboard?

notmd:
> > 'm not sure the new UI is better. IMO it's more ugly than the current (I'm sorry). And changing the position of the message base on role is quite weird compared to other chat UI. Most chat UI out there align the message to the left and hight light the different base on color and avatar (user messages have the same color as the background and only highlight the assistant's message, cause most of users want to read the assistant's instead of their own message). And about the chat list, I'm planning to have a secondary sidebar (something like this https://minimals.cc/dashboard/chat) when I have time. So the user can quickly switch between chats. The main sidebar will be collapsed. IMO, OpenAI UI is really good, just missing a few features for better UX. But anyway, it's up to you.
> 
> Thank you for your feedback! no need to apologize, it is the reason why we create pull requests is to get feedback.
> 
> I can revert all the stuff to the chat conversation, I was also not 100% convinced, but for the dashboard it fixed a bug where the text would overflow past the container, which got me to fix the whole width thingy. Unless you are currently working on branch where this would create many conflicts?
> 
> What do you say? revert chat, keep dashboard?

Yeah let's keep the dashboard and revert the chat for now. I haven't started the work on new UI yet (will do when I have time)

PR #2227, opened 2023-03-26
update deps

notmd:


website/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
notmd, line 1: This is required to update TS to 5.0.

AbdBarho: I unfortunately had bad experiences with this parameter  in the past, package conflicts without end.

Would it be okay if we stick with ts 4 for now?

notmd: Done!

website/package.json
```diff
@@ -58,30 +58,30 @@
...
-    "react-markdown": "^8.0.5",
+    "react-markdown": "^8.0.6",
     "react-syntax-highlighter": "^15.5.0",
-    "react-textarea-autosize": "^8.4.0",
+    "react-textarea-autosize": "^8.4.1",
     "remark-gfm": "^3.0.1",
     "sharp": "^0.31.3",
     "storybook-addon-next-router": "^4.0.2",
-    "swr": "^2.0.3",
+    "swr": "2.0.3",
```
notmd, line 79: There is some typing issue with swr, so I pin it

AbdBarho (approved):

PR #2228, opened 2023-03-26
Various improvements to the dev setup

yk:


inference/server/oasst_inference_server/settings.py
```diff
@@ -68,7 +68,7 @@ def debug_api_keys_list(self) -> list[str]:
     # we decided on letting the nextjs / website backend handle the token at first
     # and then proxy this information back to the inference server
     # in short: this should refer to the website, not to this server
-    auth_callback_root: str = "https://open-assistant.io/api/inference_auth"
+    auth_callback_root: str = "http://localhost:3000/api/inference_auth"
```
AbdBarho, line 71: Just for info: this variable is already set in the docker compose, which is what I use for development.

https://github.com/LAION-AI/Open-Assistant/blob/c0f3bafb4cbdbbad5711d1adb4e997317ee0b1d3/docker-compose.yaml#L227

AbdBarho (approved):

PR #2230, opened 2023-03-26
Sending MessageRead along with error to client

yk:


AbdBarho:
Just for my understanding: when an error happens, we receive the generated message so far, or do we receive a "message" that contains the error?

yk:
> Just for my understanding: when an error happens, we receive the generated message so far, or do we receive a "message" that contains the error?

You'll get an `InferenceMessage` object with the state being some error state, essentially the same as if you were to call GET on the message endpoint. In the future, I can also put the partially-generated content into the message, for now "content" is null.

AbdBarho (approved):

PR #2231, opened 2023-03-26
Export script: Fix duplicate loading of models

andreaskoepf:
Fix merge artifact: Models were loaded twice during export after the llama merge.


dvruette (approved):
lgtm

PR #2232, opened 2023-03-26
Enabling Threads and Retry for Web Chat

yk:
- separates prompter and assistant messages
- enables finding the current thread in the conversation tree in the web frontend
- adds a retry button on failed messages


AbdBarho:
> separates prompter and assistant messages

I am curious as to which problem is solved by this?

yk:
> I am curious as to which problem is solved by this?

- can re-try on error
- can suggest to re-sample if the user votes thumbs-down
- can sample multiple times and let the user compare

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -129,11 +185,13 @@ export const ChatConversation = ({ chatId }: ChatConversationProps) => {
           score={message.score}
           state={message.state}
           onVote={handleOnVote}
+          onRetry={() => handleOnRetry(message.parent_id!)}
```
notmd, line 188: can you add a `useCallback` here?

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -48,53 +124,33 @@ export const ChatConversation = ({ chatId }: ChatConversationProps) => {
...
-      });
-    }
-    if (message) {
-      setMessages((messages) => [...messages, message]);
-    }
-    setQueueInfo(null);
-    setResponse(null);
-    setIsSending.off();
-  }, [chatId, getFormValues, messages, setIsSending]);
+    initiate_assistant_message(prompter_message.id);
```
notmd, line 141:
```suggestion
   await initiate_assistant_message(prompter_message.id);
```

notmd (approved):
Nice feature. But I wonder why we have to do 2 request to just send and get a message. Why we don't stream the message immediately when user post a prompter message? But anyway, I'm fine with it.

andreaskoepf (approved):

PR #2233, opened 2023-03-26
Add migrations to web db

AbdBarho:
We fucked up real bad here:

We started using `prisma db push` to sync the prisma schema with the db, with the intention of moving to migrations "soon":
https://github.com/LAION-AI/Open-Assistant/blob/000fb0641bbb0fe9f599bf2373ec57c2bd404275/docker/Dockerfile.prisma#L1-L10

But we never did this (because we never needed it), now I wanted to add a new table to the database and realized that the infrastructure for something like this is not in place.

Here I created an initial migration from the already available db scheme, to be able to deploy this without losing any data, we would need to force prisma to mark this migration as applied (even though it is not)

The command is as follows:
```
npx prisma migrate resolve --applied 20230326131923_initial_migration
```

So this means, we have to ssh into the production web container and run this command to be have the ability to apply future migrations.

All the changes to the schema.prisma file are auto-formatting by the prisma vscode extension


notmd:
What happens when you add a new table and run `prisma db push`? I thought it just work, although we cant rollback the table.

AbdBarho:
> What happens when you add a new table and run `prisma db push`? I thought it just work,  although we cant rollback the table.

it works if all you do is add new tables or columns, but it is not safe, prisma will push everything and drop tables if necessary. i.e. data loss.

notmd (approved):

PR #2234, opened 2023-03-26
Update web container to include migrations

AbdBarho:
Follow up to #2233 

* update website container to include migrations
* remove outdated prisma container
* update other containers for better layer caching.

This PR contains no behavioural changes.

More context:

We started using `prisma db push` to sync our web db prisma schema with the db, with the intention of moving to migrations soon™, but we never did this because we never needed it.

Now I wanted to add a new table to the database and realized that the infrastructure for something like this is not in place.

I created an initial migration from the already available db scheme, to be able to deploy this without losing any data, we would need to force prisma to mark this migration as applied.

The command is as follows:
```
npx prisma migrate resolve --applied 20230326131923_initial_migration
```

This PR prepare the web docker container to have migrations as a first class citizen.


notmd (approved):

yk (approved):
thank you!

PR #2239, opened 2023-03-26
Updated nginx config for prod, including streaming headers

yk:
also some deployment playbook stuff


AbdBarho:
Is this already deployed? the streaming works on dev.

deploy/dev-node/nginx/nginx.conf
```diff
@@ -28,6 +28,12 @@ http {
 		    proxy_set_header Host $host;
 		    proxy_set_header X-Real-IP $remote_addr;
 		    proxy_pass http://127.0.0.1:3000;
+
+			proxy_set_header Connection '';
+			proxy_http_version 1.1;
```
andreaskoepf, line 33: was the different indentation made intentionally?

yk: this is somehow the github diff. locally it looks fine

yk:
> Is this already deployed? the streaming works on dev.

yes I deployed this on dev. it's a manual process for the nginx

notmd:
Can we hold this? Maybe streaming does not work because we didn't explicit add the header to the response. See https://github.com/vercel/next.js/issues/9965

yk:
> Can we hold this? Maybe streaming does not work because we didn't explicit add the header to the response. See [vercel/next.js#9965](https://github.com/vercel/next.js/issues/9965)

yes I found that too. But streaming on dev works, I've tried with exactly the settings in this PR. so that fixes it (at least for us)

deploy/dev-node/nginx/nginx.conf
yk: it was a spaces and tabs issue. should be fixed now

notmd:
But does this PR disable caching for the hold website?

yk:
> But does this PR disable caching to entire the website?

Probably. But for now I'd take the hit until we figure out a better solution. The caching is not really our performance bottleneck as our content is in large part dynamically produced and we don't have many media files etc to serve

deploy/prod-node/nginx/nginx.conf
```diff
@@ -68,9 +80,37 @@ http {
...
+			proxy_set_header X-Real-IP $remote_addr;
+
+			proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+			proxy_set_header X-Forwarded-Proto $scheme;
+			proxy_set_header Upgrade $http_upgrade;
+			proxy_set_header Connection $connection_upgrade;
+			proxy_redirect off;
+			proxy_buffering off;
+
+			proxy_pass http://127.0.0.1:8285;
```
melvinebenezer, line 113: I normally used something like this instead of hard coding the IP address. Helped in load balancing and scaling

 - docker will automatically resolve this to the correct address
  because we use the same name as the service: e.g "inference-server" in the docker compose
    
`
upstream inference_servers {
   
    server inference-server:8285;
}
`

`
 proxy_pass inference_servers;
`

melvinebenezer (changes requested):

deploy/prod-node/nginx/nginx.conf
yk: doesn't work because nginx is running in docker-compose while the rest of the services are deployed via ansible

melvinebenezer: oh ok, understood.

melvinebenezer (approved):

AbdBarho (approved):
fingers crossed

andreaskoepf (approved):

PR #2240, opened 2023-03-26
Fix GPTNeoX-20B training

dvruette:
Notable config settings that are required to get GPTNeoX-20B to behave nicely:
- bf16: when training in fp16, gradients often over/underflow, leading to bad convergence. (bf16 could also potentially improve the loss of other models, we should try)
- More warmup steps: To fill the optimizer buffers, a gentle warmup is required. Right now, we do this by just using more warmup steps, but in the future a linear LR warmup schedule could be used to achieve the same effect.
- Gradient checkpointing is faster than gradient accumulation. This could also translate to other models.
- Stage 3 is required to fit the 20B model in bf16. For some reason, fp16 was possible to fit with stage 2 but bf16 wasn't.

Other learnings:
- Residual dropout quickly degrades performance of the pre-trained model. Even `p=0.1` leads to an initial loss of > 5.
- Flash attention does not go too well with the 20B model. There are slight numerical differences between flash attention and the GPTNeoX attention implementation that accumulate layer-by-layer, ultimately leading to vastly different results.
  - Run in bf16 with regular implementation: https://wandb.ai/open-assistant/supervised-finetuning/runs/x2pczqa9
  - Run in bf16 with flash attention: https://wandb.ai/open-assistant/supervised-finetuning/runs/cvv3edm8
  - Run in fp32 with regular implementation: https://wandb.ai/open-assistant/supervised-finetuning/runs/shrzz3xp

EDIT: Updated fp32 comparison run. It actually behaves nicely just like bf16, so the most likely explanation for flash attention not working is accumulating errors.


sanagno:
Are all the requirements necessary?

If I run pipreqs I get:

```
bitsandbytes==0.36.0.post2
datasets==2.8.0
deepspeed==0.7.7
evaluate==0.4.0
fastlangid==1.0.11
flash_attn==0.2.8
gdown==4.7.1
ipython==8.11.0
model_training==1.0.0
numpy==1.24.2
pytest==7.2.2
PyYAML==6.0
scikit_learn==1.2.2
sentencepiece==0.1.97
tokenizers==0.13.2
torch==1.13.1
tqdm==4.65.0
transformers==4.26.1
trlx==0.3.0
```

dvruette:
Maybe not, I think @andreaskoepf added them at some point, maybe I wasn't supposed to commit them.

andreaskoepf:
> Maybe not, I think @andreaskoepf added them at some point, maybe I wasn't supposed to commit them.

If it was me who changed requirements.txt it was not intentionally ... IMO we should only add the requirements that are really necessary.

dvruette:
Ok, I'll adopt the requirements by @sanagno.

andreaskoepf:
> Maybe not, I think @andreaskoepf added them at some point, maybe I wasn't supposed to commit them.

If it was me who changed requirements.txt it was not intentionally ... IMO we should only add the requirements that are really necessary.

model/model_training/tools/export.py
```diff
@@ -0,0 +1,62 @@
...
+    parser.add_argument("--dtype", type=str, default="float16", help="float16, float32")
+    parser.add_argument("--hf_repo_name", type=str, help="Huggingface repository name (used with --push)")
+    parser.add_argument("--auth_token", type=str, help="")
+    parser.add_argument("--output_folder", type=str, help="output folder path (used with)")
+    parser.add_argument("--max_shard_size", type=str, default="10GB")
+    parser.add_argument("--cache_dir", type=str)
+    return parser.parse_args()
+
+
+def main():
```
andreaskoepf, line 20: This file can be deleted. The actual export tool is called export_model.py in the same directory ;-).

model/model_training/configs/config.yaml
```diff
@@ -223,16 +206,39 @@ pythia-6.9B:
   per_device_train_batch_size: 4
   per_device_eval_batch_size: 4
 
-pythia-12B:
-  learning_rate: 6e-6
+pythia-12b:
```
andreaskoepf, line 209: best would probably be to leave the pythia-12B config unchanged...

model/model_training/requirements.txt
```diff
@@ -0,0 +1,19 @@
+bitsandbytes==0.36.0.post2
+datasets==2.8.0
+deepspeed==0.7.7
+evaluate==0.4.0
+fastlangid==1.0.11
+flash_attn==0.2.8
+gdown==4.7.1
+ipython==8.11.0
```
andreaskoepf, line 8: hm. normally the requirements should reside in `pyproject.toml` which is now in the `model/` directory (parent of model_training). The file was moved allow to make model_training a "real" python module that can be imported everywhere.

model/model_training/tools/export.py
dvruette: Not sure why this was still here...

model/model_training/configs/config.yaml
dvruette: i'm pretty sure the unchanged config is rather out of date, but this probably isn't the place to update it

model/model_training/requirements.txt
dvruette: moved dependencies to pyproject.toml and deleted requirements.txt

model/pyproject.toml
```diff
@@ -15,18 +15,27 @@ dependencies = [
     "datasets==2.8.0",
     "deepspeed==0.7.7",
```
andreaskoepf, line 16: I am currently training with newer version of deepspeed everywhere since new HF transformers version complaine abut old deepspeed version .. IMO we can safely update this to latest version.

andreaskoepf (approved):
looks, good .. left comment regarding deepspeed version .. but could also be changed in a separate PR.

model/model_training/models/patching.py
```diff
@@ -127,6 +127,9 @@ def patch_model(
     if resid_pdrop is not None and (resid_pdrop < 0 or resid_pdrop > 1.0):
         raise ValueError("Invalid argument: `resid_pdrop` must be between 0.0 and 1.0")
 
+    if not flash_attention and (resid_pdrop is None or resid_pdrop == 0.0):
```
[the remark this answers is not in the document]
dvruette: I personally like the explicitness, but can change if preferred

PR #2241, opened 2023-03-26
Store inference info in DB

AbdBarho:
Blocked by #2234

Move credentials from cookies to DB, we already do a db request, so not much additional latency is added.


notmd:
Why we have to store the credentials in DB?

AbdBarho:
> Why we have to store the credentials in DB?

When I tried saving everything in cookies, I was basically trying to re-implement the `next-auth` workflows manually, and I realized that this is redundant work, especially keeping the state in sync, and we have to handle edge case like logout from app equals logout from inference and stuff.

Furthermore, creds in DB is more consistent with how next-auth stores credentials now, and it makes it much easier to add or remove any information we want to the JWT or session.

notmd:
I believe we are making our infrastructure more complicated with this. We can implement a simple client-side only authentication system. And remove the proxy call through nodejs completely. We also having problem with streaming when proxy though node.

notmd:
I believe we are making our infrastructure more complicated with this. We can implement a simple client-side only authentication system. And remove the proxy call through nodejs completely. We also having problem with streaming when proxy though node.

AbdBarho:
> I believe we are making our infrastructure more complicated with this.

Agree, it is a bit more complicated than before.

> We can implement a simple client-side only authentication system. And remove the proxy call through nodejs completely. We also having problem with streaming when proxy though node.

it would have to happen entirely cross origin, so maybe it is not that simple, and unfortunately, the bulk of the work would have to be done in the inference server. So it is less complex for us, it is more complex for inference.

Maybe we can find some middle ground, where we don't proxy everything through node, but we still manage the complexity there.

Getting some opinions on this from the others would be great.

notmd:
> it would have to happen entirely cross origin, so maybe it is not that simple, and unfortunately, the bulk of the work would have to be done in the inference server. So it is less complex for us, it is more complex for inference.

I will try to implement a POC for client-side authentication.

AbdBarho:
> I will try to implement a POC for client-side authentication.

Great, tell me if I can help.

AbdBarho:
Superseded by #2278

PR #2244, opened 2023-03-27
Expose env vars globally

notmd:
A small downside is that we have to use getServerSideProps on every page.

### DIFF
--- a/website/src/pages/_app.tsx
+++ b/website/src/pages/_app.tsx
@@ -39,9 +39,14 @@ function MyApp({ Component, pageProps: { session, ...pageProps }, env, cookie }:
 
   // expose env vars on the client
   if (typeof window !== "undefined") {
-    for (const key in env) {
-      process.env[key] = env[key];
-    }
+    process.env = new Proxy(env, {
+      get(env, key: string) {
+        if (!(key in env)) {
+          console.warn(`Environment variable ${key} not set in _app.tsx`);
+        }
+        return env[key];
+      },
+    });
   }
 
   return (
@@ -68,6 +73,7 @@ MyApp.getInitialProps = ({ ctx: { req } }: AppContext) => {
       ENABLE_EMAIL_SIGNIN: boolean(process.env.ENABLE_EMAIL_SIGNIN),
       ENABLE_EMAIL_SIGNIN_CAPTCHA: boolean(process.env.ENABLE_EMAIL_SIGNIN_CAPTCHA),
       CLOUDFLARE_CAPTCHA_SITE_KEY: boolean(process.env.CLOUDFLARE_CAPTCHA_SITE_KEY),
+      NODE_ENV: process.env.NODE_ENV,
     },
     cookie: req?.headers.cookie,
   };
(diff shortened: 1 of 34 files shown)


website/src/pages/_app.tsx
```diff
@@ -1,43 +1,56 @@
...
   const direction = i18n.dir();
   useEffect(() => {
     document.body.dir = direction;
   }, [direction]);
+
+  // expose env vars on the client
+  if (typeof window !== "undefined") {
+    for (const key in env) {
+      process.env[key] = env[key];
+    }
```
AbdBarho, line 44: could we maybe do it like this?
```ts
process.env = new Proxy(env, {
    get(env, key) {
        const value = env[key];
        if (!value) {
           throw new Error("Environment variable not set in _app.tsx")
        }
        return value;
    }
)
```

I think it is very easy to make this mistake, which is why we should log how to fix it..

I hope nothing else breaks, otherwise just `Object.assign(process.env, env)`

website/src/pages/_app.tsx
```diff
@@ -46,5 +59,18 @@ function MyApp({ Component, pageProps: { session, cookies, ...pageProps } }: App
...
+MyApp.getInitialProps = ({ ctx: { req } }: AppContext) => {
+  return {
+    env: {
+      INFERENCE_SERVER_HOST: process.env.INFERENCE_SERVER_HOST,
+      ENABLE_CHAT: isChatEnabled(),
+      ENABLE_EMAIL_SIGNIN: boolean(process.env.ENABLE_EMAIL_SIGNIN),
+      ENABLE_EMAIL_SIGNIN_CAPTCHA: boolean(process.env.ENABLE_EMAIL_SIGNIN_CAPTCHA),
+      CLOUDFLARE_CAPTCHA_SITE_KEY: boolean(process.env.CLOUDFLARE_CAPTCHA_SITE_KEY),
+    },
+    cookie: req?.headers.cookie,
```
AbdBarho, line 72: can you check if this is still needed? I remember that I did some debugging, and this cookie field was always empty.

AbdBarho (commented):
Do you have a rough estimation of how much more additional computation required for the server-side rendering? we don't want CPU usage to sky rocket for some env variables.

notmd:
> Do you have a rough estimation of how much more additional computation required for the server-side rendering? we don't want CPU usage to sky rocket for some env variables.

Haven't any benchmarks yet. But I'm running a NextJS app (quite small) that do full SSR, it doesn't use so much CPU. NextJs is really lightweight, SSR only adds extra few milliseconds to the response latency. Let me try to switch to page do not use env var to static first.

website/src/pages/_app.tsx
notmd: The req object only available when we use getServerSideProps. So it still works if the page is used with getServerSideProps.

notmd: So now we are using mixed getStaticProps and getServerSideProps, it only partially works. I think that's fine

notmd:
I managed to use getStaticProps. In my previous implementation, my concern is when the user navigates from static page to ssr page the process.env is not available. But look like next/router will handle all of this (it's still call the server to get the props.). 

I also add a warning when we try to access env that is not available (I can't throw an error because the markdown lib also uses env for some special rendering). But this still has some caveats, if we try to access env var "statically" eg: `const a = process.env.SOME_THING` on a top-level scope of the module, env var won't be able at that time, and the warning is not triggered too. We have to use env var carefully on the client side.

AbdBarho (approved):

PR #2245, opened 2023-03-27
T-Few baseline

mikichan007:


mikichan007:
I have a problem with deepspeed

I have patched llama7b (IA3), so there are very few trainable parameters.

Now I run experiment localy (just for testing) with 2 Nvidia GeForce 10GB and 12GB with stage 3. I think that with half precision and offload this model (with gradients, etc.) must fit in those 2 gpus (right?).

However, I catch cuda memory exceed error, and I can’t figure out why

andreaskoepf:
Closing this since development in this direction seems to have been discontinued.

PR #2248, opened 2023-03-27
Workaround for Apple silicon hosts

k-nearest-neighbor:
`psutils.cpu_freq()` doesn't currently support M1 / M2 chips. There is an open issue in psutils for this: https://github.com/giampaolo/psutil/issues/1892

`psutils.cpu_freq()` actually raises an exception when running on the Apple silicon host, and returns `None` when running in a container hosted on Apple silicon. I didn't add the exception handling since our case is the container case.

### DIFF
--- a/oasst-shared/oasst_shared/schemas/inference.py
+++ b/oasst-shared/oasst_shared/schemas/inference.py
@@ -41,8 +41,11 @@ class WorkerHardwareInfo(pydantic.BaseModel):
         data["uname_processor"] = platform.uname().processor
         data["cpu_count_physical"] = psutil.cpu_count(logical=False)
         data["cpu_count_logical"] = psutil.cpu_count(logical=True)
-        data["cpu_freq_max"] = psutil.cpu_freq().max
-        data["cpu_freq_min"] = psutil.cpu_freq().min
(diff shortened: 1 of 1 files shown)


k-nearest-neighbor:
fixing the syntax, also looking at an alternative

oasst-shared/oasst_shared/schemas/inference.py
```diff
@@ -41,8 +41,11 @@ def __init__(self, **data):
         data["uname_processor"] = platform.uname().processor
         data["cpu_count_physical"] = psutil.cpu_count(logical=False)
         data["cpu_count_logical"] = psutil.cpu_count(logical=True)
-        data["cpu_freq_max"] = psutil.cpu_freq().max
-        data["cpu_freq_min"] = psutil.cpu_freq().min
+        # psutil.cpu_freq()is None for conatainers running on an Apple silicon host
+        # https://github.com/giampaolo/psutil/issues/1892
+        cpu_freq = psutil.cpu_freq()
```
melvinebenezer, line 46: Until 1892 is fixed
if platform.machine() == 'arm64' can be used to update cpu_freq with 0 or None

melvinebenezer (changes requested):

Issue #2122, opened 2023-03-19
Correlation metrics during reward model training


dvruette:
As suggested by [ZickZack on Discord](https://discord.com/channels/1055935572465700980/1059076134807023656/1086275001809911900), the Spearman and/or Kendall correlation would be a good metric for evaluating our reward models during training, in addition to validation accuracy and loss.

Scipy provides implementations for both of these (here, Kendall correlation):
```python
from scipy import stats
a = [1,2,3]
b = [2,1,3]
stats.kendalltau(a,b)
# returns KendalltauResult(correlation=0.33333333333333337, pvalue=1.0)
```

shahules786:
I can take this up @dvruette .  For RM evaluation I assume that each batch contains rankings for answers to a single prompt.

dvruette:
Thank you! One batch can actually contain multiple rankings. The batch has a key `cu_lengts` which indicates the borders between samples. You can check the `compute_loss` function or the `RMLoss` class for examples.

shahules786:
cool. I'll do that @dvruette

PR #2251, opened 2023-03-28
Correlation metrics for Reward Model

shahules786:
Added support for tracking multiple metrics while RM training.
Supported metrics
- Kendalls tau
- Spearman corr
- Accuracy

An extra argument `metrics` should be specified in config_rm.yaml

fixes : #2122


theblackcat102 (approved):
LGTM

andreaskoepf (approved):
ok, I will directly give it a try...

PR #2254, opened 2023-03-28
Add rng_seed parameter to trainers

andreaskoepf:
Add `rng_seed` command line/config parameter. Command line arg will overwrite yaml-config value.


andreaskoepf:
I tested sft with two independent runs on galactica-125m-base-finetuned .. both loss-curves were nearly identical:

image

Only problem I see is the random state manipulation in `PerDatasetSampler` [utils.py#L92](https://github.com/LAION-AI/Open-Assistant/blob/9528a4c62c5a1bb0689dda58948ea9185de4aaea/model/model_training/utils.py#L92).

PR #2259, opened 2023-03-28
Migrate all inference api call to client side

notmd:
Migrate all api calls of inference to the client side (no proxy through NodeJS anymore). I haven't verified the full UX of chat yet. Will do when I update the chat layout.


website/src/types/Chat.ts
```diff
@@ -12,10 +12,10 @@ export interface ChatItem {
   id: string;
   created_at: string; //timestamp
   modified_at: string; //timestamp
+  messages: InferenceMessage[];
```
notmd, line 15: Inference backend always returns an array even when there is no message yet.

website/src/hooks/chat/useInferenceAuth.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const useInferenceAuth = () => {
+  const { data, isLoading, mutate } = useSWR<boolean>("/auth/check", () => {
+    return new OasstInferenceClient().check_auth();
+  });
+
+  return {
+    isAuthenticated: data,
+    isLoading,
+    mutate,
```
AbdBarho, line 12: could you please rename to something that reflects the function? like `refreshAuth` or similar?

website/src/pages/chat/[id].tsx
```diff
@@ -45,19 +44,14 @@ const Chat = ({ id, modelInfos }: ChatProps) => {
...
+export const getServerSideProps: GetServerSideProps<ChatProps, { id: string }> = async ({ locale = "en", params }) => {
   if (!isChatEnabled()) {
     return {
       notFound: true,
     };
   }
 
-  const token = await getToken({ req });
-  const client = createInferenceClient(token);
+  const client = new OasstInferenceClient();
```
Comment (AbdBarho) [inline website/src/pages/chat/[id].tsx:54]: Since we do all everything on the client, we should probably move this to the client at some point

AbdBarho (commented):
I will have to look over this again

website/src/pages/chat/[id].tsx
Comment (notmd) [reply website/src/pages/chat/[id].tsx:54]: For API the doesn't require auth, I want to do it in server side. It must more performance and the logic on the client less complicated (loading/error state).

Issue #2257, opened 2023-03-28
Add loader for high-quality OIG datasets


andreaskoepf:
OIG Dataset: https://huggingface.co/datasets/laion/OIG/tree/main

PR #2260, opened 2023-03-28
Add simple OIG data loader

andreaskoepf:
Add loader class to directly fetch individual OIG jsonl files from the web.

resolves: #2257


sanagno (approved):
looks good

PR #2263, opened 2023-03-29
Use LLaMA impl of Huggingface Transformers

andreaskoepf:


model/model_training/configs/config.yaml
```diff
@@ -113,41 +113,50 @@ oasst_export_latin_cyrillic:
     - oasst_export:
         lang: "bg,ca,cs,da,de,en,es,fr,hr,hu,it,nl,pl,pt,ro,ru,sl,sr,sv,uk"
         input_file_path: 2023-03-27_oasst_research_ready_synth.jsonl.gz
+    - alpaca
+    - oig_file:
+        source_url: https://huggingface.co/datasets/laion/OIG/resolve/main/unified_chip2.jsonl
+        max_count: 10000
+        min_length: 1000
+        val_split: 0.2
```
dvruette, line 121: for 10k samples, val_split=0.1 is probably enough

model/model_training/configs/config.yaml
```diff
@@ -98,7 +98,7 @@ oasst_export_eu:
     - alpaca
     - oig_file:
         source_url: https://huggingface.co/datasets/laion/OIG/resolve/main/unified_chip2.jsonl
-        max_count: 30000
+        max_count: 10000
         min_length: 1000
         val_split: 0.2
```
dvruette, line 103: Same here, val_split 0.1 is probably enough

dvruette (approved):
LGTM

model/model_training/configs/config.yaml
andreaskoepf: yes definitely, now working on the effective values in other pr.

model/model_training/configs/config.yaml
andreaskoepf: When pre-training works I will probably remove this here .. the GPT4all dataset re-generade all chip answers with GPT3.5 .. we'll use them.

PR #2268, opened 2023-03-30
typo in parsing openai/summarize_from_feedback

mikegarts:


mikegarts:
The field that contains the text is called 'article'
image

'posts' is always None

andreaskoepf (approved):
good catch!

andreaskoepf (commented):
Just looked at my old own code and I had special handling for post vs. article there, e.g. see: https://github.com/andreaskoepf/laion_sum_data/blob/d2c1338f1406be473b458638eb1a432fc74419af/reward_model/train.py#L66-L74

I don't remember exactly but also the fact that the example of the OAI [learning to summarize readme](https://github.com/openai/summarize-from-feedback/blob/master/README.md) shows an example with 'post' looks as if both "post" and "article" are used .. depending on the source, e.g. reddit post or news article ...
 "post" and other times "

mikegarts:
Actually it's more interesting than that. 
At the axis subset - the long content is in 'article' for test split, and under 'post' for validation.
Taking the first field that contains data.

andreaskoepf (approved):
Thanks a lot!

PR #2277, opened 2023-03-31
Flash attention support for Llama

dvruette:
Adds support for patching flash attention into Llama models. This is done by replacing the `forward` method of `LlamaAttention` and using `flash_attention` for the attention computation part.

### DIFF
--- a/model/model_training/models/patching.py
+++ b/model/model_training/models/patching.py
@@ -6,10 +6,11 @@ from typing import Callable, Optional
 
 import torch
 import torch.nn as nn
-import torch.nn.functional as F
-from torch.nn.utils.rnn import pad_sequence
+import transformers
 from transformers import GPTNeoXForCausalLM, GPTNeoXModel, LlamaForCausalLM, LlamaModel
 
+from .patching_llama import llama_forward_with_flash_attn
+from .patching_utils import compute_flash_attention
 from .reward_model import GPTNeoXRewardModel
 
 SUPPORTED_MODELS = [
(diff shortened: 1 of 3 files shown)


model/model_training/configs/config.yaml
```diff
@@ -149,6 +162,7 @@ oasst_export_eu:
   use_custom_sampler: false
```
andreaskoepf, line 162: this configuration is a bit strange (I know .. you didn't touch it) but I don't think it was ever used but for one test run .. e.g. it consists mainly of gpt4all with alone is >400k samples ... if it should be used for finetuning it would probably good to balance the datasets with the custom-sampler ...

model/model_training/configs/zero3_config.json
```diff
@@ -12,12 +20,12 @@
     }
   },
   "scheduler": {
-    "type": "WarmupDecayLR",
+    "type": "WarmupLR",
```
andreaskoepf, line 23: Can we create different zero config files for WarmupLR and WarmupDecayLR? For the last runs I changed it manually but I think it would be better to have pre-training (which would refer the WarmupLR zero config) + fine-tuning (using WarmupDecalyLR zero config) .. this will make it simpler/possible to reproduce our results.

model/model_training/tests/test_patched_gpt_neox.py
```diff
@@ -65,9 +65,9 @@ def test_resid_dropout_patch():
 
         x = model.get_input_embeddings()(input_ids)
         for layer in model.layers:
-            y1 = layer.attention(x, None)[0]
-            y2 = layer.attention(x, None)[0]
-            assert (y1 - y2).abs().sum() > 1e-5, "attention output is the same for different forward passes"
+            # y1 = layer.attention(x, None)[0]
+            # y2 = layer.attention(x, None)[0]
+            # assert (y1 - y2).abs().sum() > 1e-5, "attention output is the same for different forward passes"
```
andreaskoepf, line 70: why was this commented out?

model/model_training/tests/test_patched_llama.py
```diff
@@ -0,0 +1,43 @@
+import torch
+from model_training.models.patching import patch_model
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+
+def test_flash_attention_patch(dtype=torch.float16, device="cuda:0"):
+    tokenizer = AutoTokenizer.from_pretrained("/home/ubuntu/llama_hf/7B")
+    tokenizer.add_special_tokens({"pad_token": "</s>", "eos_token": "</s>", "sep_token": "<s>"})
```
andreaskoepf, line 8: necessary?

model/model_training/trainer_sft.py
```diff
@@ -408,11 +409,12 @@ def main():
     if training_conf.log_wandb and (not training_conf.deepspeed or training_conf.local_rank == 0):
         import wandb
 
+        wandb_name = training_conf.model_name.replace(os.getenv("HOME", "/home/ubuntu"), "")
```
andreaskoepf, line 412: :+1:

andreaskoepf (approved):
Most of my comments were regarding config files. Actual flash attention + test looks good!

model/model_training/configs/config.yaml
```diff
@@ -118,6 +121,15 @@ pretrain:
     - explain_prosocial:
         fraction: 0.05
 
+oasst_only:
```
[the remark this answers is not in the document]
dvruette: i'd actually propose to make `oasst_only` the `oasst_only_latin_cyrillic` config. see followup PR later.

model/model_training/configs/config.yaml
dvruette: i don't know how exactly the custom sampler works, so i'm not going to touch it for now..

model/model_training/tests/test_patched_llama.py
```diff
@@ -0,0 +1,43 @@
...
+
+        batch = tokenizer(["hello world", "lorem ipsum dolor sit amet"], padding=True, return_tensors="pt").to(device)
+        out1 = model(**batch).logits
+        out2 = patched_model(**batch).logits
+
+        diff = (out1 - out2) * batch["attention_mask"].unsqueeze(-1)
+        assert (diff.abs() < 0.1).all()
+
+    input_ids = torch.randint(0, model.config.vocab_size, size=(2, 10), device=device)
+    patched_model(input_ids).logits.mean().backward()
```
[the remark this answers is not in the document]
dvruette: No, haven't compared gradients.. Indeed, I just don't want an error to occur during the backward pass. I'm willing to trust that if the forward pass gives correct numbers, the backward pass does as well.

PR #2279, opened 2023-03-31
Added max size to work queue and an error response if full when enqueuing

yk:


olliestanley (approved):

andreaskoepf (approved):
IMO much better than long busy-anim without error after timeout!

PR #2280, opened 2023-03-31
Remove assigning eos token id (llama compatibility)

andreaskoepf:
- removed setting the eos id property because this property is read-only at the lama tokenizer (and it is not necessary to explicitly set it for pythia models)


dvruette (approved):
LGTM

PR #2292, opened 2023-04-02
Move create chat button to the top

### LINKED ISSUE #2281: Sort list of chats by date
descending, from latest to oldest

AbdBarho:
I moved the "create" button to the top, on long lists it is annoying to scroll down. We can of course change the design later, as long as the functionality is ready.

image


notmd (commented):
We better do this in the inference backend instead.

AbdBarho:
> We better do this in the inference backend instead.

Agree 😅, I was too lazy to look up how to do this. I changed this PR to be just about moving the button to the top.

notmd (approved):

PR #2293, opened 2023-04-02
changed basic hf server to support quantization and streaming

yk:
I now also put the multi-worker-image PR in here because it builds heavily upon it


inference/worker/basic_hf_server.py
```diff
@@ -42,6 +47,51 @@ def terminate_server(signum, frame):
...
 model: transformers.PreTrainedModel
 tokenizer: transformers.PreTrainedTokenizer
+fully_loaded: bool = False
+model_input_queue: Queue = Queue()
+
+
+def model_thread():
+    request: interface.GenerateStreamRequest
+    output_queue: Queue
+    eos_token = ""
```
olliestanley, line 57: Preference really but I feel a one-liner is clearer here

`eos_token = tokenizer.eos_token if hasattr(tokenizer, "eos_token") else ""`

inference/worker/settings.py
```diff
@@ -17,5 +17,8 @@ class Settings(pydantic.BaseSettings):
     perform_oom_test: bool = False
     oom_test_max_length: int | None = None
 
+    # for hf basic server
+    quantize: bool = False
```
olliestanley, line 21: Should we have a more descriptive setting name here so people don't expect it to have an effect when not using the basic server?

olliestanley (approved):
Looks good. I wonder if we should refactor the worker directory structure a bit to have a subdirectory for basic HF server functionality, the code (and the difference between basic HF server vs text-generation-inference server) is possibly a bit confusing for new developers at this point.

Also seems like something weird has happened with the diffs from main.

inference/worker/settings.py
yk: this needs to be called quantize because the hf-inference server also expects it to be called like this

model/model_training/trainer_rm.py
```diff
@@ -184,6 +184,8 @@ def main():
 
     init_rng(training_conf)
 
+    init_rng(training_conf)
+
```
dvruette, line 188: line is duplicate of previous line
```suggestion
```

inference/worker/__main__.py
```diff
@@ -31,10 +31,11 @@ def main():
 
     if model_config.is_lorem:
         tokenizer = None
-    elif model_config.is_llama:
-        tokenizer: transformers.PreTrainedTokenizer = transformers.LlamaTokenizer.from_pretrained(model_config.model_id)
+    # elif model_config.is_llama:
+    #     tokenizer: transformers.PreTrainedTokenizer = transformers.LlamaTokenizer.from_pretrained(model_config.model_id)
```
dvruette, line 35: can be removed with latest HF version (4.28.0.dev0)

dvruette (approved):

olliestanley (approved):
Reapproving with new changes (apologies for the early review yesterday, somehow didn't notice it was a draft!)

Issue #2174, opened 2023-03-22
Create dataset configuration for reward model training


andreaskoepf:
Currently the reward model is only trained on open-assistant data. Further datasets should be incorporated.

- create configuration to using import-code for HF datasets e.g. Oasst, AnthropicRLHF, HFSummary ..
- verify correct batch formatting v2_5
- check/document sample counts of the different datasets
- are there important other HF datasets that could be easily added?
- test training with small debug RM model

andreaskoepf:
> @andreaskoepf - does anybody working on this one?

No, are you interested in working on it? We currently train a 2nd batch of SFT models and would like to run RL fine-tuning on the results soon.

PR #2294, opened 2023-04-02
add dataset counts script

CloseChoice:
This can be used for a part of #2174.

This PR contains:
 - a generic script to count the dataset entries for any dataset and any mode
 - updated model/model_training/README.md with some counts of datasets
 - some minor adjustments to model/model_training/utils.py so that the datasets can be returned as list. This is done so that this PR can reuse existing code and does not need to copy+paste.


model/model_training/configs/config.yaml
```diff
@@ -29,7 +29,7 @@ defaults:
     - cnn_dailymail
     - multi_news
     - scitldr
-    - soda
+    # - soda  # TODO: fix list index out of range error
```
dvruette, line 32: still to do?

dvruette (approved):
LGTM, thanks!

model/model_training/configs/config.yaml
CloseChoice: that's still an open todo, just checked it. Will create a seperate PR to fix this.

model/model_training/utils.py
```diff
@@ -365,6 +367,8 @@ def get_dataset(conf, mode="sft"):
         if val is not None:
             evals[dataset_name] = Subset(val, list(range(min(len(val), conf.eval_size)))) if conf.eval_size else val
 
+    if return_datasets_as_list:
```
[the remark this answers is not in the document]
CloseChoice: Thanks for pointing me towards this. Just removed the `return_datasets_as_list` part ;)

PR #2299, opened 2023-04-02
Fix DeepSpeed 0.8.3 training

### LINKED ISSUE #2298: SFT-Training with DeepSpeed 0.8.3 produces significantly worse results than with 0.7.7
While comparing to previous results we noticed significantly worse training curves (higher loss, lower eval accuracy) for pythia-12b runs. We tracked this down to a newer version of DeepSpeed. We recently updated to DeepSpeed 0.8.3 because the git head version of HF transformers (llama support etc.) prints warnings when an old DeepSpeed version is detected.

We tried `trainer_sft.py` latest version (7ee3ac077a708047c3c3d1d1531c86e436be0f03) on two 8x A100 machines and noticed that downgrading to DeepSpeed 0.7.7 again produces the earlier loss/eval trajectories.

Zero configuration used:

```
{
  "fp16": {
    "enabled": "auto",
    "loss_scale": 0,
    "loss_scale_window": 1000,
    "initial_scale_power": 16,
    "hysteresis": 2,
    "min_loss_scale": 1
  },
  "optimizer": {
    "type": "AdamW",
    "params": {
      "lr": "auto",
      "betas": "auto",
      "eps": "auto",
      "weight_decay": "auto"
    }
  },
  "scheduler": {
    "type": "WarmupDecayLR",
    "params": {
      "warmup_min_lr": "auto",
      "warmup_max_lr": "auto",
      "warmup_num_steps": "auto",
      "total_num_steps": "auto"
    }
  },
  "zero_optimization": {
    "stage": 2,
    "allgather_partitions": true,
    "allgather_bucket_size": 2e8,
    "overlap_comm": true,
    "reduce_scatter": true,
    "reduce_bucket_size": 2e8,
    "contiguous_gradients": true,
    "offload_optimizer": {
      "device": "cpu",
      "pin_memory": true
    }
  },
  "gradient_accumulation_steps": "auto",
  "gradient_clipping": "auto",
  "steps_per_print": 2000,
  "train_batch_size": "auto",
  "train_micro_batch_size_per_gpu": "auto",
  "wall_clock_breakdown": false
}
```

dvruette:
Fixes #2298

### DIFF
--- a/model/model_training/configs/config.yaml
+++ b/model/model_training/configs/config.yaml
@@ -117,6 +117,16 @@ pretrain:
     - explain_prosocial:
         fraction: 0.05
 
+reference-data:
+  datasets:
+    - oasst_export:
+        lang: "en,es,de,fr"
+        input_file_path: 2023-03-27_oasst_research_ready_synth.jsonl.gz
+        val_split: 0.05
+    - alpaca
+  sort_by_length: false
+  use_custom_sampler: false
+
 oasst_export_eu:
   datasets:
     - oasst_export:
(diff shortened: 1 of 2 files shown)


model/model_training/configs/config.yaml
```diff
@@ -265,12 +275,17 @@ pythia-12B:
   learning_rate: 6e-6
   model_name: EleutherAI/pythia-12b-deduped
   weight_decay: 0.0
-  max_length: 2048
-  warmup_steps: 20
+  max_length: 620
```
olliestanley, line 278: Is this right? I thought we have been doing SFT runs with 2048?

dvruette: yes, you're right! was confused by a previous config

model/model_training/configs/config.yaml
```diff
@@ -265,12 +275,17 @@ pythia-12B:
   learning_rate: 6e-6
   model_name: EleutherAI/pythia-12b-deduped
   weight_decay: 0.0
-  max_length: 2048
```
andreaskoepf, line 268: I still think this should be `max_length: 2048`

andreaskoepf (approved):
ok, lgtm!

PR #2301, opened 2023-04-03
two more datasets

mikegarts:


model/model_training/custom_datasets/extra_rm_datasets.py
```diff
@@ -0,0 +1,26 @@
...
+def load_anthroopic_rlhf():
+    train = AnthropicRLHF(split="train", mode="rm")
+    validation = AnthropicRLHF(split="test", mode="rm")
+
+    print(f"AnthropicRLHF extra dataset size: {len(train)}, {len(validation)}")
+
+    return train, validation
+
+
+def load_open_ai_summarize_from_feedback():
```
mikegarts, line 14: Maybe it's better to just make the split manually here and in load_open_ai_summarize_from_feedback ? 
I'm not sure what's best, but the default split seems to be not balanced in terms of sizes...

model/reward/instructor/rank_datasets.py
```diff
@@ -261,11 +261,16 @@ def preprocess_dialogue(self, text):
...
 
-    def __init__(self, split="train", sep_token="<sep>") -> None:
+    def __init__(self, split="train", sep_token="<sep>", mode=None) -> None:
         super().__init__()
+        self.mode = mode
         assert split in ("train", "test")
         if sep_token is None:
             sep_token = " . "
+
+        if mode == "rm":
```
andreaskoepf, line 271: This are changes in the reward/instructor folder. They are not used for the trainer_rm code. I suggest to remove these changes.

andreaskoepf (approved):
thank!

PR #2308, opened 2023-04-03
Set validation max length to a different value.

jordiclive:
Option to set the validation max_length to be different from the training max_length.

As the validation max length influences accuracy/loss results. 
It would be good to keep it fixed at say, 2048, and vary the train max length. This might also mean configuring the per_device_eval_batch_size to be lower/running less frequent validation.

### DIFF
--- a/model/model_training/configs/config.yaml
+++ b/model/model_training/configs/config.yaml
@@ -13,6 +13,7 @@ defaults:
   eval_steps: 200
   save_steps: 1000
   max_length: 512
+  val_max_length:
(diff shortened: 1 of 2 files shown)


dvruette (approved):
Very nice! looks good, thx!

andreaskoepf (approved):
With `val_max_length` None by default this should be ok. When setting it to different values than max_len it will no longer be possible to directly compare the validation loss/accuracy values with earlier runs of otherwise identical configurations.

PR #2314, opened 2023-04-03
Reward Model evaluation

shahules786:
Changes:
- Added script to evaluate RM  with datasets like Hellaswag, Standford SHP, etc 

Results are saved in below format,
```
{
    "model": "OpenAssistant/reward-model-deberta-v3-large",
    "dataset": "Hellaswag",
    "split": [
        "train"
    ],
    "accuracy": "0.928"
}
```


andreaskoepf (approved):
thanks for fixing the dataset loading.

PR #2317, opened 2023-04-03
Show model config in the chat UI

AbdBarho:
So that user knows which config is shown, also moved the config button next to the submit button.

The PR is more about functionality and less about design. We can change the design entirely in the future.

image

Also, I updated charka UI, since VSCode was having many problems with it.

### DIFF
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -4,10 +4,16 @@
   "config_title": "Chat configuration",
   "empty": "(empty)",
   "login_message": "To use this feature, you need to login again. Login using one of these providers:",
+  "max_new_tokens": "Max new tokens",
   "model": "Model",
   "preset": "Preset",
   "preset_custom": "Custom",
   "queue_info": "Your message is queued, you are at position {{ queuePosition, number, integer }} in the queue.",
+  "repetition_penalty": "Repetition penalty",
+  "temperature": "Temperature",
+  "top_k": "Top K",
+  "top_p": "Top P",
+  "typical_p": "Typical P",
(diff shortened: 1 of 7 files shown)


website/public/locales/en/chat.json
```diff
@@ -4,10 +4,16 @@
...
+  "max_new_tokens": "Max new tokens",
   "model": "Model",
   "preset": "Preset",
   "preset_custom": "Custom",
   "queue_info": "Your message is queued, you are at position {{ queuePosition, number, integer }} in the queue.",
+  "repetition_penalty": "Repetition penalty",
+  "temperature": "Temperature",
+  "top_k": "Top K",
+  "top_p": "Top P",
+  "typical_p": "Typical P",
```
notmd, line 16: I'm not sure we want to translate this. For example, translating this to Vietnamese have completely no meaning at all

notmd (approved):

notmd:
> Also, I updated charka UI, since VSCode was having many problems with it.

Just curious, what happen with charka and vscode?

AbdBarho:
> Just curious, what happen with charka and vscode?

https://github.com/chakra-ui/chakra-ui/issues/7526

Basically, VSCode internal typescript version got updated to 5.x, which was not compatible was chakra UI, and would show errors everywhere.

website/public/locales/en/chat.json
AbdBarho: These are very technical parameters, and even if we could translate them, that does not make them understandable, maybe this could give a short summary:

https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.GenerationConfig.typical_p

I think that's ok like this for the first prototype.

PR #2318, opened 2023-04-03
Refactor inference backend auth and switch to authlib

### LINKED ISSUE #2286: Cleanup inference backend auth
We have added `authlib` dependency for Google's OAuth. I want to switch GitHub and Discord to also use authlib, which will allow us to remove a bunch of custom code. We can also remove some duplicated code as part of this.

olliestanley:
Close #2286. Backend auth got very messy in going from one to three providers so this refactors a bunch of the common code and switches everything to authlib to strip out custom code. This also fixes a couple of bugs.

Adding session middleware means authlib can track state for us. This will also require adding a session middleware secret key variable to our environments.

### DIFF
--- a/inference/server/main.py
+++ b/inference/server/main.py
@@ -11,6 +11,7 @@ from oasst_inference_server.routes import admin, auth, chats, configs, workers
 from oasst_inference_server.settings import settings
 from oasst_shared.schemas import inference
 from prometheus_fastapi_instrumentator import Instrumentator
+from starlette.middleware.sessions import SessionMiddleware
 
 app = fastapi.FastAPI()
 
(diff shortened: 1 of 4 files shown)


yk:
does this work with the new pass-through auth mode we have where the backend just "tells" us who the user is?

olliestanley:
> does this work with the new pass-through auth mode we have where the backend just "tells" us who the user is?

I believe so. These changes just simplify (outsourcing much of the work to authlib) how the `login` and `callback` routes for each provider work. The `trusted` route should continue to function identically. I am not exactly sure how the pass-through auth works though so let me double check to confirm before we merge this.

AbdBarho:
> does this work with the new pass-through auth mode we have where the backend just "tells" us who the user is?

I tested it locally and it works

AbdBarho (approved):

Issue #2265, opened 2023-03-29
Augment Training set with model data


sanagno:
Currently, the quality of our data is almost "too" good. To help with the RL training, our RM should also have seen "bad" examples. One idea would be to augment our current training set with replies that we expect to be worse than average.

Task:
- Find "bad" SFT models, preferably multiple of multiple bad quality, e.g. ~100m, 1b pythia models.
- Use them to generate replies to our data. You can see a preview here https://github.com/Open-Assistant/oasst-model-eval/blob/main/model_eval/manual/data/en_100_tree.jsonl. These replies should be labelled as having lower rank as the current data.
- Train an RM based on these data as well.

theblackcat102:
@sanagno  I have finish sampling augment results and here's the reward scores on the first en_100.jsonl score.

```
Prompt : How can I find the best 401k plan for my needs?
Human Rank (ordered by rank):
1. There are many factors to consider when selecting a retirement savings plan, such as tax advantages, 0.95556640625
2. The first step is to determine your current financial situation and goals. You should consider how m, 0.490234375
3. The first step in finding a good retirement savings plan is to determine your current financial situ, 0.794921875
4. There are many factors to consider when choosing a qualified 401(k) plan, such as your income level, 0.72021484375
5. If you're looking to invest money in a retirement savings account, there are many factors to conside, -0.3828125
-----
chip20b synthetic answers (ordered by rank)
1. The first step is to research your options. You should look at different plans and compare them base -0.1843261
2. The first step is to determine your financial goals. Once you have a clear idea of what you are .. -0.007488250
3. The best way to find a 401k plan that meets your needs is to do research and compare different optio -0.085998
```

andreaskoepf:
The synthetic data generated by @theblackcat102  was used in training our models. We might consider releasing the synthetic data-set.

PR #2321, opened 2023-04-04
fix deepspeed issue on trainer_rm.py, add crossentropy support

theblackcat102:
# How to use augmented results [#2265](https://github.com/LAION-AI/Open-Assistant/issues/2265)

1. Generate bad responses

* sample using code under tools/augment_oasst.py, please modify the model_name for different models for sample different bad samples. Note the sampling would took a few days to finish.

* Once all different samples are generated, use the AggregateResults class to merge multiple bad response into one dataset and move it under the cache_dir

2. Use the augmented results as replacement for oasst dataset

```
oasst_export_augmented:
  datasets:
    - augment_oasst:
        input_file_path: augmented_latin_cyrillic_oasst_2023-03-27.jsonl
```

# Other contribution of this branch

I was training some baseline for RM models and found some issues running different models (deberta, electra, bloom) under deepspeed

* fix summarized from human feedback (hf_summary) and webgpt dataset for rm, sft and rl mode. Note the implementation of hfsummary under rank_datasets.py is slightly wrong.

* deepspeed distributed gradient issue fixed in trainer_rm.py

* add crossentropy loss support for new code `RMCLSLoss`, mainly because WebGPT seems to use this instead of rankloss and part of the old code migration. (see wandb reward-model search for pythia-410m )

* added pretrained model support for non pythia model for RM training. ( see utils.py get_model )

* and I was having gradient unscale error by setting dtype=fp16 on model initialization as mentioned by dimitri in discord ([link](https://discuss.pytorch.org/t/valueerror-attemting-to-unscale-fp16-gradients/81372)). 

@andreaskoepf I also wanted to discuss about the way how trainer code (trainer_sft.py trainer_rm.py) are executed now. Cause its placed under `model/model_training/` and import modules from under `model_training`. So the problem is how do you execute the code?

1. prepend system path? and run `deepspeed trainer_sft.py ...` under model_training?

2. `python -m model_training.trainer_rm ...` works but doesn't support deepspeed

3. move trainer code under `model`

I am raising this issue mainly because there isn't any instruction on how to execute the trainer in under the model_training/readme


model/model_training/trainer_rm.py
```diff
@@ -158,7 +160,7 @@ def argument_parsing(notebook=False, notebook_args=None):
     conf["wandb_entity"] = args.wandb_entity
     conf["local_rank"] = args.local_rank
     conf["deepspeed"] = args.deepspeed
-    if args.rng_seed is not None:
+    if "rng_seed" not in conf and args.rng_seed is not None:
```
andreaskoepf, line 163: I don't think this logic is correct, but I can correct it after merge in the improve_small_rm branch..

model/model_training/utils.py
```diff
@@ -296,12 +299,14 @@ def get_model(conf, tokenizer, pad_vocab_size_to_multiple_of=16):
 
     if conf.is_reward_model:
         if "pythia" in conf.model_name:
-            model = GPTNeoXRewardModel.from_pretrained(conf.model_name, cache_dir=conf.cache_dir, torch_dtype=dtype)
```
andreaskoepf, line 299: hm. why was the dtype specification here removed .. ?

andreaskoepf (approved):
ok, very cool! thank you.

model/model_training/trainer_rm.py
theblackcat102: its my bad, I have reverted back

model/model_training/utils.py
theblackcat102: mainly due to unable to rescale fp16, but as you mentioned, we could just assign fp32 for that

PR #2319, opened 2023-04-03
Switched ENABLE_CHAT to env to have it client-side exposed

yk:

PR #2323, opened 2023-04-04
Populate env vars before render layout

notmd:
close #2315 close #2319

### DIFF
--- a/website/src/components/Layout.tsx
+++ b/website/src/components/Layout.tsx
@@ -57,12 +57,6 @@ export const getDashboardLayout = (page: React.ReactElement) => (
             pathname: "/stats",
             icon: TrendingUp,
           },
-          {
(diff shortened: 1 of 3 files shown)


AbdBarho:
> What's wrong with that? I tested with production mode on my machine and everything still works.

It works, it is just not consistent

https://github.com/LAION-AI/Open-Assistant/blob/main/website/src/lib/chat_enabled.ts

On the server it would run this:
```js
 process.env.NODE_ENV === "development" || process.env.ENABLE_CHAT === "true"
```
on the client it would resolve to this
```js
process.env.NODE_ENV === "development" || process.env.NODE_ENV === "development" || process.env.ENABLE_CHAT === "true"
```

AbdBarho (approved):

PR #2324, opened 2023-04-04
Use fixed RNG seed value for all DeepSpeed workers

andreaskoepf:
- remove pre-worker-different seed value initialization


dvruette (approved):
Good catch! this could actually have an impact...

PR #2325, opened 2023-04-04
Update configs according to feedback to #2277

dvruette:
Implements feedback to #2277 that was left unadressed.


andreaskoepf (approved):
thnx! lgtm.

PR #2326, opened 2023-04-04
fix list index out of range error for soda dataset

CloseChoice:
- fixed "index out of range" error in soda dataset
- simplified the logic in the dataset preprocessing
- added some plausibility checks
@theblackcat102 I am pinging you here directly, since you created it. Would be good to get your feedback on this.


model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -211,48 +212,57 @@ def __getitem__(self, index):
...
+            pairs = self._partition_list(data["dialogue"], 2)
+            # add dialog background to first pair.
+            # [[Q1, A1], [Q2, A2]] -> [[B + Q1, A1], [Q2, A2]]
+            pairs[0][0] = f"{dialogue_bg} {pairs[0][0]}"
+            # add previous history to conversation pairs
+            # e.g. [[B + Q1, A1], [Q2, A2]] -> [[B + Q1, A1], [B + Q1, A1, Q2, A2]]
+            pairs = list(accumulate(pairs, lambda x, y: x + y))
+            return pairs
 
     def __init__(self, cache_dir, input_max_length=1024) -> None:
```
andreaskoepf, line 258: Is `input_max_length` used?

CloseChoice: nope, just in [commented out code](https://github.com/LAION-AI/Open-Assistant/pull/2326/files#diff-c2caae63856734e93f51b5cbcb1695605d53a69538b8ea0d45471e2c94f95ee9L257). But I guess @theblackcat102 commented that out

theblackcat102: Yeah, we could, but the problem is how to pass in the parameters?

theblackcat102 (commented):
LGTM

model/model_training/custom_datasets/qa_datasets.py
CloseChoice: Added the option to pass the parameter right from the config. Using `**kwargs` should IMO be a general pattern that we could use more often for our datasets

andreaskoepf (approved):
Nice, thanks a lot for adapting to latest sft & rl loader format!

PR #2327, opened 2023-04-04
inference: use uuid v7 for most of table

notmd:
use this lib https://github.com/stevesimmons/uuid7

### DIFF
--- a/inference/server/oasst_inference_server/models/worker.py
+++ b/inference/server/oasst_inference_server/models/worker.py
@@ -8,6 +8,7 @@ from loguru import logger
 from oasst_inference_server.settings import settings
 from oasst_shared.schemas import inference
 from sqlmodel import Field, Relationship, SQLModel
+from uuid_extensions import uuid7str
 
 
 class WorkerEventType(str, enum.Enum):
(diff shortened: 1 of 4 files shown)


inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -17,7 +17,7 @@ class Config:
     async def get_chats(self) -> list[models.DbChat]:
         query = sqlmodel.select(models.DbChat)
         query = query.where(models.DbChat.user_id == self.user_id)
-        query = query.order_by(models.DbChat.created_at.desc())
+        query = query.order_by(models.DbChat.id.desc())
```
AbdBarho, line 20: could we leave this sorted by date? this has the advantage of being functional regardless of which id type we use. Also perhaps slightly faster but not that much.

inference/server/oasst_inference_server/models/worker.py
```diff
@@ -17,7 +18,7 @@ class WorkerEventType(str, enum.Enum):
 class DbWorkerComplianceCheck(SQLModel, table=True):
     __tablename__ = "worker_compliance_check"
 
-    id: str = Field(default_factory=lambda: str(uuid4()), primary_key=True)
+    id: str = Field(default_factory=lambda: uuid7str(), primary_key=True)
```
andreaskoepf, line 21: is the lambda necessary or could it be `default_factory=uuid7str` ?

notmd: Thanks, I'm not really familiar with python syntax.

inference/server/oasst_inference_server/user_chat_repository.py
notmd: Done!

olliestanley (approved):

PR #2334, opened 2023-04-05
Use SSR for every pages

notmd:
~~It's ready now. `_document` only runs once serverside, so this is the expected behavior we want. `getInitialProps` run on both server and client. On the client it populates the env wrong, it uses static env from `.env` instead of actual env, I'm not sure why.~~

Using SSR is the only way to make it work. This PR is quite similar to #2343 but I expose it to a global var for flexibility (accessing the env outside of react scope) and more performance.


AbdBarho:
> On the client it populates the env wrong, it uses static env from `.env` instead of actual env, I'm not sure why.

Because the props are created at build time, where the `ENABLE_CHAT` variable is not available. If we change `getStaticProps` to `getServerSideProps` on `/dashboard` then the chat icon is visible because it forces a render in runtime, where the variable is available.

website/src/pages/_document.tsx
```diff
@@ -1,10 +1,27 @@
+import { boolean } from "boolean";
 import { Head, Html, Main, NextScript } from "next/document";
 
 export default function Document() {
   return (
     <Html className="h-full antialiased" lang="en">
       <Head>
         <link rel="shortcut icon" type="image/png" href="/images/logos/favicon.png" />
+        <script
+          dangerouslySetInnerHTML={{
```
AbdBarho, line 10: is there really no other way?

website/src/styles/Theme/index.ts
```diff
@@ -47,7 +47,7 @@ const styles: Styles = {
     },
     body: {
       position: "relative",
-      bg: props.colorMode === "light" ? "gray.50" : colors.dark.bg,
```
notmd, line 50: I fix a small UI issue where the color the not match on light mode.

website/.env
```diff
@@ -24,3 +24,4 @@ ENABLE_EMAIL_SIGNIN=true
 # inference config
 INFERENCE_SERVER_HOST="http://localhost:8000"
 INFERENCE_SERVER_API_KEY="6969"
+ENABLE_CHAT=true
```
AbdBarho, line 27: it should be disabled by default, otherwise the flag should be `DISABLE_CHAT`

website/src/pages/api/chat/assistant_message.ts
```diff
@@ -1,8 +1,12 @@
+import { boolean } from "boolean";
 import { withoutRole } from "src/lib/auth";
 import { createInferenceClient } from "src/lib/oasst_inference_client";
 import { InferencePostAssistantMessageParams } from "src/types/Chat";
 
 const handler = withoutRole("banned", async (req, res, token) => {
+  if (!boolean(process.env.ENABLE_CHAT)) {
```
AbdBarho, line 7: this line is repeated too many times, maybe we can re-add the `isChatEnabled` but we can use the new `getEnv()` in it

website/types/env.d.ts
```diff
@@ -9,6 +9,7 @@ declare global {
       ADMIN_USERS: string;
       MODERATOR_USERS: string;
       INFERENCE_SERVER_HOST: string;
+      ENABLE_CHAT: boolean;
```
AbdBarho, line 12: this should be string, after all, are environment variables are strings.

AbdBarho (commented):
I am disappointed that next js does not allow for an easier solution for handling this.

maybe they expect everyone to have `/api/config` or similar anyway.

arguably speaking, that would still be the easiest solution.

website/.env
notmd: This env only used in local or static page (which we don't use any more). So I think it's fine

AbdBarho: Well, then who ever is deploying the project should know, that if they just delete the variable, chat will remain enabled, they have to explicitly set the value to false for it to be disabled.

website/types/env.d.ts
AbdBarho: Then we need to remove all the `boolean(...)` from our code because it is redundant?

https://github.com/LAION-AI/Open-Assistant/blob/c67e1962904c895d6f9a0e7ba8c430aebe39623c/website/src/pages/api/chat/message.ts#L6

But that would not work, right?

website/.env
notmd: I think those behaviors should be documented instead.

website/src/pages/api/chat/assistant_message.ts
notmd: Done, I dont still dont like the inChatEnable solution, but it's simple for now. The perfect solution should be using some data transformer lib like `zod` or something

website/.env
AbdBarho: Ok, please don't forget to document it somewhere!

AbdBarho (approved):

Issue #261, opened 2023-01-02
Creating augmented data using few-shot prompts for explanations of jokes, logical inferences, etc.


ontocord:
See https://www.lesswrong.com/posts/EHbJ69JDs4suovpLw/testing-palm-prompts-on-gpt3.

Try doing 2, 3 or 4 shot inference  on something like JT or neox 20B or galactica. 

After we find a promising model and configuration, we can scrape the net for jokes and paragraphs with logical inferences to create dialog data.

Human: Tell me a joke about {extract keywords from joke}
Assistant: {joke}
Human: Explain the joke.
Assisant: {explanation}

See also https://storage.googleapis.com/pathways-language-model/PaLM-paper.pdf

smytjf11:
Going with the movie reviews idea, could we use the Rotten Tomatoes dataset to generate prompts, maybe supplement with one of the models fine tuned on it as well?

https://huggingface.co/datasets/rotten_tomatoes

momegas:
The idea is to create a dataset with explanations. Like for example take the movie dataset and do this:
`This is a movie review for the movie {movie}: {review}. This movie review is {classifciaiton} because ...[your created answer]`
Am I right?
I'm interested in picking this up. How large should the dataset be?

ontocord:
@momegas yes. if it is very compute intensive, it doesn't need to be large. maybe see if you can get it to work first. And then we can discuss size. we can run it on some extra compute.

mikegarts:
Sound like a very cool task and I would love to give it a try if it is still relevant :)

mikegarts:
@kkie02 Sure, I'm in discord as mikegarts. Feel free to ping me.
Btw I just opened a pr with somewhat relevant instruction dataset https://github.com/LAION-AI/Open-Assistant/pull/2209 but would love to cooperate on further work.

0x22almostEvil:
Going to work in this field, but with more specific tasks (semantics, logic, reasoning)
https://github.com/LAION-AI/Open-Assistant/issues/3122

andreaskoepf:
Closing old data issue.

PR #2337, opened 2023-04-05
add Logic Inference Dataset

kkie02:
Regarding https://github.com/LAION-AI/Open-Assistant/issues/261.

This is an re-produce of the dataset from LogicInference Dataset in paper: https://openreview.net/pdf?id=HAGeIS_Lcg9. I think it will helpful for improving logic inference ability of the model.

The github page of LogicInference Dataset: https://github.com/google-research/google-research/tree/master/logic_inference_dataset.

This dataset is aimed to offer more dataset for Open Assistant project, depending on their demands, there three columns: INSTRUCTION, RESPONSE, SOURCE.

The results in this dataset is a little different from which was introduced in the original paper:

1.For all three splits (IID/OOD/length), only IID is used. In the original paper, it seems that model can reach better performance with data generated by this split method.

2.In the original paper, there are two form of responses: LOGICINFERENCEb (with the answer at the beginning) and LOGICINFERENCEe (with the answer at the end). This dataset uses LOGICINFERENCEe, that means: for all questions, the model will first do logic inference, and give the final answer at the end.

3.The original paper, some parameters in generate_dataset.py are:

N_INFERENCE_PROBLEMS = 5000

N_VARIATIONS = 25

N_EXAMPLES = 200000

TRAIN_RATIO = 0.9

LENGTH_SPLIT_THRESHOLD = 4

RANDOM_SEED = 0

I choose some new parameters:

N_INFERENCE_PROBLEMS = 10000

N_VARIATIONS = 25

N_EXAMPLES = 55000

TRAIN_RATIO = 1

LENGTH_SPLIT_THRESHOLD = 4

RANDOM_SEED = 1111

The original script generated 4814 different inference problems and extended all those inference problems to around 200,000 Q-A pairs. My settings generated 5491 different inference problems and extended them to around 54,607 Instruction-Response pairs. I think for Open Assistant projects, maybe the number of different inference problems is more important, and generated many similar Instruction-Response pairs will only add training time and doesn't make much sense.


olliestanley:
Hi! Before we can review this we will need you to delete the `.idea/` so that the PR no longer tries to add it to the repository

kkie02:
> Hi! Before we can review this we will need you to delete the `.idea/` directory so that the PR no longer tries to add it to the repository

Hi, I think that you can check this now if you have time. I just bought a new laptop and uses Windows wsl there, so I met many problems but now I fixed all of them. On the other hand, I forked the files of https://github.com/google-research/google-research/tree/master/logic_inference_dataset at the beginning, but their code doesn't follow the flake8 format, so I only keep the generate_dataset.py now and make it fitting the flake8 format. I did some changes in this file to make it suitable for our OA project.

By the way, should I still need to delete .idea/ file now? I am not sure.

olliestanley:
> > Hi! Before we can review this we will need you to delete the `.idea/` directory so that the PR no longer tries to add it to the repository
> 
> Hi, I think that you can check this now if you have time. I just bought a new laptop and uses Windows wsl there, so I met many problems but now I fixed all of them. On the other hand, I forked the files of https://github.com/google-research/google-research/tree/master/logic_inference_dataset at the beginning, but their code doesn't follow the flake8 format, so I only keep the generate_dataset.py now and make it fitting the flake8 format. I did some changes in this file to make it suitable for our OA project.
> 
> By the way, should I still need to delete .idea/ file now? I am not sure.

Thanks. Yes we would still need the `.idea/` directory to be removed, as it's related to your local IDE setup so we should avoid having it in the repository

.idea/.gitignore
andreaskoepf, line 1: Would it be possible to remove the .idea folder from this PR please?

andreaskoepf:
I removed the .idea folder & resolved the conflict...

andreaskoepf (approved):
Interesting dataset, curious if it has a noticeable impact on logical reasoning capabilities.

olliestanley (approved):

PR #2342, opened 2023-04-06
Add alpaca reverse augmentation possibility

### LINKED ISSUE #2335: Implement inversion samples for alpaca.
The idea is from this [tweet](https://twitter.com/abacaj/status/1643416835843293186). I'm not sure if they include the inversion samples in the validation set, I think we should obviously exclude them from the validation. 

But is a cool idea nevertheless. I can take a look at this later this week.

CloseChoice:
closes #2335

- Add possiblity to reverse question and response for alpaca dataset
- add debug configs to show how to used

Alpaca with reverse augmentation can be run with:
```bash
python trainer_sft.py --configs pythia-70m-deduped
```
I couldn't run this due to unscale gradient errors, so before this gets merged we should do a run to check whether this reversal really improves the loss.


CloseChoice:
Eval dataset is still missing

CloseChoice:
One thing that is not quite clear to me is if we should the reverse augmentation samples ADDITIONALLY to the train set

andreaskoepf:
> One thing that is not quite clear to me is if we should the reverse augmentation samples ADDITIONALLY to the train set

Yes, I think forward + reverse could both be part of the training set. Ideally the same examples of the alpaca set should be used for evaluation of both forward/reverse (in order to avoid leaking of eval-data of the other dataset into the train set).

andreaskoepf:
> I couldn't run this due to unscale gradient errors, so before this gets merged we should do a run to check whether this reversal really improves the loss.

For the smaller models it is necessary to specify `dtype: float32` in the configuration to avoid this error.

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -419,37 +420,78 @@ def __getitem__(self, index):
         return self.pairs[index]
 
 
-class AlpacaBase(Dataset):
-    def __init__(self, dataset_name: str, mode: str, cache_dir: str = None) -> None:
+class AlpacaBaseDataset(Dataset):
+    def __init__(self, data: list, mode: str):
         super().__init__()
+        self.data = data
+        if mode not in ["sft", "rl"]:
```
andreaskoepf, line 427: could be tuple instead of list

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -419,37 +420,78 @@ def __getitem__(self, index):
...
         if self.mode == "sft":
             return (question, answer)
         elif self.mode == "rl":
             return (question,)
 
 
-class Alpaca(AlpacaBase):
-    def __init__(self, mode: str = "sft", cache_dir: str = None) -> None:
-        super().__init__(dataset_name="yahma/alpaca-cleaned", mode=mode, cache_dir=cache_dir)
+class AlpacaDataset(AlpacaBaseDataset):
```
andreaskoepf, line 444: If these classes don't add any functionality they could be removed  (i.e. base class be used directly).

model/model_training/custom_datasets/__init__.py
```diff
@@ -114,10 +113,8 @@ def get_one_dataset(
         dataset = DiveMT()
     elif dataset_name == "webgpt":
         dataset = WebGPT(mode=mode)
-    elif dataset_name == "alpaca":
-        dataset = Alpaca(mode=mode, cache_dir=data_path)
-    elif dataset_name == "code_alpaca":
-        dataset = CodeAlpaca(mode=mode, cache_dir=data_path)
+    elif dataset_name in ["alpaca", "code_alpaca"]:
```
andreaskoepf, line 116: use immutable tuple instead of list

andreaskoepf (changes requested):
Very nice, thank you. IMO the now 'blood-empty' AlpacaDataset & CodeAlpacaDataset could be removed.

model/model_training/custom_datasets/qa_datasets.py

model/model_training/custom_datasets/qa_datasets.py

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -419,37 +420,78 @@ def __getitem__(self, index):
...
 
+def load_alpaca_dataset(
+    dataset_name: str,
+    val_split: float,
+    cache_dir: str,
+    mode: str = "sft",
+    manual_seed: int = 287631038922,
+    reverse_augmentation: bool = False,
+    keep_unreversed: bool = True,
+) -> tuple[AlpacaDataset, AlpacaDataset] | tuple[CodeAlpacaDataset, CodeAlpacaDataset]:
```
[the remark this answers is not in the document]

model/model_training/custom_datasets/__init__.py
CloseChoice: done

model/model_training/configs/config.yaml
```diff
@@ -267,6 +268,10 @@ pythia:
   per_device_train_batch_size: 2
   per_device_eval_batch_size: 4
   output_dir: pythia_model
+  datasets:
+    - alpaca:
+        reverse_augmentation: True
+    - webgpt
```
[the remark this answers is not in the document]
CloseChoice: done

andreaskoepf (approved):
lgtm!

PR #2343, opened 2023-04-06
Use a provider for client env

AbdBarho:
Probably the most consistent way of providing config to the app.

Easily extensible, well understood, no side effects, and not fragile.

A lot of changes are just renaming.

Closes #2334

### DIFF
--- a/website/src/components/Hero.tsx
+++ b/website/src/components/Hero.tsx
@@ -1,15 +1,15 @@
-import { Box, Button, Text, useColorMode, Flex } from "@chakra-ui/react";
-import { User } from "lucide-react";
+import { Box, Button, Text, useColorMode } from "@chakra-ui/react";
 import Image from "next/image";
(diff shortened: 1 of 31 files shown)


notmd:
I believe this doesn't work, NextJs still use the static env var. However, the new API is much safe than the current one.

AbdBarho:
> I believe this doesn't work, NextJs still use the static env var. However, the new API is much safe than the current one.

🤦‍♂️ you are right, when I tried it with `npm run build` it worked, when I ran it from docker it did not work.


Can we just say fuck it and add `/api/config` which returns the config as json?

notmd:
I got a working version now (hopefully). Will update my PR in a few minutes.

notmd:
@AbdBarho can this be closed now?

AbdBarho:
> @AbdBarho can this be closed now?

yeah true

PR #2350, opened 2023-04-06
Add chat deletion functionality

AbdBarho:
Allow users to delete their own chats

Maybe some python wizard can clean up the monstrosity of query I wrote for deletion, I could not get the chat deletion to cascade, so I had to clear the messages first.

### DIFF
--- a/inference/server/oasst_inference_server/routes/chats.py
+++ b/inference/server/oasst_inference_server/routes/chats.py
@@ -58,7 +58,6 @@ async def delete_chat(
     return fastapi.Response(status_code=200)
 
 
-
 @router.post("/{chat_id}/prompter_message")
 async def create_prompter_message(
     chat_id: str,
--- a/inference/server/oasst_inference_server/routes/chats.py
+++ b/inference/server/oasst_inference_server/routes/chats.py
@@ -49,6 +49,16 @@ async def get_chat(
     return chat.to_read()
 
 
+@router.delete("/{chat_id}")
+async def delete_chat(
+    chat_id: str,
+    ucr: UserChatRepository = Depends(deps.create_user_chat_repository),
+):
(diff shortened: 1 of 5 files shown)


notmd:
Can we make it soft delete instead?

AbdBarho:
> Can we make it soft delete instead?

We can, I am not sure if GDPR would be happy about it, unless we have it in our terms of service.

notmd:
Be careful when merging this, it will delete messages that we will use for RL. we better remove the foreign key or soft delete the message.

olliestanley:
> Be careful when merging this, it will delete messages that we will use for RL. we better remove the foreign key or soft delete the message.

I see the logic to favour soft deletion but I would be concerned about people not being able to delete their data fully if they accidentally input private/compromising info

notmd:
> I see the logic to favour soft deletion but I would be concerned about people not being able to delete their data fully if they accidentally input private/compromising info

Yeah, we should ask the user for this. But for me, most of the time I want to delete all the chats because I want to make it easier to manage chats (I believe many other users would do the same). If we always delete messages, we wont get any data to improve the model.

notmd (approved):
web changes LGTM, but we better discuss this before merging the PR

AbdBarho:
I am going to merge this one with the functionality disabled, meaning the user still can't delete, but the code is there.

I think we should add the "hide" option first, then enable this one.

I unlinked the issues.

olliestanley:
> I am going to merge this one with the functionality disabled, meaning the user still can't delete, but the code is there.
> 
> I think we should add the "hide" option first, then enable this one.
> 
> I unlinked the issues.

I think we should also make sure the "hide" option is clearer and easier to click in the UI when we do this

PR #2356, opened 2023-04-06
Add option to find missing strings in only some locales

Psychpsyo:
This adds the option to filter the output of find-missing-locales.py by passing locale codes (like ``en``) to it so that translators don't need to search for their specific languages in the output.


AbdBarho (approved):

yk (approved):
thank you

PR #2361, opened 2023-04-06
Update team page

dvruette:
Added my profile to the team page. I hope twitter is fine as bio link.


AbdBarho (approved):
works fine for now, we should probably add more "link" types

PR #2362, opened 2023-04-07
Add HFSummaryPairs class & fix AnthropicRLHF parsing

andreaskoepf:
- add simpler version of HFSummaries which directly uses the pairs of the OpenAI dataset
- combine consecutive messages of same speaker during loading of AnthropicRLHF dataset, ignore examples with different structure between positive & negative example (only very few)


theblackcat102 (approved):
LGTM, nice work

Issue #2341, opened 2023-04-06
Can we learn from Vicuna to train and fine-tune the model?


libai-lab:
I learned that Vicuña can train a large language model with a very low configuration, but only for research purposes；
Can we learn from Vicuna to train and fine-tune the model?
It would be great if I could, I'm a novice and don't understand anything, but I'm very interested and I'm very sorry to bother

andreaskoepf:
Many people seem to refer Vicuna .. i.e. they seem to have a good PR campaign. Could someone make a comparison between OA llama based models and theirs? It Vicuna actually better and if so in which way...?

kostum123:
> 

Koala model performed better in other languages than Vicuna during my testing. Using more data improved the model's ability to provide more accurate answers in other languages. However, as Koala wrote in their announcement blog, the Koala all model is slightly worse against the Koala distill. Of course, this assessment was for English, and I don't think the test coverage was sufficient. It is also unclear whether they extended the training time or kept it equal when they used more data. If they kept the training time equal, the model may not have seen enough quality data to fully benefit from the increased amount of data. Some sources: 
https://www.chatorg.ai/blog/chat-language-models-tracker
https://chat.lmsys.org/
https://bair.berkeley.edu/blog/2023/04/03/koala/
https://github.com/arnav-gudibande/koala-test-set
https://github.com/lm-sys/FastChat/tree/main
https://vicuna.lmsys.org/

jerzydziewierz:
@andreaskoepf 

> Is Vicuna actually better and if so in which way...?

It appears that at this time, Vicuna is actually better, according to user preference capture on 345 samples with a win probability of 0.77. 

  
Source:

https://lmsys.org/blog/2023-05-03-arena/

olliestanley:
> It appears that at this time, Vicuna is actually better, according to user preference capture on 345 samples with a win probability of 0.77.
> 
> Source:
> 
> https://lmsys.org/blog/2023-05-03-arena/

That is comparing to OA Pythia, where the base model is worse than Vicuna (based on LLaMA), so it makes sense

PR #2364, opened 2023-04-07
Add vicuna dataset

CloseChoice:
#2341

This PR adds the vicuna dataset. We use the filtered vicuna dataset since it only contains english conversations.

The sanity checks are necessary here and filter quite a number of IMO unuseful conversations (where the conversation is empty, where we have less than 2 speakers and where the speakers are not in adjacent order). After all filterings we are left with 84260 of 85071 conversations.


model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -453,3 +456,64 @@ def __init__(self, mode: str = "sft", cache_dir: str = None) -> None:
...
+        if all(speaker1_idx) and all(speaker2_idx):
+            # [Q1, A1, Q2, A2] -> [Q1, A1, Q2, A2]
+            # Use only input_max_length characters
+            truncated_dialogue = [k["value"][:input_max_length] for k in data["conversations"]]
+            # split dialogue in question and answer pairs.
+            # [Q1, A1, Q2, A2] -> [[Q1, A1], [Q2, A2]]
+            pairs = self._partition_list(truncated_dialogue, 2)
+            # add previous history to conversation pairs
+            # e.g. [[Q1, A1], [Q2, A2]] -> [[Q1, A1], [Q1, A1, Q2, A2]]
+            pairs = list(accumulate(pairs, lambda x, y: x + y))
```
CloseChoice, line 497: This is not correct. Should be a reduce.

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -453,3 +456,64 @@ def __init__(self, mode: str = "sft", cache_dir: str = None) -> None:
...
+            if (
+                processed_data := self.process_vicuna_conversations(data, input_max_length=input_max_length)
+            ) is not None:
+                self.pairs.append(processed_data)
+
+    def __len__(self):
+        return len(self.pairs)
+
+    def __getitem__(self, index):
+        return self.pairs[index]
```
CloseChoice, line 519: TODO: Change return depending on mode

model/model_training/custom_datasets/qa_datasets.py
CloseChoice, line 497: This is not yet correct, should just be the truncated dialogue

model/model_training/custom_datasets/qa_datasets.py

model/model_training/custom_datasets/__init__.py
```diff
@@ -45,13 +46,7 @@
...
 ]
 
-RL_DATASETS = [
-    "webgpt",
-    "private_tuning",
-    "alpaca",
-    "hf_summary",
-    "hf_summary_pairs",
-]
+RL_DATASETS = ["webgpt", "private_tuning", "alpaca", "hf_summary", "hf_summary_pairs", "vicuna"]
```
andreaskoepf, line 49: please use ',' at the end to have one entry per line.

andreaskoepf (changes requested):
looks good, just minor formatting change requested.

andreaskoepf (approved):

PR #2367, opened 2023-04-07
Persist chat config

AbdBarho:
Closes https://github.com/LAION-AI/Open-Assistant/issues/2224

### DIFF
--- a/website/src/components/Chat/ChatSection.tsx
+++ b/website/src/components/Chat/ChatSection.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import { ChatConfigForm } from "src/types/Chat";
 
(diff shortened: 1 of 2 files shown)


website/src/components/Chat/ChatSection.tsx
```diff
@@ -1,21 +1,35 @@
...
+          ...modelInfos[0].parameter_configs[0].sampling_parameters,
+          model_config_name: modelInfos[0].name,
+        },
   });
 
+  const config = form.watch();
+
+  useEffect(() => {
+    localStorage.setItem(CHAT_CONFIG_KEY, JSON.stringify(config));
+  }, [config]);
```
notmd, line 31: can use extract this to an empty component? watching on root component causes a lot rerender. You result the form context, so no need to pass any props.

website/src/components/Chat/ChatSection.tsx
```diff
@@ -1,21 +1,35 @@
...
 import { ChatConfigSummary } from "./ChatConfigSummary";
 import { useChatContext } from "./ChatContext";
 import { ChatConversation } from "./ChatConversation";
 
+const CHAT_CONFIG_KEY = "CHAT_CONFIG";
+
 export const ChatSection = ({ chatId }: { chatId: string }) => {
   const { modelInfos } = useChatContext();
 
+  const oldConfig = localStorage.getItem(CHAT_CONFIG_KEY);
```
notmd, line 14: Isn't this crash on the server side?

website/src/components/Chat/ChatSection.tsx
AbdBarho: The `ChatSection` re-renders every time we change anything on the form, so this hook won't render any more than usual.

Anyway, I moved it to its own hook.

website/src/components/Chat/ChatSection.tsx
AbdBarho: It is weird why it did not crash on my side or in the CI pipeline?

I added a check for it

notmd (approved):
The rerender when updating config is caused by the form context. I will refactor this when I work on new chat layout.

PR #2372, opened 2023-04-07
Show work config on each message

AbdBarho:
I considered using a tooltip at the beginning, but I decided to show the config inline just so it is easier to copy for debugging / reproducability purposes. Design could be improved, especially when expanded, but functionality is there.

I am not sure if we want to show all of this info, or if some stuff is "confidential" and should be hidden.

image

image

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -33,6 +33,7 @@ import { MessageEmojiButton } from "../Messages/MessageEmojiButton";
 import { MessageInlineEmojiRow } from "../Messages/MessageInlineEmojiRow";
 import { ChatConfigDrawer } from "./ChatConfigDrawer";
 import { QueueInfoMessage } from "./QueueInfoMessage";
+import { WorkParametersDisplay } from "./WorkParameters";
 interface ChatConversationProps {
   chatId: string;
 }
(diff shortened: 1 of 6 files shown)


olliestanley (approved):
Haven't reviewed the web code but the inference backend changes look good

website/src/components/Messages/MessageInlineEmojiRow.tsx
```diff
@@ -1,9 +1,11 @@
 import { HStack } from "@chakra-ui/react";
 import { ReactNode } from "react";
 
+const stopPropagation = (e) => e.stopPropagation();
```
notmd, line 4: Can we add the typing here?

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -313,6 +317,7 @@ const ChatMessageEntry = memo(function ChatMessageEntry({
                 forceHideCount
                 onClick={handleThumbsDown}
               />
+              <WorkParametersDisplay parameters={workParameters} />
```
notmd, line 320: The params might be null here

AbdBarho: Since we only execute this code block on assistant messages, it won't be null, but I added a check regardless.

andreaskoepf:
> I am not sure if we want to show all of this info, or if some stuff is "confidential" and should be hidden.

nothing confidential here .. but showing the sampling config all the time seems to take a lot of screen space. Maybe could we reduce it to a single line shown with a tiny font? seed is probably not important, model-name + sampling params are most interesting.

AbdBarho:
> Or is is the display toggled?

It is toggled, you can see how it looks in the first screenshot

notmd (approved):

PR #2375, opened 2023-04-07
Updated localization files for Hungarian language for the Chat tab

sedthh:
- updated localization files for chat and stats in Hungarian
- updated old localization files with more translations (only errors.json is omited as
translating error messages would only make debugging more cumbersome)
- updated my details in the team.json file to match my git profile
- ran pre-commit


AbdBarho (approved):
Thanks!

Glad to see more wonky profiles!

Issue #2346, opened 2023-04-06
Add support for loading GPT4All multi-round conversations


olliestanley:
Follow-on to #2331. For now we have corrected the issue of loading multi-round conversations as if single round, but ideally we can also add support for loading the multi-round conversations.

The multi-round data is formatted as JSONL where each line represents a single conversation and looks like this:

```
{
    "source": "...",
    "conversation": [
        {
            "User": "..."
        },
        {
            "Bot": "..."
        },
        {
            "User": "..."
        },
        {
            "Bot": "..."
        }
    ]
}
```

PR #2377, opened 2023-04-07
fix types for prompt dataset

CloseChoice:
Followup #2346:
- fixed the output types for gpt4all

### DIFF
--- a/model/model_training/custom_datasets/prompt_dialogue.py
+++ b/model/model_training/custom_datasets/prompt_dialogue.py
@@ -111,18 +111,16 @@ class Gpt4All(Dataset):
             data_files="data_singleround_pruned_3.jsonl",
             cache_dir=cache_dir,
         )
-        self.rows = [(row["prompt"], row["response"]) for row in dataset["train"]]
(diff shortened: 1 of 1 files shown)


model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -155,9 +153,9 @@ def process_conversation(conv: list[dict[str, None | str]]) -> list[str] | None:
...
 
-    def __getitem__(self, index):
-        dialogue = self.rows[index]
+    def __getitem__(self, index: int) -> list[str] | tuple[str]:
+        dialogue: list = self.rows[index]
         if self.mode == "sft":
             return dialogue
         elif self.mode == "rl":
-            return (dialogue[:-1],)
+            return tuple(dialogue[:-1])
```
andreaskoepf, line 161: This is not correct, for "rl" mode it must be a tuple with the first element being the conversation (e.g. it is `tuple[list[str]]`)... in general it is not so much important whether tuples or lists are used but the nesting structure must be correct.

sft: list[str]
rm: (list[str], list[str])
rl: (list[str],)

andreaskoepf (approved):

PR #2383, opened 2023-04-07
Added stability, and added sponsors to inference interface

yk:


website/src/components/Chat/InferencePoweredBy.tsx
```diff
@@ -0,0 +1,30 @@
+import { Grid, Heading } from "@chakra-ui/react";
+import { TeamMember } from "src/components/TeamMember";
+
+import data from "../../data/team.json";
+
+const SponsorGroup = ({ heading, members }) => {
```
AbdBarho, line 6: Would be great if we could rename this to something like `MemberGroup` and move it to its own file, there is nothing about this that is sponsor specific

AbdBarho (approved):

notmd:
I'm gonna merge this so I can work on the new chat layout. Will do the refactor later.

PR #2391, opened 2023-04-08
DRAFT: Google auth website

AbdBarho:


website/src/pages/admin/manage_user/[id].tsx
```diff
@@ -51,7 +52,8 @@ const ManageUser = ({ user }: InferGetServerSidePropsType<typeof getServerSidePr
   const toast = useToast();
 
   const { data: stats } = uswSWRImmutable<Partial<{ [time in LeaderboardTimeFrame]: LeaderboardEntity }>>(
-    "/api/user_stats?uid=" + user.id,
+    "/api/user_stats?" +
+      new URLSearchParams({ id: user.id, auth_method: user.auth_method, display_name: user.display_name }),
```
Comment (AbdBarho) [inline website/src/pages/admin/manage_user/[id].tsx:56]: fix a bug we already have now:

if you go to the prod deployment / admin / open a user that is logged in with  discord, the stats are missing at the bottom.

notmd (approved):

website/src/pages/admin/manage_user/[id].tsx
Comment (notmd) [reply website/src/pages/admin/manage_user/[id].tsx:56]: Notice that as well, but I'm too lazy to fix it lol

PR #2396, opened 2023-04-08
Improve Chinese translation of 'Is it a bad reply, as an answer to th…

### LINKED ISSUE #2274: Potential vague translation on the question "Is it a bad reply, as an answer to the prompt task?" in Chinese when label assistant reply
The line "Is it a bad reply, as an answer to the prompt task?" is translated as "对于该提示任务来说这不是个好的回复？" in our site, which is literally "Isn't that a good reply for the prompt quest?"

I've always thought it means I need to choose "Yes"（"是"） for a good reply, as "Yes, it is (a good reply)". Only today I decide to try out the English version, do I realize that I have choose the wrong option as the "Yes" here mean "Yes, it is a bad reply".

I suggest translating it as "作为对于该提示任务的回答，这是一个糟糕的回复吗？" which is more clear.

lone-wolf-akela:
…e prompt task?', prevent misunderstanding
Resolves #2274

### DIFF
--- a/website/public/locales/zh/labelling.json
+++ b/website/public/locales/zh/labelling.json
@@ -1,5 +1,5 @@
 {
-  "fails_task.question": "对于该提示任务来说这不是个好的回复？",
+  "fails_task.question": "作为对于该提示任务的回答，这是一个糟糕的回复吗？",
   "hate_speech": "仇恨言论",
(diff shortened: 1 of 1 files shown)


AbdBarho (approved):
Thanks!

PR #2405, opened 2023-04-08
Model training dockerfile

andreaskoepf:
- add minimal dockerfile to build an image to run our {sft,rm,rl}_trainer code

I placed the file in the `./docker` folder (since it's a docker file). This image is useful for development/model-training and will not be deployed on production machines.


yk (approved):
thank you

PR #2411, opened 2023-04-08
Reduced max tokens of llama 30b to 1792 because of OOMs at 2048

yk:


andreaskoepf (approved):
we need flash-attn for inference ..

PR #2412, opened 2023-04-08
Propose an android app

hibatica:
I would like to propose an android front end for Open-Assistant.
I haven't created anything yet but I will try to develop an application for Android.
It might not be pretty, but I hope to make it at least functional, and hopefully others will make it better.

Please note that I'm not very experienced with GitHub, so I fully expect this pull request to be denied.

If you would like me to create a first version and not just an empty folder, let me know.

### DIFF
--- a/android-app/README.md
+++ b/android-app/README.md
@@ -1,0 +1,3 @@
+# Open Assistant Android Application
+
+An android client for Open-Assistant.


hibatica:
Closed pull request to work on a proof of concept first.

PR #2418, opened 2023-04-09
[fix] rephrase weird chinese translation

theblackcat102:
Some of the chinese localization was unreadable or confusing, even as a chinese native speaker I have trouble understanding the meaning without referring to the english UI

#2417


website/public/locales/zh/error.json
dddvanming, line 1: "default": "偶不，阿伯出事了",
Translation Error - the translation does not convey the same meaning as the original sentence.
Change it to, "default": "出了点问题！"

website/public/locales/zh/chat.json
dddvanming, line 1: line11:"您的消息已排队，您当前排在第 {{ queuePosition, number, integer }} 位。" is better

website/public/locales/zh/index.json
dddvanming, line 1: line13:
"a3": "是的，训练数据将以 CC BY 4.0 版权公开。"

dddvanming, line 1: line11:
"这些代码和模型是根据Apache 2.0许可证授权的。"
code translate to 代码,程式 is not often use in zh.

notmd (approved):
@theblackcat102 Please merge when you feel it's ready..

website/public/locales/zh/index.json
theblackcat102: depends on region I guess

website/public/locales/zh/error.json
theblackcat102: sure， I remove the ! to make the meaning more neutral, same as other message

website/public/locales/zh/chat.json
theblackcat102: your translation is way too strict, do you say "你的回覆正在排隊中"?

website/public/locales/zh/index.json
theblackcat102: aren't they the same meaning?

theblackcat102:
@dddvanming any idea on the rest of the translation in `stats.json`? I haven't translated these yet

website/public/locales/zh/index.json
dddvanming: Yes,the term "程式" is primarily used in Hong Kong, Macau, and Taiwan, while "程序" is more commonly used in mainland China.

dddvanming: Traditional Chinese characters are predominantly used in Hong Kong, Macau, and Taiwan, while simplified Chinese characters are used in mainland China.

website/public/locales/zh/chat.json
dddvanming: In my personal estimation, queuing is a more intuitive approach.

website/public/locales/zh/index.json
dddvanming: Yes, but the term "数据" is commonly used in computer science and represents information that is stored in digital form. In computer science, data can take the form of text, images, audio, video, or any other type of digital information. Data can be processed and manipulated by computers, including tasks such as storage, sorting, searching, filtering, conversion, and analysis. Therefore, the term "数据" is more appropriate for use in the field of computer science

dddvanming:
> @dddvanming any idea on the rest of the translation in `stats.json`? I haven't translated these yet

If the meanings of those message tree states are known, the translation will be more accessible.

dddvanming:
> @dddvanming any idea on the rest of the translation in `stats.json`? I haven't translated these yet

If the meanings of those message tree states are known, the translation will be more accessible.

website/public/locales/zh/index.json
dddvanming: and the i18n file is simplified Chinese,so we should follow the the habits of people who use Simplified Chinese.

theblackcat102: hmm, fine

PR #2442, opened 2023-04-10
ua-UK translation improvements

nmeln:
Improve wording and fix mistakes in ua-UK locale strings.
Translate new strings.


AbdBarho (approved):
Thanks!

PR #2443, opened 2023-04-10
fix extraction locales: "Sign In" on a main page

vivekpabari:
Fixed extraction locales: "Sign In" on Main Page Header

### DIFF
--- a/website/src/components/Header/Header.tsx
+++ b/website/src/components/Header/Header.tsx
@@ -13,7 +13,7 @@ import { UserMenu } from "./UserMenu";
 import { UserScore } from "./UserScore";
 
 function AccountButton() {
-  const { t } = useTranslation();
+  const { t } = useTranslation("common");
   const { data: session } = useSession();
   if (session) {
     return null;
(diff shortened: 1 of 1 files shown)


website/src/components/Header/Header.tsx
```diff
@@ -21,7 +21,7 @@ function AccountButton() {
     <Link href="/auth/signin" aria-label="Home">
       <Flex alignItems="center">
         <Button variant="outline" leftIcon={<User size={"20"} />}>
-          Sign in
+          {label}
```
AbdBarho, line 24: it does not need to be a parameter, since the button only does login.

website/src/components/Header/Header.tsx
```diff
@@ -13,6 +13,7 @@ import { UserMenu } from "./UserMenu";
 import { UserScore } from "./UserScore";
 
 function AccountButton() {
+  const { t } = useTranslation();
```
AbdBarho, line 16: ok thank you, one final comment, could you please explicitly use the common namespace?

```js
const { t } = useTranslation("common");
```

This will help with typechecking

Issue #2389, opened 2023-04-08
Special tokens in datasets


CloseChoice:
As it can be seen in [some of the answers](https://open-assistant.github.io/oasst-model-eval/?f=https%3A%2F%2Fraw.githubusercontent.com%2FOpen-Assistant%2Foasst-model-eval%2Fmain%2Fsampling_reports%2Foasst-sft%2F2023-04-07_OpenAssistant_llama-30b-sft-oa-alpaca-epoch-2_sampling_noprefix2.json%0Ahttps%3A%2F%2Fraw.githubusercontent.com%2FOpen-Assistant%2Foasst-model-eval%2Fmain%2Fsampling_reports%2Foasst-sft%2F2023-04-07_OpenAssistant_llama-30b-sft-oa-alpaca-epoch-4_sampling_noprefix2.json) our model outputs quite a number of tokens that are reserved for special purposes and should not appear in text. We need a script with which we can check a dataset for these and point us to problematic samples. 

Investigation of this should be done manually IMO.

olliestanley:
Agree this would be a really useful inclusion

I also think it might be quite beneficial to investigate all data we use for instances of "As a large language model" and related phrases which indicate a high probability of the reply being undesirable/useless, but maybe outside the scope of this issue

CloseChoice:
@olliestanley We encountered that our model believes it is created by openai because of the statements you mentioned. The linked PR removes this (from the 14 datasets mentioned in the PR).

PR #2445, opened 2023-04-10
remove openai references from datasets

CloseChoice:
#2389

remove openai from the datasets:
 - gpt4all
 - alpaca
 - code_alpaca

I checked the following datasets additionally but did not found any entries mentioning openai:
    - joke
    - webgpt
    - minimath
    - humaneval_mbpp_codegen_qa
    - humaneval_mbpp_testgen_qa
    - grade_school_math_instructions
    - recipes
    - cmu_wiki_qa
    - oa_wiki_qa_bart_10000row
    - prosocial_dialogue:
    - explain_prosocial

### DIFF
--- a/model/model_training/custom_datasets/qa_datasets.py
+++ b/model/model_training/custom_datasets/qa_datasets.py
@@ -12,6 +12,7 @@ from urllib.request import urlopen
 
 import numpy as np
 from datasets import load_dataset
+from model_training.custom_datasets.utils import _filter_by_words
 from torch import Generator
 from torch.utils.data import Dataset, Subset, random_split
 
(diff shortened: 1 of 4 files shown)


model/model_training/utils.py
```diff
@@ -364,6 +364,9 @@ def get_dataset(
 
     for data_config in conf.datasets + conf.datasets_extra:
         dataset_name, kwargs = get_dataset_name_and_kwargs_from_data_config(data_config)
+        import pdb
+
+        pdb.set_trace()
```
sanagno, line 369: remove this?

CloseChoice: done

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -503,6 +507,9 @@ def process_vicuna_conversations(data: list[dict[str, None | str]], input_max_le
         for line in data["conversations"]:
             speaker = line["from"]  # 'human' or 'gpt'
             message = line["value"]
+            # if _filter_by_words(message) is None:
```
andreaskoepf, line 510: commented, should be removed?

andreaskoepf (approved):

olliestanley:
I am curious if this has any effect on the oasst dataset(s)?

CloseChoice:
@olliestanley How do you mean that? Oasst datasets are not touched and I did not check them.

olliestanley:
> @olliestanley How do you mean that? Oasst datasets are not touched and I did not check them.

Ok, makes sense! I guess there are probably some references to OpenAI in the oasst data from people copy-pasting ChatGPT responses onto our site, but hopefully only a small amount

andreaskoepf:
@olliestanley I asked Tobias to add this filter since I noticed that a model that was pre-trained (without OA data) replied to some questions with OpenAI statements ... We could do similar filtering for the OA dataset but since we do have full control over the OA data we probably don't need to filter during loading but can directly alter or delete affected messages.

PR #2449, opened 2023-04-10
Add safety server to inference

### LINKED ISSUE #2447: Implement blade2blade safety pipeline into inference
Steps:

- Build safety model endpoint into a FastAPI server in `inference/safety` directory (see `basic_hf_server.py` for example)
- Accept safety threshold/risk tolerance parameter(s) as part of work requests (?)
- Call safety model endpoint prior to performing generation when receiving work requests

olliestanley:
Close #2447.


inference/worker/work.py
```diff
@@ -90,6 +103,11 @@ def handle_work_request(
 
     model_config = worker_config.model_config
 
+    safety_request = inference.SafetyRequest(prompt)
+    safety_response = get_safety_server_response(safety_request)
```
AbdBarho, line 107: should this be optional?

oasst-shared/oasst_shared/schemas/inference.py
```diff
@@ -256,6 +256,19 @@ class GeneralErrorResponse(WorkerResponseBase):
     error: str
 
 
+class SafetyParameters(pydantic.BaseModel):
+    ...
```
AbdBarho, line 260: what does `...` mean in this context, is this the same as `pass`?

inference/safety/main.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+# Allow CORS
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
```
AbdBarho, line 21: this is probably not needed, since no browser will try to reach this directly, right?

AbdBarho (commented):
Thank you

inference/safety/main.py
```diff
@@ -0,0 +1,65 @@
...
+    # warmup
+    input = "|prompter|Hey,how are you?|endoftext|"
+    _ = pipeline.predict(input)
+    pipeline_loaded = True
+
+
+@app.post("/safety", response_model=inference.SafetyResponse)
+async def safety(request: inference.SafetyRequest):
+    global pipeline
+    if not pipeline_loaded:
```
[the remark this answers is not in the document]
olliestanley: Pretty much just however long it takes to download a ~1GB model from HuggingFace, so it's pretty quick on a machine with fast networking. The actual model server uses a queue and separate model thread but that seemed extreme for this use case. Agree a `while` is probably a good course of action here

oasst-shared/oasst_shared/schemas/inference.py
olliestanley: Yes it's just a placeholder, will be removed in a minute as I'll add a parameter to enable/disable safety

inference/safety/main.py
olliestanley: Yeah I think you're right, removed

inference/safety/main.py
```diff
@@ -0,0 +1,57 @@
...
+    _ = pipeline.predict(input)
+    pipeline_loaded = True
+    logger.info("Pipeline loaded")
+
+
+@app.post("/safety", response_model=inference.SafetyResponse)
+async def safety(request: inference.SafetyRequest):
+    global pipeline_loaded, pipeline
+    while not pipeline_loaded:
+        await asyncio.sleep(0.1)
```
yk, line 44: maybe 1s or so :D

inference/safety/main.py
```diff
@@ -0,0 +1,57 @@
...
+    outputs = pipeline.predict(request.inputs)
```
yk, line 45: how does the parallelism of this look like? does the async make it absolutely single-threaded/event-looped?

inference/worker/work.py
```diff
@@ -79,6 +80,18 @@ def _prepare_message(message: inference.MessageRead) -> str:
     return prompt, parameters
 
 
+def prepare_safe_prompt(prompt: str, safety_opinion: str):
+    safety_opinion = re.sub(r"<pad>|</s>", "", safety_opinion).split("<sep>")
+    label, rots = safety_opinion[0], "and".join([x.strip(".") for x in safety_opinion[1:]])
+    if label.strip() == "__casual__":
+        return prompt
+    else:
+        pre_prompt = f"Answer the following request with {label} as responsible chatbot that believes that {rots}: "
```
yk, line 89: we definitely need to track when this happens and feed it back in the response to the backend, and then to the user

oasst-shared/oasst_shared/schemas/inference.py
```diff
@@ -178,6 +178,19 @@ class Thread(pydantic.BaseModel):
     messages: list[MessageRead]
 
 
+class SafetyParameters(pydantic.BaseModel):
+    enabled: bool = True
```
yk, line 182: would set to `False` by default bc if this gets merged & released without the safety server being up yet, we want it to be "backwards compatible"

inference/worker/work.py
yk: I think we can already prepare by e.g. having a safety level 0 to 9, and for now we only send 0 or 9 and we're free to extend these later, but at least we won't have to change data types

inference/worker/work.py
yk: also needs to be saved in the database

yk (commented):
looks good, just left some comments

inference/worker/work.py
yk: (not saying we necessarily need to do this now, we can also plan to do it soon)

inference/safety/main.py
olliestanley: Hmm. Because of `async` we can handle multiple requests concurrently, but it's single-threaded and `pipeline.predict()` will still block the thread so I guess we aren't really getting any benefits of parallelism

inference/worker/work.py
olliestanley: Agree, especially on saving in the database. I think we can use a new form of `Response` in the inference protocol which is sent from worker to backend via the websocket in `handle_work_request`.

inference/safety/main.py
olliestanley: I have implemented a workaround for this, let me know what you think. It's not sophisticated but I think it will help

inference/worker/work.py
olliestanley: I think we can log a new issue which covers:

- Communicating to the backend when the safety model intervenes
- Adding a mechanism for the backend to include safety parameters (based on frontend settings) in work requests

melvinebenezer (approved):
looks fine to me.

inference/worker/work.py
yk: yes we definitely need this. I think it's equally bad that OAI messes with your input than the fact that they don't tell you that they messed with your input

yk (approved):

PR #2452, opened 2023-04-10
Add oa leet10k dataset for SFT training

andreaskoepf:
- add current llama-30 SFT training configuration
- print eval dataset sizes
- remove some verbose prints during startup


model/model_training/configs/config.yaml
```diff
@@ -95,39 +95,67 @@ math:
...
+        val_split: 0.05
     - webgpt:
-        val_split: 0.1
+        val_split: 0.05
+        max_val_set: 250
     - gpt4all:
         val_split: 0.01
+        max_val_set: 250
     - alpaca:
+        reverse_augmentation: true
```
andreaskoepf, line 111: very bad idea

sanagno (approved):
LGTM

Issue #2374, opened 2023-04-07
First few words not appearing in chat tab


Thomasnerd06:
When chatting with the model, I noticed that the first few words dont appear, until the responce is complete. I am not certain, but I think its when there is a comma near the start it removes all the text before the comma until its completed everything.

Not very high priority, just thought I should mention it.

olliestanley:
Thanks for the report! This looks like a duplicate of #2354 so I'll close this thread, but I'll add your observation re this happening after commas to the other issue :)

PR #2458, opened 2023-04-11
Fix #2354, introduce line-buffering + yield multiple objects from list in chat_stream.ts

### LINKED ISSUE #2354: Inference doesn't show whole text during incremental updates.
In the currently deployed version on `open-assistant.io/chat`, the incremental inference updates start from some later tokens. 
1. During generation:
Screenshot at 2023-04-06 19-23-25
2. Final response:
Screenshot at 2023-04-06 19-23-31

In case this isn't resolved yet, I can try to tackle this.

kpoeppel:
I found two bugs in chat_stream.ts, one is the possibility of incomplete lines, so the `iteratorSSE` has to buffer those. The second is the main cause for issue #2354, the Object.fromEntries dictionary creator iteratively overwrites 'data' keys, so the yield has to be wrapped in a loop for yielding all lines.

This PR should resolve issue #2354

### DIFF
--- a/website/src/lib/chat_stream.ts
+++ b/website/src/lib/chat_stream.ts
@@ -64,24 +64,20 @@ export async function* iteratorSSE(stream: ReadableStream<Uint8Array>) {
       continue;
     }
     const full_value = unfinished_line + value;
-    const lines = full_value
-    .split(/\r?\n/)
-    .filter(Boolean);
+    const lines = full_value.split(/\r?\n/).filter(Boolean);
     // do line buffering - otherwise leads to parsing errors
-    if (full_value[full_value.length - 1] != '\n'){
+    if (full_value[full_value.length - 1] != "\n") {
       unfinished_line = lines.pop();
-    }
-    else{
+    } else {
       unfinished_line = "";
     }
 
-    const fields = 
-      lines.map((line) => {
-        const colonIdx = line.indexOf(":");
-        return [line.slice(0, colonIdx), line.slice(colonIdx + 1).trimStart()];
-      });
+    const fields = lines.map((line) => {
+      const colonIdx = line.indexOf(":");
+      return [line.slice(0, colonIdx), line.slice(colonIdx + 1).trimStart()];
+    });
     // yield multiple messages distinctly
-    for (const fidx in fields){
+    for (const fidx in fields) {
       yield Object.fromEntries([fields[fidx]]);
     }
   }
--- a/website/src/lib/chat_stream.ts
+++ b/website/src/lib/chat_stream.ts
@@ -54,6 +54,7 @@ export async function* iteratorSSE(stream: ReadableStream<Uint8Array>) {
 
   let done = false,
     value: string | undefined = "";
+  let unfinished_line = "";
   while (!done) {
     ({ value, done } = await reader.read());
     if (done) {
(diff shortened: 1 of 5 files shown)


website/src/lib/chat_stream.ts
```diff
@@ -61,15 +62,22 @@ export async function* iteratorSSE(stream: ReadableStream<Uint8Array>) {
...
-
-    yield Object.fromEntries(fields);
+    const fields = lines.map((line) => {
+      const colonIdx = line.indexOf(":");
+      return [line.slice(0, colonIdx), line.slice(colonIdx + 1).trimStart()];
+    });
+    // yield multiple messages distinctly
+    for (const fidx in fields) {
+      yield Object.fromEntries([fields[fidx]]);
+    }
```
notmd, line 81: You can use `for of ` here

notmd (commented):
Thank you, it would be great if you could add a test case for the `iteratorSSE`.

website/src/lib/chat_stream.ts
kpoeppel: > Thank you, it would be great if you could add a test case for the `iteratorSSE`.

Test cases are done, it took quite some time to fix issues with `jest` and `TextDecoderStream` and others. With my additions also others shouldn't run into problems now, as I extended the `jest.setup.js`

notmd:
Thanks, can you add test cases that contains "," and ","? https://github.com/LAION-AI/Open-Assistant/issues/2374 report he has problem when the response contains comma and I also facing a similar problem with dot. I think the bug is not on our side, but just to make sure.

kpoeppel:
> Thanks, can you add test cases that contains "," and ","? #2374 report he has problem when the response contains comma and I also facing a similar problem with dot. I think the bug is not on our side, but just to make sure.

Another test was added. 
The problem was on our side in the following sense:
During debugging the main problem was that the first part of the message came in at once and only the last token was taken. Interestingly, it seems like this first comma or similar was the point where the actual token streaming began.

notmd (approved):
The change LGTM, but I would like to have @AbdBarho a final look since he wrote this. But I wonder is there any safer way instead string parsing like this? If the "comma bug" is on our end, I believe we still have many more bugs like this. There is a user report he also has problem with Chinese character

kpoeppel:
> Thank you so much! The change LGTM, but I would like to have @AbdBarho a final look since he wrote this. But I wonder is there any safer way instead string parsing like this? If the "comma bug" is on our end, I believe we still have many more bugs like this. There is a user report he also has problem with Chinese character

I agree that there should be better ways to do this. But if the format of commands stays exactly the same:
"[a-cAC]+: .*\n", the way it is done now including should be the correct one. 

The Object.fromEntries / dictionary style makes little sense IMO, since now it will be single entry dictionaries only. But it's not a problem.

AbdBarho:
> The Object.fromEntries / dictionary style makes little sense IMO, since now it will be single entry dictionaries only. But I kept it for now.

I was trying to stick to the format returned by `EventSource`, but instead of events, we have objects with keys.

kpoeppel:
> > The Object.fromEntries / dictionary style makes little sense IMO, since now it will be single entry dictionaries only. But I kept it for now.
> 
> I was trying to stick to the format returned by `EventSource`, but instead of events, we have objects with keys.

Given the urgency of the issue, I think there should be no further changes for now. If I didn't make that clear yet.

AbdBarho (approved):
LGTM

PR #2477, opened 2023-04-12
Dev_mark

gmarko:
Changes in Readme Files to markdown style.


gmarko:
If it's true, I haven't properly checked the outcome of the changes. We need to re-compose it correctly, separating code from text.

PR #2484, opened 2023-04-12
Implement chat threading UI

notmd:
- Allow editing a prompt (except the root one)
- Allow regenerating a message (only last one for now)

There still a lot layout shift when regenerate message, I will try to address them later.

https://user-images.githubusercontent.com/33456881/231652205-da80aaa2-c18f-4068-aec0-f6640a3781ad.mp4


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -165,23 +170,61 @@ export const ChatConversation = memo(function ChatConversation({ chatId: chatIdP
...
+        ...prompter_arg,
+        created_at: new Date().toISOString(),
+        role: "prompter",
+        state: "complete",
+        score: 0,
+        reports: [],
+      };
+
+      try {
+        // push the dummy message first to avoid layout shift
```
AbdBarho, line 199: what layout shift happens here? and how does this solve it?

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -165,23 +170,61 @@ export const ChatConversation = memo(function ChatConversation({ chatId: chatIdP
...
+        setMessages((messages) => [...messages, dummyMessage]);
+        prompter_message = await post(API_ROUTES.CREATE_PROMPTER_MESSAGE, { arg: prompter_arg });
+        // filter out the dummy message and push the real one
+        setMessages((messages) => [...messages.filter((m) => m.id !== "__dummy__"), prompter_message!]);
+      } catch {
+        // revert on any error
+        setMessages((messages) => [...messages.filter((m) => m.id !== "__dummy__")]);
```
AbdBarho, line 206: if an error happens, we arguably should not continue executing this function at all.

Also, I don't think handling the error gracefully is necessary, there are so many edge cases where stuff could go wrong in the entire app, and we don't handle any of them gracefully.

website/src/components/Chat/ChatConversationTree.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <>
+      <ChatMessageEntry message={tree} {...props}></ChatMessageEntry>
+      {tree.children.length > 0 && <TreeChildren trees={tree.children} {...props} />}
+    </>
+  );
+});
+
+const TreeChildren = ({
+  trees,
```
AbdBarho, line 34: I think the name here might be a bit confusing, this is not an array of trees, but one tree, correct?

website/src/components/Chat/ChatConversationTree.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const hasSibling = length > 1;
+
+  const handlePrevious = useCallback(() => {
+    setIndex((i) => (i > 0 ? i - 1 : i));
+  }, []);
+  const handleNext = useCallback(() => {
+    setIndex((i) => (i < length - 1 ? i + 1 : i));
+  }, [length]);
+
+  const node = (
```
AbdBarho, line 61: can we memo this? or it does not make sense here?

AbdBarho (commented):
Thank you

website/src/components/Chat/ChatConversationTree.tsx
notmd: Yeah it doesn't make sense here. It still renders anyway, there is only one branch it doesn't render, I will update it now.

website/src/components/Chat/ChatConversationTree.tsx
notmd: This is an array of trees, that why I called it `children`. But anyway if you have any better name please suggest.

website/src/components/Chat/ChatMessageEntry.tsx
```diff
@@ -79,10 +79,25 @@ export const ChatMessageEntry = memo(function ChatMessageEntry({
...
+      setIsEditing.off();
+    },
   });
 
+  const handleKeydown = useCallback(
+    (e: KeyboardEvent) => {
+      if (e.key === "Escape") {
+        setIsEditing.off();
+      }
+      if (e.key === "Enter") {
```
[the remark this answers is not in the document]
notmd: I believe using Enter will provide a much better UX here. OpenAI, Discord, and many other chat apps also do the same. I think we should switch the ctrl+enter to enter other places.

website/src/components/Chat/ChatConversation.tsx
notmd: To update a tree, we now relying on `currentRetryingParentId`, we travel the tree until we meet`currentRetryingParentId` and replace all of its children. Without the dummy message, everything still work fine. But from the UX perspective, user expect their edited message to be displayed instantly. Without the dummy, the user has to wait a bit, and then we will render the edited message when it available, so the layout will be jumping around for a bit.

website/src/components/Chat/ChatConversation.tsx
notmd: > if an error happens, we arguably should not continue executing this function at all.

Yes we only continue the function if the prompter message is created successfully. The code below already handle this

```js
if (prompter_message) {
    await initiate_assistant_message({ parentId: prompter_message.id, chatId: chatId });
}
```

website/src/components/Chat/ChatConversationTree.tsx
notmd: Just updated the logic. The node will be rendered anyway when I rework the streaming logic for better UI/UX.

notmd:
@AbdBarho I just push a change that contains important logic for finding parentId when sending a message. The old implementation always use the last message id as parentId which lead to incorrect thread. You can reproduce this bug by following step
- Create new chat, send a prompt and then regenerate a assistant message.
- Now you should see a new node with display index like 2/2. Navigate to the first node (1/2) and send a prompt.
- The new message is a child of node index 2 instead 1.

With the new fix, the message is now child of node 1 which is expected.
The fix is really hacky since the ChatConversation doesn't have information about which thread is being displayed. I will refactor the ChatConversationTree to fully controlled mode later.

AbdBarho:
@notmd I am gettinig the error in the console, I have generated a couple of messages for one chat, and then refereshed the page:
```
next-dev.js?3515:20 Warning: React does not recognize the `retryingParentId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `retryingparentid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```
image

website/src/components/Chat/ChatConversation.tsx
AbdBarho: I see, I wonder if it is necessary to remove the old message in the first place? when we get the response we can remove the last 2 messages and add the new 2? does our architecture allow this?

notmd:
> retryingParentId

Should be fixed by latest commit

AbdBarho (approved):
Maybe for the future?  Don't allow the user to edit their message when a new message is being generating?

notmd:
> Maybe for the future? Don't allow the user to edit their message when a new message is being generating?

I thought I already do that. Let me check again and update in the next PR

PR #2486, opened 2023-04-12
Add support for user deletion

olliestanley:
The aim here is to retain the database "User" entry but wholly anonymised (getting rid of their display name and their provider username)


backend/oasst_backend/user_repository.py
```diff
@@ -119,6 +120,12 @@ def mark_user_deleted(self, id: UUID) -> None:
 
         user.deleted = True
 
+        # Anonymise user data
+        user.display_name = "Deleted User"
```
melvinebenezer, line 124: I see the intention is it to keep this info coherent for inference and backend user data.
like it is used in inference/server/oasst_inference_server/routes/admin.py as well.

Can this be defined in oasst-shared/utils.py.  ?
example
` DELETED_USER = "Deleted User"` 
Similar changes can be thought for the hardcoded prefix "deleted_"
` user.username = f"deleted_{user.id}`

melvinebenezer (changes requested):

melvinebenezer (approved):
thanks :)

AbdBarho (approved):

PR #2490, opened 2023-04-12
Fix python package naming for zhihu-kol package, add gitignore.

kpoeppel:
This PR fixes bad python package naming for the `zhihu-kol` dataset and adds a small gitignore for the build directory of `oasst-shared`.


olliestanley (approved):
Thanks! This will also need an approval from an ML code owner as it touches the ML configs

melvinebenezer (approved):

oasst-shared/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
andreaskoepf, line 1: What does generate this folder? May I ask which OS you are running?

kpoeppel: As far as I remember, for running the inference at one point, oasst_shared needs to be installed (in some venv). It might be that I messed up this installation process, so that it first built the package locally in this subfolder (I guess there was a setup.py still). I can update the PR to remove it. But I think it's always good to keep these potential generated folders out, so that they are not commited by accident. Unsure if pre-commit already takes care of it.

OS is Linux MATE.

PR #2491, opened 2023-04-12
Don't refresh the page after each message

AbdBarho:
I had to revert the change to the `/chat` endpoint, we don't show the empty chat anymore, however, this functionality is still preserved when currently in a chat.


website/src/pages/chat/index.tsx
```diff
@@ -4,33 +4,30 @@ import { getToken } from "next-auth/jwt";
...
 
   return (
     <>
       <Head>
         <title>{t("chat")}</title>
       </Head>
-      <ChatContextProvider modelInfos={modelInfos} messages={[]}>
-        <ChatSection chatId={null} />
-      </ChatContextProvider>
+      <ChatListBase chats={chatResponse} isSideBar={false} className="max-w-5xl mx-auto" />
```
notmd, line 25: This seems to be an ugly UI IMO.

AbdBarho: no problem, we can create another PR and change the design.

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -119,27 +111,28 @@ export const ChatConversation = memo(function ChatConversation({ chatId: chatIdP
...
 
     const prompter_arg: InferencePostPrompterMessageParams = {
-      chat_id: currentChatId,
+      chat_id: chatId,
       content,
       parent_id,
     };
 
     const prompter_message: InferenceMessage = await post(API_ROUTES.CREATE_PROMPTER_MESSAGE, { arg: prompter_arg });
-    mutate(API_ROUTES.LIST_CHAT); // revalidte chat list after create prompter message to make sure the message already has title
```
notmd, line 128: To prevent refresh, we simply can check whether the chatIdProps is null or not here.

notmd (commented):
This change LGTM, it's also prevent a bug when we not finish creating chat and receive the first message the page will not be push to /[id], then we won't able to create new chat, since click create new chat is not trigger rerender the page. However I have a fix in my mind for this, it's quite tricky though. I also notice OAI does face a similar problem, but they decide not to push the page and manage it by a state, which has a downside too.

website/src/components/Chat/ChatConversation.tsx
AbdBarho: `chatIdProps` will never be null in this new version, which is why I check the messages.

notmd: I mean you can do this in the old version. I will switch to the current UI after this anyway since that much better UI. Btw I would like to have #2484 merge first because this PR may cause a lot of conflict with it.

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -119,27 +111,28 @@ export const ChatConversation = memo(function ChatConversation({ chatId: chatIdP
...
+    if (messages.length < 1) {
```
notmd, line 120: `=== 0` is more readable here

AbdBarho: updated

notmd (approved):

PR #2505, opened 2023-04-13
Inference worker inform backend on safety intervention

olliestanley:


inference/worker/work.py
```diff
@@ -115,7 +115,18 @@ def handle_work_request(
     if settings.enable_safety and work_request.safety_parameters.level:
         safety_request = inference.SafetyRequest(inputs=prompt, parameters=work_request.safety_parameters)
         safety_response = get_safety_server_response(safety_request)
-        prompt = get_safety_opinion(prompt, safety_response.outputs, work_request.safety_parameters.level)
+        safe_prompt = get_safety_opinion(prompt, safety_response.outputs, work_request.safety_parameters.level)
+
+        if safe_prompt != prompt:
+            utils.send_response(
```
yk, line 121: could we get as much as possible of the safety response in here? like what was the detected intervention level?

yk (changes requested):

inference/worker/work.py
olliestanley: > could we get as much as possible of the safety response in here? like what was the detected intervention level?

Yeah think so, I wrote this PR before we had that as fleshed out. Will take a look now

yk (approved):
thank you

PR #2507, opened 2023-04-13
implement update chat title UI

notmd:
please merge #2484 before this to avoid conflict.

image

### DIFF
--- a/website/src/components/Chat/ChatListBase.tsx
+++ b/website/src/components/Chat/ChatListBase.tsx
@@ -4,7 +4,7 @@ import { Button, Card, CardProps } from "@chakra-ui/react";
 import { Plus } from "lucide-react";
 import Link from "next/link";
 import { useTranslation } from "next-i18next";
-import { memo } from "react";
+import { memo, useCallback } from "react";
 import SimpleBar from "simplebar-react";
 import { get } from "src/lib/api";
 import { API_ROUTES } from "src/lib/routes";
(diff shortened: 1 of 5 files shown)


website/src/components/Chat/ChatListBase.tsx
```diff
@@ -15,11 +15,24 @@ import { HEADER_HEIGHT } from "../Header/Header";
...
 
+  const handleUpdateTitle = useCallback(
+    ({ chatId, title }: { chatId: string; title: string }) => {
+      mutateChats(
+        (chatResponse) => ({
+          ...chatResponse,
+          chats: chatResponse?.chats.map((chat) => (chat.id === chatId ? { ...chat, title } : chat)) || [],
+        }),
+        false
+      );
```
AbdBarho, line 31: is this necessary? I assume we update the title in the `ChatItem`, and then we just refresh here? do we need to re-assign the title here again?.

website/src/components/Chat/ChatListItem.tsx
```diff
@@ -1,41 +1,119 @@
...
+        setIsEditing.off();
+      }
+    },
+  });
+  const { trigger: updateChatTitle, isMutating: isUpdatingTitle } = useSWRMutation(
+    API_ROUTES.UPDATE_CHAT_TITLE(chat.id),
+    put
+  );
+  const handleConfirmEdit = useCallback(async () => {
+    const title = inputRef.current?.value;
```
AbdBarho, line 38: `value.trim()`

website/src/components/Chat/ChatListItem.tsx
```diff
@@ -1,41 +1,119 @@
...
+    [handleConfirmEdit, setIsEditing]
+  );
+
   return (
     <Button
-      as={Link}
-      href={ROUTES.CHAT(chat.id)}
+      // @ts-expect-error error due to dynamicly changing as prop
+      ref={rootRef}
+      {...(!isEditing ? { as: Link, href: ROUTES.CHAT(chat.id) } : { as: "div" })}
```
AbdBarho, line 61: Does this work? keep it as `Link` & remove the href when editing (or set to '#'), then it would not navigate.

website/src/components/Chat/ChatListBase.tsx
notmd: Just for instantly update. It also saves us a request since we already have enough information to display. The second param `false` means not to revalidate.

AbdBarho: aha, I see, I assumed if we wanted to show instant updates, we would do that in the `ChatItem` directly? maybe it is simpler?

website/src/components/Chat/ChatListItem.tsx
notmd: Done!

website/src/components/Chat/ChatListBase.tsx
notmd: How can we update it in the ChatItem? Using a state?

AbdBarho: would be the most straight forward solution. what do you think?

AbdBarho: I see, totally forgot about the other menu for mobile.

AbdBarho (approved):

PR #2511, opened 2023-04-14
fix custom preset selection

notmd:
The old implementation will not work if the custom preset have the same config as one of the predefined presets. The predefined preset will take president over the custom making the inputs is disabled.
I also fix the config cache bug.


website/src/components/Chat/ChatConfigDrawer.tsx
```diff
@@ -118,7 +118,9 @@ const ChatConfigForm = () => {
   const { control, register, reset, getValues } = useFormContext<ChatConfigFormData>();
   const selectedModel = useWatch({ name: "model_config_name", control: control });
   const presets = modelInfos.find((model) => model.name === selectedModel)!.parameter_configs;
-
+  const [selectedPresetName, setSelectedPresetName] = useState(
+    () => presets.find((preset) => areParametersEqual(preset.sampling_parameters, getValues()))?.name ?? presets[0].name
```
AbdBarho, line 122: should `presets[0].name` be `customPresetName`?

notmd: Done

AbdBarho (approved):

PR #2513, opened 2023-04-14
fix double emoji in `MessageTableEntry` component

notmd:


AbdBarho (approved):
haha! looks cool

image

PR #2520, opened 2023-04-14
Feature/dataset entry

### LINKED ISSUE #2708: Fill <|system|> prompt during supervised fine-tuning with lang, len & text-labels
Optional "background" information about the style of the assistant-response to generate should be provided as a prompt via `<|system|>...<eos>` messages (invisible during chat). To make effective use of this `<|system|>` information during inference (beyond zero shot) it needs to be already present during SFT.

During dialogue-formatting in the [DialogueDataCollator](https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/custom_datasets/dialogue_collator.py)  an optional `<|system|>` messages should be placed before `<|assistant|>` responses for OA training examples (with a configurable probability). More information than just the message texts will be required for this (e.g. access to the full oasst1 items). @CloseChoice already worked into this direction by adding a `DatasetEntry` class in PR https://github.com/LAION-AI/Open-Assistant/pull/2520. This feature should be built on top of this pr by extending `DatasetEntry` to offer full access to the oasst1 items.

The following information should be added to the system-prompt:
- language-tag: of the response message
- length: (number of tokens to generate)
- text labels: "quality", "humor", "creativity",

A system message should look like this:
```
<|system|>
lang: en
length: 123
quality: 1.0
humor: 0.2
creativity: 0.3
<|endoftext|> 
```

Presence and order of fields could be randomized. For other datasets than OA only basic information like maybe lang and length could be added.

CloseChoice:
closes #2708

Add pydantic basemodel class (equivalent to dataclass but with stronger guarantees) to return from dolly dataset. Add the formatting functionality in the dataset entry class.
This PR does quite a bit:
 - add pydantic dependency
 - introduce a new DatasetEntry class, which provides a method to do the formatting based on the mode and the `QA_SPECIAL_TOKENS` and the `eos_token`. This class should work as a general pattern to store and format single dataset entries. This class should also remove all the formatting errors we had previously with the datasets.
 - add tests for `DialogueDataCollator`. I trained a minimal tokenizer only on the tokens that are present in the tests to not bloat the code (still a lot of LOC).
 - added tests for the interplay of `DialogueDataCollator` and the newly introduced `DatasetEntry` class.
 - add small fixes to be backwards compatible in handling the new `DatasetEntry` and old rows from the dataset.

todos:
 - [x] mask system
 - [x] remove changes in config
 - [x] write tests for formatting of dataset entry


model/model_training/tests/resources/data_collator/tokenizer.json
dvruette, line 1: What's this tokenizer for? is it committed on purpose?

model/model_training/tests/test_dialogue_data_collator.py
```diff
@@ -0,0 +1,213 @@
...
+import torch
+from model_training.custom_datasets.dialogue_collator import DialogueDataCollator
+from model_training.custom_datasets.formatting import QA_SPECIAL_TOKENS, DatasetEntry
+from model_training.utils import match_tokenizer_name
+from transformers.models.auto.tokenization_auto import AutoTokenizer
+
+
+@pytest.fixture
+def pythia_tokenizer():
+    tokenizer = AutoTokenizer.from_pretrained("tests/resources/data_collator", local_files_only=True)
```
dvruette, line 11: aah i see now, the tokenizer is for the tests

model/model_training/custom_datasets/formatting.py
```diff
@@ -15,20 +28,291 @@ def format_system_prefix(prefix, eos_token):
...
+                case (str(), None):
+                    qa_list.append(f"{QA_SPECIAL_TOKENS['Question']}{q}{eos_token}")
+                case (None, None):
+                    break
+                case (None, str()):
+                    raise ValueError("Received answer without getting corresponding question. Aborting")
+        if mode == Mode.sft:
+            return qa_list
+        elif mode == Mode.rm:
+            raise NotImplementedError("This is currently not implemented.")
```
dvruette, line 282: Does this mean that RM training will not work after this is merged?

model/model_training/custom_datasets/formatting.py
```diff
@@ -15,20 +28,291 @@ def format_system_prefix(prefix, eos_token):
...
+
+    def system_tag(self, eos_token: str) -> str | None:
+        relevant_system_infos = [
+            (k, v)
+            for k, v in self.__dict__.items()
+            if k not in ["questions", "answers"] and v is not None and v.replace("\n", "")
+        ]
+        if len(relevant_system_infos) > 0:
+            shuffle(relevant_system_infos)
+            system_tag_key_values = "\n".join([f"{k}: {v}" for k, v in relevant_system_infos])
```
dvruette, line 249: Should the values be quantized here or are they already within a certain set? (I think real values could be difficult to understand for the model)

model/model_training/custom_datasets/formatting.py
```diff
@@ -15,20 +28,291 @@ def format_system_prefix(prefix, eos_token):
...
-def format_pairs(pairs, eos_token, add_initial_reply_token=False):
-    conversations = [
-        "{}{}{}".format(QA_SPECIAL_TOKENS["Question" if i % 2 == 0 else "Answer"], pairs[i], eos_token)
-        for i in range(len(pairs))
-    ]
-    if add_initial_reply_token:
-        conversations.append(QA_SPECIAL_TOKENS["Answer"])
-    return conversations
+@unique
+class Language(str, Enum):
```
dvruette, line 32: worth considering extracting this class into a separate file. not necessary, i'll leave it up to you

dvruette (approved):
Do I understand correctly that right now this system message format isn't used e.g. for the OA dataset? Just because I didn't spot where it gets hooked into the data pipeline.

But looks great! can be merged

model/model_training/custom_datasets/formatting.py
CloseChoice: Yep, that is correct. Currently it is only implemented for Dolly, which does not support rm, so no change for now. But this needs to be implemented soon.

model/model_training/custom_datasets/formatting.py
CloseChoice: The keys here are one of `context`, `lang`, `length`, `quality`, `humor` and `creativity`. The values for `quality`, `humor` and `creativity` can be floating point numbers between 0 and 1 (this is checked by the validators and a test for this is added aswell, see [here](https://github.com/LAION-AI/Open-Assistant/pull/2520/files#diff-591736dab256f33df5fdc944907185433119c5c8032cdc40ada2beeba0a793d7R49)).

model/model_training/custom_datasets/formatting.py
CloseChoice: good point. Done

model/model_training/custom_datasets/formatting.py
dvruette: yes, so if they can be anything between 0 and 1 (including e.g. 0.1826357216), then I think we should round them (e.g. to a single decimal) to make it easier for the model to understand.

CloseChoice: done

PR #2523, opened 2023-04-14
Add delete account page

AbdBarho:


inference/server/oasst_inference_server/admin.py
```diff
@@ -0,0 +1,20 @@
...
+from oasst_inference_server import database, models
+from oasst_shared import utils as shared_utils
+
+
+async def delete_user_from_db(session: database.AsyncSession, user_id: str) -> bool:
+    """Deletes a user."""
+    logger.info(f"Deleting user {user_id}")
+    user = await session.get(models.DbUser, user_id)
+    if not user:
+        return False
```
olliestanley, line 11: I think we should just raise an exception (404) on fail here instead of returning `bool`, that's more in line with the rest of the codebase

website/src/pages/api/account/delete.ts
```diff
@@ -0,0 +1,55 @@
...
+    const client = createInferenceClient(token);
+    await client.delete_account(token.sub);
+    console.log(`user ${token.sub} deleted from inference`);
+  } catch (err) {
+    if (err instanceof AxiosError && err.response.status === 404) {
+      // user does not exist in the inference backend, they have not send any chats
+      // that is okay, we can continue
+      console.log(`user ${token.sub} does not exist on inference`);
+    } else {
+      console.error("could not delete user from inference backend", err);
```
notmd, line 40: I think we need to setup the logger for frontend soon, since console.log is blocking the cpu

notmd (approved):
Lgtm

olliestanley (approved):
Backend changes LGTM

PR #2526, opened 2023-04-14
Add default to user deleted column

olliestanley:
Not having this doesn't appear to be causing any issues but making this change now might prevent future headaches.


yk (approved):
thank you

PR #2535, opened 2023-04-15
remove references from dolly

CloseChoice:
Remove references from the context in the dolly dataset. Found and removed the following type of references;
 - referencing by number `[1]`
 - referencing by lower case letters `[a]`
 - indicating missing references `[citation needed]`

Also added dolly to the custom dataset entry point

### DIFF
--- a/model/model_training/custom_datasets/qa_datasets.py
+++ b/model/model_training/custom_datasets/qa_datasets.py
@@ -552,6 +552,7 @@ class DatabricksDolly15k(Dataset):
     def __init__(self, cache_dir: str | Path, mode: str = "sft", input_max_length: int = 2048) -> None:
         super().__init__()
         self.rows = []
+        self.citation_regex = re.compile(r"\[[a-z]\]")  # removes citations in the form of e.g. [a]
         if mode not in ("sft", "rl"):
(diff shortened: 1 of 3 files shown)


model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -552,6 +552,7 @@ class DatabricksDolly15k(Dataset):
     def __init__(self, cache_dir: str | Path, mode: str = "sft", input_max_length: int = 2048) -> None:
         super().__init__()
         self.rows = []
+        self.citation_regex = re.compile(r"\[[a-z]\]")  # removes citations in the form of e.g. [a]
```
andreaskoepf, line 555: could we please also use the same regex that is defined here: https://github.com/LAION-AI/Open-Assistant/blob/2e6f22c2a137ee60ef68cf908073c1e29babd633/model/model_training/custom_datasets/qa_datasets.py#L20

maybe also add upper case single latter references?

CloseChoice: we use that regex already, see line 565. Will add upper case letters aswell

andreaskoepf (approved):
thanks!

PR #2536, opened 2023-04-15
Fix chat layout overflow issue

notmd:
Just a dirty workaround. Probably will be reverted when I refactor the chat layout again.
see https://web.dev.open-assistant.io/chat/02ae0c6a-6cd3-4f6c-b9b4-754f37fa7b1d


AbdBarho (approved):
Lets hope this does not get too complex to salvage.

PR #2538, opened 2023-04-15
hotfix for cache & drawer bug

yk:


AbdBarho (approved):
We did this in pair programming!

PR #2541, opened 2023-04-15
Allow ctrl for new lines

AbdBarho:


notmd:
Does this make sense? I haven't see any chat app support this.

andreaskoepf:
I would think shift + enter ..

AbdBarho:
not really, this was just an idea I had 🤷‍♀️

notmd:
I dont think we should support this

PR #2542, opened 2023-04-15
Multiple lines in chat

alexandrelefourner:
Allowing users to send multiple lines in the request to the assistant.


AbdBarho:
@alexandrelefourner thank you! can you also check for the ctrl key? similar to #2541

then we can merge this one!

AbdBarho (approved):

PR #2559, opened 2023-04-15
Merge leetcode contest dataset with leet10k

ehartford:


olliestanley:
I am slightly concerned by the behaviour we would get when loading this in HuggingFace Datasets with these changes.

As I understand the code, we are pushing two `.parquet` files to the dataset [here](https://huggingface.co/datasets/ehartford/oa_leet10k/tree/main/data). One contains original oa_leet10k, and the other contains oa_leet10k *and* the new contests data. If so, I believe that

`datasets.load_dataset("ehartford/oa_leet10k")`

Would load a Dataset which contains a duplicate of each oa_leet10k problem, right? Please correct me if I am wrong here, but if not I think we could fix this by not uploading a "merged" parquet at all, and just uploading one for oa_leet10k and one for contests. Then the `load_dataset()` call would load both or we could specify one by passing the `data_files` argument.

ehartford:
Thank you for that feedback I'll implement by Tuesday

ehartford:
Ok will do

ehartford:
@olliestanley I updated this; does this address your concerns?

olliestanley (approved):
Yes, this is great thank you! Just pulled the data down locally and the behaviour is right now. Apologies for the delay in getting this merged

PR #2562, opened 2023-04-15
Fix custom preset display

### LINKED ISSUE #2544: Preset is always stuck to custom
in the chat UI, the preset shown below the message is always "custom"

We could in theory also remove the model name and preset entirely, since the user can already see all of this information from the expanded config, and it makes the UI less clutterd

notmd:
fix #2544
The problem is the `model_id` in `work_parameters` is not the same as the name in `ModelInfo`. So I assume all models have the same preset config.

### DIFF
--- a/website/src/components/Chat/WorkParameters.tsx
+++ b/website/src/components/Chat/WorkParameters.tsx
@@ -5,8 +5,6 @@ import { memo } from "react";
 import { JsonCard } from "src/components/JsonCard";
 import { InferenceMessage, SamplingParameters } from "src/types/Chat";
 
-import { useChatContext } from "./ChatContext";
-
 export const areParametersEqual = (a: SamplingParameters, b: SamplingParameters) => {
   return (
     a.top_k === b.top_k &&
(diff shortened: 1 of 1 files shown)


website/src/components/Chat/WorkParameters.tsx
```diff
@@ -24,12 +24,11 @@ export const WorkParametersDisplay = memo(function WorkParametersDisplay({
   parameters: NonNullable<InferenceMessage["work_parameters"]>;
 }) {
   const { seed: _, ...rest } = parameters;
-  const model_id = parameters.model_config.model_id;
+  const model_id = parameters.model_config.model_id.replace("OpenAssistant/", "");
```
notmd, line 27: I also remove the prefix here

AbdBarho: this is only for display, correct?

notmd: Yes

notmd:
Tbh I think hiding the preset would make more sense. Since we don't have enough information to show the correct preset. Yannic just make the change to `k50` and with the current logic, old message with `k50` will be display as custom

AbdBarho:
makes sense, can we remove the preset and model name entirely? there is only one model anyway...

notmd:
@AbdBarho I remove the preset, still keep the model for now. We might need that in future, because the Pythia is coming

AbdBarho (approved):
Ok, we can always expand the config json and see all info there, so maybe we could reduce clutter in the chat page by removing the model name at some point.

PR #2566, opened 2023-04-15
Fix update chat visibility

notmd:
Axios send data as body instead query. I can change to send as query instead, but I think send as body is more consistent.
cc @yk


yk (approved):
Nice catch, thank you!

AbdBarho (approved):

PR #2578, opened 2023-04-16
Only fetch stats for logged in users

### LINKED ISSUE #2569: Understand error logs
I see a lot of errors like this on the web container:
```
OasstError {
  message: 'User not found',
  errorCode: 4002,
  httpStatusCode: 404,
  path: '/api/v1/frontend_users/discord/1111',
  method: 'GET'
}
OasstError {
  message: 'User not found',
  errorCode: 4002
  httpStatusCode: 404,
  path: '/api/v1/leaderboards/day?max_count=5',
  method: 'GET'
}
```

we should try and understand what is causing this.

My hypothsis:
new users login to the webpage, the website will try to get their stats & leaderboards, but fails to do so because they have not contributed any data yet, and therefor still don't have any account in the backend.

AbdBarho:
Related to #2569

I will have to check in the console after deployment to see if it actually fixes the problem, but it worked on my local setup.

### DIFF
--- a/website/src/pages/api/leaderboard.ts
+++ b/website/src/pages/api/leaderboard.ts
@@ -5,6 +5,12 @@ import { getBackendUserCore } from "src/lib/users";
 import { LeaderboardTimeFrame } from "src/types/Leaderboard";
 
 export default withoutRole("banned", async (req, res, token) => {
+  if (!token.backendUserId) {
+    // user has not yet accepted the terms of service, and therefor does not yet exist in the backend
+    /// skip the request entirely
+    return res.status(200).json(null);
+  }
+
   const oasstApiClient = await createApiClient(token);
   const backendUser = await getBackendUserCore(token.sub);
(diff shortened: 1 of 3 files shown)


website/src/pages/api/leaderboard.ts
```diff
@@ -5,6 +5,12 @@ import { getBackendUserCore } from "src/lib/users";
 import { LeaderboardTimeFrame } from "src/types/Leaderboard";
 
 export default withoutRole("banned", async (req, res, token) => {
+  if (!token.backendUserId) {
+    // user has not yet accepted the terms of service, and therefor does not yet exist in the backend
+    /// skip the request entirely
+    return res.status(200).json(null);
```
notmd, line 11: Does this break the UI?

AbdBarho: No it should not, the leaderboard will be empty.

It is also not visble since the ToS wrapper covers the entire screen, after the user accepts ToS the webpage is refreshed and a new request is made.

notmd (approved):

PR #2595, opened 2023-04-16
Fix instructions on how to run the inference server variant 1 

Rudd-O:
Fixes https://github.com/LAION-AI/Open-Assistant/issues/2594 .

### DIFF
--- a/inference/README.md
+++ b/inference/README.md
@@ -26,31 +26,37 @@ docker compose --profile inference up -d
 Tail the logs:
 
 ```shell
+# cd /root/of/this/repository
 docker compose logs -f    \
     inference-server      \
     inference-worker      \
-    inference-text-client \
-    inference-text-generation-server
+    inference-safety
 ```
-
-Attach to the text-client, and start chatting:
-
-```shell
-docker attach open-assistant-inference-text-client-1
-```
-
-> **Note:** In the last step, `open-assistant-inference-text-client-1` refers to
-> the name of the `text-client` container started in step 2.
 
 > **Note:** The compose file contains the bind mounts enabling you to develop on
 > the modules of the inference stack, and the `oasst-shared` package, without
 > rebuilding.
 
-> **Note:** You can spin up any number of workers by adjusting the number of
-> replicas of the `inference-worker` service to your liking.
+> **Note:** You can change the model by editing variable `MODEL_CONFIG_NAME`
+> in the `docker-compose.yaml` file.  Valid model names can be found at
+> https://huggingface.co/OpenAssistant .
 
-> **Note:** Please wait for the `inference-text-generation-server` service to
-> output `{"message":"Connected"}` before starting to chat.
+> **Note:** You can spin up any number of workers by adjusting the number of
+> replicas of the `inference-worker` service (in the `docker-compose.yaml` file)
+> to your liking.
+
+> **Note:** Please wait for the `open-assistant-inference-worker` services to
+> finish downloading models before continuing to the next step.
+
+Spin up the text-client, and start chatting:
+
+```shell
+# cd /root/of/this/repository
+cd inference/text-client
+pip3 install --user -r requirements.txt
+python __main__.py
+# You'll soon see a `User:` prompt, where you can type your prompts.
+```
 
 ## Development Variant 2 (tmux terminal multiplexing)
 
--- a/inference/README.md
+++ b/inference/README.md
@@ -1,9 +1,9 @@
 # OpenAssistant Inference
 
-Preliminary implementation of the inference engine for OpenAssistant.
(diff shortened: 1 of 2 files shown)


Rudd-O:
I haven't yet figured out how to make the models downloaded by the worker persist across dockercompose up and dockercompose down.  If a kind soul would let me know what needs to be done in order for the user to not have to redownload the models every time the stack starts, I would be so happy and I would add it to the README too.

AbdBarho:
probably like this:

```yml
# docker-compose.override.yml
services:
  inference-worker:
    volumes:
      - "./oasst-shared:/opt/inference/lib/oasst-shared"
      - "./inference/worker:/opt/inference/worker"
      - worker-cache:/data

volumes:
  worker-cache:
```

I would just like to state again that this should not be necessary unless you wan't develop the website.

yk:
thank you for this! yes I'd add a description that this is strictly a development setup to develop the system and website. quite a number of people are looking to run OA locally.

inference/README.md
```diff
@@ -26,31 +26,37 @@ docker compose --profile inference up -d
...
-> the name of the `text-client` container started in step 2.
-
 > **Note:** The compose file contains the bind mounts enabling you to develop on
 > the modules of the inference stack, and the `oasst-shared` package, without
 > rebuilding.
 
+> **Note:** You can change the model by editing variable `MODEL_CONFIG_NAME` in
+> the `docker-compose.yaml` file. Valid model names can be found at
+> https://huggingface.co/OpenAssistant .
+
```
farmisen, line 43: huggingface only gives the model names, the valid model config names mapping to the model names, can be found in https://github.com/LAION-AI/Open-Assistant/blob/main/oasst-shared/oasst_shared/model_configs.py#L23

pevogam:
> probably like this:
> 
> ```yaml
> # docker-compose.override.yml
> services:
>   inference-worker:
>     volumes:
>       - "./oasst-shared:/opt/inference/lib/oasst-shared"
>       - "./inference/worker:/opt/inference/worker"
>       - worker-cache:/data
> 
> volumes:
>   worker-cache:
> ```
> 
> I would just like to state again that this should not be necessary unless you wan't develop the website.

@Rudd-O Are you planning to integrate this into the current PR or should we all rely on overriding YAML config to achieve it?

inference/README.md
Rudd-O: Will fix.

Rudd-O:
@AbdBarho 

> probably like this:

No, I don't think that works.  Stuff is stored in a bunch of places including `~/cache` according to the `worker_full_main.sh` script.  It isn't clear to me where.

This stuff needs to be cleaned up for real, as in formally define where these programs are storing downloaded data — so that users can define mounts that will preserve data across runs.

Rudd-O:
> @Rudd-O Are you planning to integrate this into the current PR or should we all rely on overriding YAML config to achieve it?

Not planning to integrate these, the instructions seem false to me.

inference/README.md
```diff
@@ -26,31 +29,37 @@ docker compose --profile inference up -d
 Tail the logs:
 
 ```shell
+# cd /root/of/this/repository
 docker compose logs -f    \
     inference-server      \
     inference-worker      \
-    inference-text-client \
-    inference-text-generation-server
+    inference-safety
```
olliestanley, line 36: inference-safety is removed from inference profile in recent main

Rudd-O: OK.  I will kill it here too.

pevogam:
> @AbdBarho
> 
> > probably like this:
> 
> No, I don't think that works. Stuff is stored in a bunch of places including `~/.cache` according to the `worker_full_main.sh` script. It isn't clear to me where.
> 
> This stuff needs to be cleaned up for real, as in formally define where these programs are storing downloaded data — so that users can define mounts that will preserve data across runs.

Hmmm, it worked just fine for me with
```diff
diff --git a/docker-compose.yaml b/docker-compose.yaml
index c6333dac..345df8f4 100644
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -252,6 +252,7 @@ services:
     volumes:
       - "./oasst-shared:/opt/inference/lib/oasst-shared"
       - "./inference/worker:/opt/inference/worker"
+      - worker-cache:/data
     deploy:
       replicas: 1
     profiles: ["inference"]
@@ -335,3 +336,4 @@ volumes:
   netdataconfig:
   netdatalib:
   netdatacache:
+  worker-cache:
```
So if you want you can try to see if the model is always downloaded for you with a diff like this one.

olliestanley:
These changes were integrated into #2969, which I have added a co-author tag to for @Rudd-O. Thanks :)

Issue #1832, opened 2023-02-24
Load font using `next/font`


notmd:
https://github.com/LAION-AI/Open-Assistant/blob/d897be935706fde223b0283ce198c501a329f29a/website/src/styles/globals.css#L11
We are loading font in CSS but using [next/font](https://nextjs.org/docs/basic-features/font-optimization) will improve a lot loading performance

CheckMC:
I can work on this one

notmd:
@CheckMC thanks, it's yours!

spencerwmiles:
@notmd I'll take this if @CheckMC is no longer interested!

PR #2600, opened 2023-04-16
Replace fonts loaded via stylesheet with use of next/font

spencerwmiles:
Solves #1832 

I wasn't sure to use local font or just use google font. I can undo that change if use of Google font is undesirable


website/package.json
```diff
@@ -37,7 +37,7 @@
     "@marsidev/react-turnstile": "^0.0.7",
     "@next-auth/prisma-adapter": "^1.0.5",
     "@next/bundle-analyzer": "^13.2.4",
-    "@next/font": "^13.2.4",
+    "@next/font": "13.0.6",
```
notmd, line 40: Is there any reason that you have pin the version?

notmd (commented):
Thanks

website/package.json
spencerwmiles: To match the version of `next` that is currently also using the same version (also pinned to 13.0.6)

The same was also done for `eslint-config-next` and `next` itself. The previously installed version was incompatible  and would result in error in dev and/or builds.

notmd: So we need to update next then, it's been long time since our last deps bump

notmd (approved):

PR #2626, opened 2023-04-16
Remove unused `npm` dependencies from website

lgeiger:
This PR removes a few unused `npm` dependencies from the website requirements.


website/package.json
```diff
@@ -62,7 +60,6 @@
     "next-auth": "^4.20.1",
     "next-i18next": "^13.2.2",
     "nodemailer": "^6.9.1",
-    "npm": "^9.6.2",
```
lgeiger, line 65: `npm` is meant to be installed globally, so there is no need to include it in `package.json`

website/package.json
```diff
@@ -42,13 +42,11 @@
     "@prisma/client": "^4.11.0",
     "@tailwindcss/forms": "^0.5.3",
     "@tanstack/react-table": "^8.8.2",
-    "accept-language-parser": "^1.5.0",
     "autoprefixer": "^10.4.14",
     "axios": "^1.3.4",
     "boolean": "^3.2.0",
     "chart.js": "^4.2.1",
     "clsx": "^1.2.1",
-    "cookies": "^0.8.0",
```
lgeiger, line 51: This dependency was added in #1893 but appears to be unused now.

website/package.json
```diff
@@ -42,13 +42,11 @@
     "@prisma/client": "^4.11.0",
     "@tailwindcss/forms": "^0.5.3",
     "@tanstack/react-table": "^8.8.2",
-    "accept-language-parser": "^1.5.0",
```
lgeiger, line 45: This dependency was added in #881 but appears to be unused now.

notmd (approved):
Thanks, can you resolve the conflict?

PR #2631, opened 2023-04-16
Forever Dreaming TV Transcript Dataset

sedthh:
I have finally finished crawling foreverdreaming.com. The transcripts are about 5% of all the content on the website.
However, we have decided not to share the crawler notebook this time, because it would allow anyone to mirror all the contents on the website just by changing a few lines of code. The owner of foreverdreaming has invested a lot of time and resources into running their website and it simply would not be fair towards them. 
We have discussed this on Discord. 

The dataset is https://huggingface.co/datasets/sedthh/fd_dialogue

### DIFF
--- a/data/datasets/README.md
+++ b/data/datasets/README.md
@@ -180,7 +180,7 @@ import pandas as pd
 df = pd.read_json(...) # or any other way
 
 # Save the file in the Parquet format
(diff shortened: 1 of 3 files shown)


olliestanley (approved):
Thank you!

PR #2641, opened 2023-04-17
send flagged message notification to discord

notmd:
Send a notification to Discord so the mod team can quickly take action. This is a temporal workaround for https://github.com/LAION-AI/Open-Assistant/issues/968. Also we don't need a UI for it.

image


backend/oasst_backend/prompt_repository.py
```diff
@@ -546,6 +547,8 @@ def store_text_labels(self, text_labels: protocol_schema.TextLabels) -> tuple[Te
                         message_id, protocol_schema.EmojiOp.add, protocol_schema.EmojiCode.red_flag
                     )
 
+                    discord.send_new_report_message(message=message, label_text=text_labels.text)
```
melvinebenezer, line 550: send message_id, this would be easier for the mod team to search the DB

melvinebenezer (changes requested):

backend/oasst_backend/prompt_repository.py
notmd: I already send the message url. Is that ok?

melvinebenezer:
Thanks @notmd :). 
while you are at it. was thinking if you could 

1. move the sending to a celery task (check scheduled_tasks.py). 
2. save the flagged message ids in Redis list with a  "date in the name" and  TTL=24hrs
3. have a scheduled task in celery that will compose a discord message with the number of flagged messages, their IDs etc .. and send them every 4hrs say (configurable)

This would help the mods even when #968 is implemented and also not overwhelm discord with all the flagged messages.

@andreaskoepf what is your opinion?

notmd:
Thank, just thinking about moving it to celery too. But 4h is so long. I saw many times the tree grow so fast, new children might come in less than 1h. Also, notification means I want to send it immediately so the mods can take action quickly. Is it ok to have the celery tasks run every 5 min or something? Discord allow to send up to 5000 message per min or 10000 per 10 mins.

.github/workflows/deploy-to-node.yaml
```diff
@@ -64,6 +64,8 @@ jobs:
       STATS_INTERVAL_WEEK: ${{ vars.STATS_INTERVAL_WEEK }}
       STATS_INTERVAL_MONTH: ${{ vars.STATS_INTERVAL_MONTH }}
       STATS_INTERVAL_TOTAL: ${{ vars.STATS_INTERVAL_TOTAL }}
+      DISCORD_API_KEY: ${{ vars.DISCORD_API_KEY }}
+      DISCORD_CHANNEL_ID: ${{ vars.DISCORD_CHANNEL_ID }}
```
yk, line 68: these need to be secrets, not vars

backend/oasst_backend/utils/discord.py
```diff
@@ -0,0 +1,45 @@
+import requests
+from loguru import logger
+from oasst_backend.config import settings
+from oasst_backend.models.message import Message
+
+ROOT_ENDPOINT = "https://discord.com/api/v10"
+
+
+def send_new_report_message(
```
yk, line 9: maybe rename to `maybe_send_discord_report_message`

yk (changes requested):

melvinebenezer:
@notmd yes, I agree. 4hrs is long. Yes, you can have celery tasks run every 5 mins or lesser.

backend/oasst_backend/prompt_repository.py
melvinebenezer: Yes. we could enhance it later with another PR.

melvinebenezer (approved):

melvinebenezer:
Will need a follow-up PR for the celery task

yk (approved):
thank you

PR #2654, opened 2023-04-17
complete and optimize the zh-CN

moshizx:
The missing and new words have been complete, and the wrong or imperfect translation have been modified.


notmd (approved):
Thanks, cc @theblackcat102

PR #2656, opened 2023-04-17
lazily render message menu actions

notmd:
NextLink auto prefetch the page (I think because we are using root getInitialProps).


yk (approved):
Thank you

PR #2658, opened 2023-04-17
added a useEffect to give focus to Textarea when not sending and streamingResponse is complete

### LINKED ISSUE #2628: Chat requires clicking input text box again and again
When chatting (https://open-assistant.io/chat), you have to click the input text box after each message you type. The default mode for a chat application should be that you can just keep typing.

spencerwmiles:
Closes #2628


spencerwmiles:
> Thanks but we want to avoid lose focus of the input instead re-focus like this. Because some time user want to not focus the input. Also this will force input focus after regenerate response. Maybe remove the disable state on input will work

Great point. I'll retool and have this back up shortly.

PR #2664, opened 2023-04-17
Removed disabled attribute from textarea so chat can continue without…

### LINKED ISSUE #2628: Chat requires clicking input text box again and again

spencerwmiles:
Closes #2628


notmd:
Thank you, can you also add the loading check to the `sendPrompterMessage`? Since now there is no disabled state, so it's possible to send many messages without the check.

spencerwmiles:
@notmd Should be good to go now!

notmd (approved):
Thanks

PR #2683, opened 2023-04-17
Fix website component storybook

lgeiger:
This PR fixes a few issues with the storybook component rendering:
- Removes extra semicolons in JSX components that show up in the HTML
- Changes `getEnv()` to return an empty dictionary if no environment values are set as otherwise `<Header/>` would throw `Cannot read properties of undefined (reading 'CURRENT_ANNOUNCEMENT')` since storybook doesn't have access to `window.__env`
- Fix props passed to `<Sortable/>` which prevented the text to be properly displayed


notmd (approved):
Thanks

PR #2692, opened 2023-04-18
feat: add language Chinese

dengshenkk:


dengshenkk:
@yk @andreaskoepf should I do something next?

olliestanley:
Can we move this into `docs/` directory please? Will make it easier to manage if we end up with many different translations

olliestanley:
Hi, thanks for your PR. We have decided to handle doc translations via links to automated machine translations at the top of each readme, to reduce need for constant updates of docs.

PR #2699, opened 2023-04-18
Fix trollboard crash

notmd:
I haven't tested it yet since I'm not able to reproduce the crash locally. But by reading the code, it should work lol


yk (approved):
seems reasonable :)

PR #2702, opened 2023-04-18
create backend user when generate jwt

notmd:
Avoid flooding the backend log with user not found error

### DIFF
--- a/website/src/pages/api/auth/[...nextauth].ts
+++ b/website/src/pages/api/auth/[...nextauth].ts
@@ -137,7 +137,7 @@ const authOptions: AuthOptions = {
      * Update the user's role after they have successfully signed in
      */
     async signIn({ user, account, isNewUser }) {
-      if (isNewUser && account.provider === "email") {
+      if (isNewUser && account.provider === "email" && !user.name) {
         await prisma.user.update({
           data: {
             name: generateUsername(),
--- a/website/src/pages/api/auth/[...nextauth].ts
+++ b/website/src/pages/api/auth/[...nextauth].ts
@@ -138,6 +138,8 @@ const authOptions: AuthOptions = {
      */
     async signIn({ user, account, isNewUser }) {
       if (isNewUser && account.provider === "email" && !user.name) {
+        // only generate a username if the user is new and they signed up with email and they don't have a name
+        // although the name already assigned in the jwt callback, this is to ensure notthing breaks, and we should never reach here.
         await prisma.user.update({
           data: {
             name: generateUsername(),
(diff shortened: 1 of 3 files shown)


website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -222,9 +227,9 @@ export default function auth(req: NextApiRequest, res: NextApiResponse) {
              * in the frontend, when the user accepts the tos, we do a full refresh
              * which means this function will be called again.
              */
-            const { user_id, tos_acceptance_date } = await oasstApiClient.fetch_frontend_user(backendUser);
-            token.backendUserId = user_id;
-            token.tosAcceptanceDate = tos_acceptance_date;
+            const frontendUser = await oasstApiClient.upsert_frontend_user(backendUser);
```
Comment (notmd) [inline website/src/pages/api/auth/[...nextauth].ts:230]: the backend does a lazy upsert so we should be safe. see https://github.com/LAION-AI/Open-Assistant/blob/6ba240f590787b40d56b3d14e19fd38b9488c850/backend/oasst_backend/api/v1/frontend_users.py#L62

website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -200,6 +200,24 @@ export default function auth(req: NextApiRequest, res: NextApiResponse) {
           select: { name: true, role: true, isNew: true, accounts: true, id: true },
         });
 
+        if (!frontendUser) {
+          // should never reach here
+          throw new Error("User not found");
+        }
+
+        // TODO avoid duplicate logic with the signIn event
+        if (frontendUser.isNew && !frontendUser.name) {
```
Comment (AbdBarho) [inline website/src/pages/api/auth/[...nextauth].ts:209]: we already create a username in line 143?

Comment (notmd) [reply website/src/pages/api/auth/[...nextauth].ts:209]: Just to ensure everything still works, since we have no test for this...

website/src/pages/api/auth/[...nextauth].ts
Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts]: Oh sorry! I only saw this comment now.

website/src/pages/api/auth/[...nextauth].ts
Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts:209]: Unfortunately, I still don't see which edge case is being caught here, or why are we checking for the name in line 140? we had the same sign up process for ever, did anything regarding the sign up requirement change?

Comment (notmd) [reply website/src/pages/api/auth/[...nextauth].ts:209]: I'm not sure how next auth work internally, but it seem the `jwt` callback is called before the `signIn` event, so in that case the name is not available yet, but the backend requires a name so I have to assign it first. The check `!user.name` on line 140 is to make sure we don't assign the name again

Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts:209]: ok I see, so the reason is the upsert, can you please comment that in the code, so we know why we did this?

Comment (notmd) [reply website/src/pages/api/auth/[...nextauth].ts:211]: Done

AbdBarho (approved):

PR #2713, opened 2023-04-18
Fullwidth comma was used by mistake for data field in Chinese translation.

lone-wolf-akela:
- Before this pull request:
image
and
image
- After this fix:
image
and
image


notmd (approved):
Thanks

Issue #2721, opened 2023-04-18
Docker | no matching manifest for linux/arm64/v8 in the manifest list entries


glebshkut:
When I'm running `docker compose --profile frontend-dev up --build --attach-dependencies` command in the root of the project
I get the following error:
```
[+] Running 0/1
 ⠴ db Pulling                                                         1.5s 
no matching manifest for linux/arm64/v8 in the manifest list entries
```
I'm using an Apple M1 chip, I tried to specify the platform in `docker-compose.yaml` file, it started composing, but later crashed

DominiqueGarmier:
you can add `platform: linux/x86_64` just below https://github.com/LAION-AI/Open-Assistant/blob/43a9d9c281e36f9640c26acb12be722f632c17ef/docker-compose.yaml#L11

Im unsure if this should be changed. Something might depend on it not being there. We could however create an extra docker-compose environment for M1

glebshkut:
Yeah, it helped. Thanks!

Should I remove this added line of code every time a do commits?

DominiqueGarmier:
I believe the pr should fix your problem. For the time being you will have to remove the line on every commit (or just `git remove docker-compose.yml`)

PR #2725, opened 2023-04-18
fix: added option to specify platform for db in compose

DominiqueGarmier:
related to issue https://github.com/LAION-AI/Open-Assistant/issues/2721

enables specifying platform for service `db` using environment variable `DB_PLATFORM`. If left unspecified (silently) defaults to current behaviour.

this solves the issue where M1 users where unable to use docker-compose as the `db` service relies on `ghcr.io/laion-ai/open-assistant/oasst-postgres` (which does not support `linux/arm64/v8` as plaform).

sadly (afaik) docker-compose does not offer something akin to `docker --platform` you have to set it in the compose-file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,9 @@ if you have problems):
 docker compose --profile ci up --build --attach-dependencies
 ```
 
+> **Note:** when running on MacOS with an M1 chip you have to use:
+> `DB_PLATFORM=linux/x86_64 docker compose ...`
+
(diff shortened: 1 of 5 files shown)


notmd (approved):
Thank you

melvinebenezer (approved):

andreaskoepf (approved):
ok, would be great if this solved the M1 docker issue

PR #2729, opened 2023-04-18
Update text_client_utils.py

mzamini92:
This implementation adds a new get_available_models() method to the DebugClient class, which retrieves the list of available model configurations from the API and returns a list of their names. The send_message() method then calls this method and checks if the provided model_config_name is in the list of available models. If it's not, a ValueError is raised with an appropriate error message.

### DIFF
--- a/inference/text-client/text_client_utils.py
+++ b/inference/text-client/text_client_utils.py
@@ -28,7 +28,18 @@ class DebugClient:
         self.message_id = None
         return self.chat_id
 
+    def get_available_models(self):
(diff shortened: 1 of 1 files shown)


AbdBarho:
we can probably get the model names once, no need to fetch every time.

mzamini92 (commented):
we can modify the `__init__` method to include the line `self.available_models = self.get_available_models()` so that the list of available models is retrieved only once during the instantiation of the DebugClient object:

AbdBarho (approved):

PR #2741, opened 2023-04-19
Rework chat layout UI

notmd:
The `QueueInfo` is removed for now. I will add it back in follow up PR. Most of the change are UI code, there is no logic here, it safe to merge
Desktop 
<details>
  <summary>Light mode</summary>
  
image
</details>
<details>
  <summary>Dark mode</summary>

image
</details>

Mobile
<details>
  <summary>Light mode</summary>
  
image</details>
<details>
  <summary>Dark mode</summary>

image</details>

### DIFF
--- a/website/src/components/Chat/ChatForm.tsx
+++ b/website/src/components/Chat/ChatForm.tsx
@@ -63,7 +63,7 @@ export const ChatForm = forwardRef<HTMLTextAreaElement, ChatFormProps>((props, r
           }}
           style={{ resize: "none" }}
         />
-        <Flex position="absolute" zIndex="10" className="right-0 top-0 h-full items-center px-4" gap="2">
+        <Flex position="absolute" zIndex="10" className="ltr:right-0 rtl:left-0 top-0 h-full items-center px-4" gap="2">
           <ChatConfigDrawer />
           {isSending ? (
             <CircularProgress isIndeterminate size="20px" />
--- a/website/src/components/Chat/ChatForm.tsx
+++ b/website/src/components/Chat/ChatForm.tsx
@@ -63,13 +63,7 @@ export const ChatForm = forwardRef<HTMLTextAreaElement, ChatFormProps>((props, r
           }}
           style={{ resize: "none" }}
         />
-        <Flex
-          position="absolute"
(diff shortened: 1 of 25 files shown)


AbdBarho:
Thank you, huge effort!

some stuff I noticed:

When refreshing a `/chat/id` page with some messages in it, I get a hydration error:
```
Warning: Extra attributes from the server: data-disabled
```

---

In dark mode, the chat list has an awkward dark box behind it
image

---

When you open an empty chat, there is a lot! of empty space below

image

Also on mobile

image


---

Pressing on the config icon in this layout will send the message
image

website/src/components/Chat/ChatSection.tsx
```diff
@@ -1,46 +1,38 @@
...
+    },
+  });
+
+  useIsomorphicLayoutEffect(() => {
+    const cache = getConfigCache();
+    const model = modelInfos.find((model) => model.name === cache?.model_config_name);
+    if (model && cache) {
+      form.reset(cache);
+    }
+  }, [form.reset, modelInfos]);
```
notmd, line 29: I have to use `useIsomorphicLayoutEffect` here to avoid hydration errors. The UI will flash a bit if they hard refresh the page

website/src/components/Chat/ChatForm.tsx
```diff
@@ -34,31 +35,43 @@ export const ChatForm = forwardRef<HTMLTextAreaElement, ChatFormProps>((props, r
...
+            boxShadow: "none",
+          }}
+          style={{ resize: "none" }}
+        />
+        <Flex position="absolute" zIndex="10" className="right-0 top-0 h-full items-center px-4" gap="2">
+          <ChatConfigDrawer />
+          {isSending ? (
+            <CircularProgress isIndeterminate size="20px" />
+          ) : (
+            <ChatInputIconButton icon={Send} onClick={handleSubmit}></ChatInputIconButton>
```
AbdBarho, line 71: the text here will go under the icons and looks a bit awkward

AbdBarho (approved):
We still need to work out the RTL stuff

image


I think this is okay for now, we can fix this later.

PR #2768, opened 2023-04-20
Clear chatbox input only after response it generated 

### LINKED ISSUE #2604: Delete chatbox text only after receiving confirmation that the request has been placed
it's quite annoying when we are under high load and the request 500s, but the chat box is cleared anyway, means I have to re-type everything.

bishowX:
Now the user will have better experience chatting with assistant. The chat box input is only cleared after response is generated. Fixes  #2604


notmd:
Thanks, but we want to clear the input after the prompter message is created. Can you update that?

bishowX:
> Thanks, but we want to clear the input after the prompter message is created. Can you update that?

I've jst updated the code @notmd

notmd (approved):
Thank you!

PR #2770, opened 2023-04-20
Better Chinese instructions to dashboard. 

yuechen-li-dev:
Current Chinese instructions on the dashboard are machine translated and pretty confusing to new people joining in, some of them have unintentionally messed up the data trees thinking it was the chat. 

I've tried to make the messaging clearer that the tasks are for dataset collection and not an actual chat.


notmd (approved):
Thank you

PR #2773, opened 2023-04-20
support pagination for chat list

### LINKED ISSUE #2584: Pagination in chat listing

notmd:
close #2584. ~~Doesn't support return null cursor when the chats in response contain first or last chat. So the frontend need to do 1 more request~~
Frontend not support this yet

### DIFF
--- a/inference/server/oasst_inference_server/routes/chats.py
+++ b/inference/server/oasst_inference_server/routes/chats.py
@@ -1,8 +1,9 @@
 import asyncio
+import base64
 
 import fastapi
 import pydantic
-from fastapi import Depends
+from fastapi import Depends, Query
 from loguru import logger
 from oasst_inference_server import auth, chat_utils, deps, models, queueing
 from oasst_inference_server.schemas import chat as chat_schema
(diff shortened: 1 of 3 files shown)


inference/server/oasst_inference_server/routes/chats.py
```diff
@@ -21,12 +22,47 @@
...
     logger.info("Listing all chats.")
-    chats = await ucr.get_chats(include_hidden=include_hidden)
+
+    def encode_cursor(chat: models.DbChat):
+        return base64.b64encode(chat.id.encode()).decode()
+
+    def decode_cursor(cursor: str | None):
+        if cursor is None:
+            return None
+        return base64.b64decode(cursor.encode()).decode()
```
AbdBarho, line 38: why do we base64 encode / decode the ids?

yk (approved):
thank you

inference/server/oasst_inference_server/routes/chats.py
notmd: Just to hide sensitive information from the eyes. Also it will prevent us from problems if the frontend forgot to url encode the cursor, not really necessary in this case.

PR #2774, opened 2023-04-20
Added max messages and max message length settings for inference

yk:
Returning 413 both when chat exceeds num messages allowed, and when the individual message is too long


olliestanley (approved):

andreaskoepf (approved):
ok, great .. should make things more stable.

PR #2786, opened 2023-04-20
Small docs fix links

andrewm4894:
- fix some links in docs to be relative so that they don't open new tabs when clicked and act just like nav clicks.


andrewm4894:
closing as messed up a bit

PR #2798, opened 2023-04-21
Added dataset of 100k c++ function annotations and processing scripts.

### LINKED ISSUE #2796: Dataset: 100k c++ function annotations
Brendan Dolan-Gavitt shared a dataset of 100K c++ functions with annotations and tags generated by GPT-3.5 turbo. 

https://twitter.com/moyix/status/1644355889602654210

For example:
```
int gg_set_focus_pos(gg_widget_t *widget, int x, int y) {
	return 1;
}

Q. What language is the above code written in?
A. C/C++.
Q. What is the purpose of the above code?
A.
This code defines a function called `gg_set_focus_pos` that takes a pointer to a `gg_widget_t` structure and two integer 
arguments `x` and `y`. The function returns an integer value of 1. The purpose of this function is not clear from this code 
snippet alone, as it does not contain any implementation details. However, based on the function name and arguments, it is 
likely that this function sets the focus position of a graphical user interface (GUI) widget to the specified `x` and `y` 
coordinates.

Tags: GUI, widget, focus, position, function.
```
I've generated an INSTRUCTION RESPONSE dataset with some prompt variety and uploaded it to HuggingFace. Hopefully, this would improve OA's code understanding/generation.

The processed dataset can be found here: https://huggingface.co/datasets/casey-martin/oa_cpp_annotate

casey-martin:
Added dataset of 100k c++ function and their summaries. Resolves #2796.

### DIFF
--- a/data/datasets/__init__.py
+++ b/data/datasets/__init__.py
@@ -23,6 +23,7 @@ INSTRUCTION_DATASETS = {
     "LogicInference_OA": "KK04/LogicInference_OA",
     "oa_dolly_15k": "OllieStanley/oa_dolly_15k",
     "poetry_instruction": "checkai/instruction-poems",
+    "oa_stackexchange": "donfu/oa-stackexchange",
 }
 
 SAFETY_DATASETS = {
--- a/data/datasets/oa_stackexchange/README.md
+++ b/data/datasets/oa_stackexchange/README.md
@@ -1,0 +1,279 @@
+---
+dataset_info:
+  features:
+    - name: INSTRUCTION
+      dtype: string
+    - name: RESPONSE
+      dtype: string
+    - name: SOURCE
+      dtype: string
+    - name: METADATA
+      struct:
+        - name: answer_score
+          dtype: int64
+        - name: question_score
+          dtype: int64
+        - name: tags
+          dtype: string
+  splits:
+    - name: train
+      num_bytes: 6549838664
+      num_examples: 6331083
+  download_size: 3755782987
+  dataset_size: 6549838664
+license: cc-by-sa-4.0
+language:
+  - en
+  - uk
+  - ru
+  - de
+  - fr
+  - it
+  - es
+pretty_name: Open-Assistant StackExchange Instruction
+---
+
+# Stackexchange Instructions for OpenAssistant
+
+This dataset is taken from https://archive.org/details/stackexchange.
+
+There's a single parquet file combining all stackexchange sites. The threads
+have been filtered as follows: only threads with an accepted answer, for which
+both the question and response is less than 1000 characters have been choosen.
+Other answers, or questions without accepted answers, or long entries have been
+droppped.
+
+Each row consists of
+
+- INSTRUCTION
+- RESPONSE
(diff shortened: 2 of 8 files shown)


olliestanley:
Thanks for submitting this! I will review tomorrow

casey-martin:
@olliestanley My pleasure. I had an idea to create an additional task where given the verbal description of the code, the LLM would need to generate the source c++ function. Essentially, it would turn the annotation task into a code generation task.

Would this be a desirable addition to the dataset?

data/datasets/oa_cpp_annotate/process_data.py
```diff
@@ -0,0 +1,93 @@
...
+import random
+
+import pandas as pd
+
+SPLIT_1 = "Q. What language is the above code written in?\nA. "
+# SPLIT_2 = "Q. What is the purpose of the above code?\nA."
+JSONL_FP = "ghdecomp_summaries_merged_small_openaisummary_combined.jsonl.bz2"
+
+PROMPTS = [
+    "{code}What language is the above code written in? What does the code do?\n\n",
```
olliestanley, line 11: The prompts which end with `\n\n` cause INSTRUCTION values ending with `\n\n` in the OA format dataset. I don't think this is desirable

casey-martin: Got it. I'll remove them and reupload the data. I'm also working on inverting the task to add some in code generation instructions. Want me to add it to this dataset, or do you think it would be better to leave this as a pure annotation task?

olliestanley: I think code generation tasks are in general useful. I am sceptical of "duplicating" data in training (e.g. having a Q-A pair where the model predicts code from description, and an inverted Q-A pair where the model predicts description from code, both derived from the same original description-code pair).

One approach which has been taken in the past (with poetry data and others) is to randomise this, such that half of the description-code pairs are used for one task and the other half for the other task.

casey-martin: I have similar qualms about duplicated data. My plan was to random assign a sample to either the code gen or code annotation task. I'm not sure what the dev team's priorities are in terms of dataset composition. My naive idea was something like a 70-30 split between annotation and generation as the crowd-sourced chat data already has a fair amount of code gen in it. Do you have any preferences?

olliestanley: > I have similar qualms about duplicated data. My plan was to random assign a sample to either the code gen or code annotation task. I'm not sure what the dev team's priorities are in terms of dataset composition. My naive idea was something like a 70-30 split between annotation and generation as the crowd-sourced chat data already has a fair amount of code gen in it. Do you have any preferences?

I will leave this up to you, it's true that there is much more code gen than annotation currently so something like the 70-30 split could make sense. It's hard to know what is "optimal" without training models to test

casey-martin: I'll get to work on this. Thank you for the feedback.

PR #2809, opened 2023-04-21
update entities

CloseChoice:
Add entities, which are refered in model/model_training/custom_datasets/formatting.py


model/model_training/custom_datasets/entities.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@unique
+class Language(str, Enum):
+    AB = "ab"  # Abkhazian
+    AA = "aa"  # Afar
+    AF = "af"  # Afrikaans
+    AK = "ak"  # Akan
+    SQ = "sq"  # Albanian
+    AM = "am"  # Amharic
```
andreaskoepf, line 17: Are these all BCP 47 lang-codes? Should document which standard is used, the human-demonstration collection backend uses BCP 47.  We might also think about using a lib like for example https://pypi.org/project/langcodes/ ?

CloseChoice: This is the iso639-1 standard.

andreaskoepf: ok, then we need to check and change this.. because they should match the lang codes of the backend, right? There we use BCP 47: https://en.wikipedia.org/wiki/IETF_language_tag

andreaskoepf: Here is a result directly from our DB:
```
postgres=# select distinct lang from message;
 lang  
-------
 ar
 bg
 bn
 ca
 cs
 da
 de
 el
 en
 eo
 es
 eu
 fa
 fi
 fr
 gl
 he
 hu
 id
 it
 ja
 ko
 nb-NO
 nl
 pl
 pt-BR
 ro
 ru
 sk
 sv
 th
 tr
 uk-UA
 vi
 zh
(35 rows)
```

andreaskoepf: To give some background why this is necessary: Portuguese is VERY different in Brazil and Portugal .. much more different than for example British and American English .. native speakers clearly explained to us hat these two languages need to be separated.

CloseChoice: changed it to langcodes

andreaskoepf (approved):
lgtm! :-)

PR #2811, opened 2023-04-21
Gracefully handle user does not exist in Inference

AbdBarho:
Much less hacky hack than the one we had before, the inference log should also be much cleaner now.


yk (approved):
Still a bit hacky, but should do it :) thank you

PR #2813, opened 2023-04-21
Add poem instruction ds

CloseChoice:
I checked the poem instruction for empty strings, references and appearances of "openai". As expected for a poem dataset, none of these things was found.


andreaskoepf (approved):
fantastic!

PR #2818, opened 2023-04-21
add check dataset appearances, update vicuna to dataset entry

CloseChoice:
- add a script which can be used to check if any dataset consists specific regular expressions or words
- update vicuna so that it can be used with the new `DatasetEntry` class
- remove single references from vicuna (so `[1]` is removed, but I found with the script mentioned above that there are a couple of occurances where our `re_reference_remove` regex hits, but it is actually a list (for language like e.g. python), therefore just remove single references)
- if human response is none, then sample from multiple answers like `['please continue', '...']`

This PR depends on https://github.com/LAION-AI/Open-Assistant/pull/2809

### DIFF
--- a/model/model_training/custom_datasets/qa_datasets.py
+++ b/model/model_training/custom_datasets/qa_datasets.py
@@ -501,10 +501,6 @@ class Vicuna(Dataset):
         for line in data["conversations"]:
             speaker = line["from"]  # 'human' or 'gpt'
             message = line["value"]
-            # if "You are a profesional software development assistant." in message:
-            #     import pdb; pdb.set_trace()
-            # if re_reference_remove.match(message):
-            #     import pdb; pdb.set_trace()
             if message is None or message == "":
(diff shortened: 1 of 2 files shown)


andreaskoepf (approved):
Super nice that we have this dataset checker tool now!

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -486,35 +488,49 @@ class Vicuna(Dataset):
...
         for line in data["conversations"]:
             speaker = line["from"]  # 'human' or 'gpt'
             message = line["value"]
-
+            if message is None or message == "":
+                if speaker == "gpt":
+                    return None
+                elif speaker == "human":
+                    # replace empty messages with one of the following
+                    message = random.choice(["...", "Please continue", "Go on", ""])
```
andreaskoepf, line 509: nice!

PR #2820, opened 2023-04-21
Add inference backend support for data opt out

olliestanley:
Currently there is no functionality to change as we don't have a mechanism of exporting chats.


yk (approved):
amazing, thank you!

andrewm4894:
wondering if something like an additional `allow_data_publish` might be useful too if we ever wanted to let users also opt in to publish the data in any additional public datasets.

so allowing OA to use for training is one thing, but allowing OA to also publish this chat as part of any open datasets as another thing. 

just thinking out loud, can add later as another PR if ends up being something of use - unsure if there is plans to actually publish a subset of "donated" or "public" chats?

olliestanley:
> wondering if something like an additional `allow_data_publish` might be useful too if we ever wanted to let users also opt in to publish the data in any additional public datasets.
> 
> so allowing OA to use for training is one thing, but allowing OA to also publish this chat as part of any open datasets as another thing.
> 
> just thinking out loud, can add later as another PR if ends up being something of use - unsure if there is plans to actually publish a subset of "donated" or "public" chats?

I think the plan is to do this already (i.e, to allow data use is to allow anonymised data publication) but maybe I'm wrong about this, not sure

PR #2821, opened 2023-04-21
update api docs

andrewm4894:
update api docs to be latest


olliestanley:
Does this now also include the inference server? If not, how hard would it be to add that?

andrewm4894:
Hmm, no would not include it. Should be easy enough if inference server also using fastapi. I think docusaurus would be able to handle multiple openapi.json files. 

Will look into it. 

Actually maybe I'll do that as a follow on PR to be cleaner.

olliestanley:
> Hmm, no would not include it. Should be easy enough if inference server also using fastapi. I think docusaurus would be able to handle multiple openapi.json files.
> 
> Will look into it.
> 
> Actually maybe I'll do that as a follow on PR to be cleaner.

Sounds good, I will approve this

olliestanley (approved):

PR #2826, opened 2023-04-21
add blog post with sample of recent yt videos

andrewm4894:
seeing quite a few YT videos on the project following recent launch.

made a yt playlist and added some in a blog post for people to discover if interested.


olliestanley:
Cool, thanks!

olliestanley (approved):

PR #2829, opened 2023-04-22
Add hash_codec.py

umbra-scientia:
Several users obtained incorrect results when patching their LLaMA models with the Open Assistant deltas, due to the simple XOR format being highly sensitive to specific versions of Python, PyTorch, HuggingFace, and likely many other factors.

This script has therefore been re-designed using an encoding format based on cryptographic hashes, to robustly apply the Open Assistant delta patches to all known versions, encodings, and packagings of the LLaMA weights.

### DIFF
--- a/scripts/xor-codec/hash_codec.py
+++ b/scripts/xor-codec/hash_codec.py
@@ -1,0 +1,91 @@
+import os
+import sys
+import numpy
+import hashlib
+
+valid_hashes = [
+	# PyTorch Checkpoint V1:
+	"f856e9d99c30855d6ead4d00cc3a5573",
+	"d9dbfbea61309dc1e087f5081e98331a",
+	"2b2bed47912ceb828c0a37aac4b99073",
+	"ea0405cdb5bc638fee12de614f729ebc",
+	"4babdbd05b8923226a9e9622492054b6",
+	# PyTorch Checkpoint V2:
+	"d9dbfbea61309dc1e087f5081e98331a",
+	"f856e9d99c30855d6ead4d00cc3a5573",
+	"4babdbd05b8923226a9e9622492054b6",
+	"ea0405cdb5bc638fee12de614f729ebc",
+	"2b2bed47912ceb828c0a37aac4b99073",
+	# HuggingFace Checkpoint V1:
+	"d0e13331c103453e9e087d59dcf05432",
+	"29aae4d31a0a4fe6906353001341d493",
+	"b40838eb4e68e087b15b3d653ca1f5d7",
+	"f845ecc481cb92b8a0586c2ce288b828",
+	"f3b13d089840e6caf22cd6dd05b77ef0",
+	"12e0d2d7a9c00c4237b1b0143c48a05e",
+	"1348f7c8bb3ee4408b69305a10bdfafb",
+	"aee09e21813368c49baaece120125ae3",
+	"eeec4125e9c7560836b4873b6f8e3025",
+	"598538f18fed1877b41f77de034c0c8a",
+	"fdb311c39b8659a5d5c1991339bafc09",
+	"b77e99aa2ddc3df500c2b2dc4455a6af",
+	"edd1a5897748864768b1fab645b31491",
+	"6b2e0a735969660e720c27061ef3f3d3",
+	# HuggingFace Checkpoint V2:
+	"5cfcb78b908ffa02e681cce69dbe4303",
+	"3eddc6fc02c0172d38727e5826181adb",
+	"fecfda4fba7bfd911e187a85db5fa2ef",
+	"6b2e0a735969660e720c27061ef3f3d3",
+	"d566f98b28bf74ce6b35d5b20a651b27",
+	"148bfd184af630a7633b4de2f41bfc49",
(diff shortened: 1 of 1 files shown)


scripts/xor-codec/hash_codec.py
```diff
@@ -0,0 +1,91 @@
...
+	# PyTorch Checkpoint V1:
+	"f856e9d99c30855d6ead4d00cc3a5573",
+	"d9dbfbea61309dc1e087f5081e98331a",
+	"2b2bed47912ceb828c0a37aac4b99073",
+	"ea0405cdb5bc638fee12de614f729ebc",
+	"4babdbd05b8923226a9e9622492054b6",
+	# PyTorch Checkpoint V2:
+	"d9dbfbea61309dc1e087f5081e98331a",
+	"f856e9d99c30855d6ead4d00cc3a5573",
+	"4babdbd05b8923226a9e9622492054b6",
```
andreaskoepf, line 16: Is there any reason to include these duplicate hashes (they are all the same as Checkpoint V1)?

scripts/xor-codec/hash_codec.py
```diff
@@ -0,0 +1,91 @@
...
+	"ea0405cdb5bc638fee12de614f729ebc",
+	"2b2bed47912ceb828c0a37aac4b99073",
+	# HuggingFace Checkpoint V1:
+	"d0e13331c103453e9e087d59dcf05432",
+	"29aae4d31a0a4fe6906353001341d493",
+	"b40838eb4e68e087b15b3d653ca1f5d7",
+	"f845ecc481cb92b8a0586c2ce288b828",
+	"f3b13d089840e6caf22cd6dd05b77ef0",
+	"12e0d2d7a9c00c4237b1b0143c48a05e",
+	"1348f7c8bb3ee4408b69305a10bdfafb",
```
andreaskoepf, line 26: I am completely confused now. What are these hashes? they look like the ones we have in the readme as outputs of the LLaMA -> HF conversion process..

scripts/xor-codec/hash_codec.py
```diff
@@ -0,0 +1,91 @@
...
+	"b77e99aa2ddc3df500c2b2dc4455a6af",
+	"edd1a5897748864768b1fab645b31491",
+	"6b2e0a735969660e720c27061ef3f3d3",
+	# HuggingFace Checkpoint V2:
+	"5cfcb78b908ffa02e681cce69dbe4303",
+	"3eddc6fc02c0172d38727e5826181adb",
+	"fecfda4fba7bfd911e187a85db5fa2ef",
+	"6b2e0a735969660e720c27061ef3f3d3",
+	"d566f98b28bf74ce6b35d5b20a651b27",
+	"148bfd184af630a7633b4de2f41bfc49",
```
andreaskoepf, line 40: But these ones here are completely different from the ones in our Readme ..

scripts/xor-codec/hash_codec.py
```diff
@@ -0,0 +1,91 @@
...
+	"e1dc8c48a65279fb1fbccff14562e6a3",
```
andreaskoepf, line 41: "e1dc8c48a65279fb1fbccff14562e6a3" was for example reported as hash of ./pytorch_model-00003-of-00007.bin after HF conversion here: https://huggingface.co/OpenAssistant/oasst-sft-6-llama-30b-xor/discussions/5 ?

but their reported "5cfcb78b908ffa02e681cce69dbe4303" for ./pytorch_model-00002-of-00007.bin is not here in the list?

scripts/xor-codec/hash_codec.py
```diff
@@ -0,0 +1,91 @@
...
+	"fdb311c39b8659a5d5c1991339bafc09",
+	"eeec4125e9c7560836b4873b6f8e3025",
```
andreaskoepf, line 43: These ones match the hashes reported in https://huggingface.co/OpenAssistant/oasst-sft-6-llama-30b-xor/discussions/5#64432684dad68e008d0f14b7
.. but they were unable to load the model afterwards?

umbra-scientia: From what I can tell, some users ended up with different files after conversion to HuggingFace format, and so the simple XOR patch resulted in an incorrect file.

scripts/xor-codec/hash_codec.py
umbra-scientia: Several users, at least 3, ended up with these hashes after HuggingFace conversion, despite having matching hashes in the unconverted files.

scripts/xor-codec/hash_codec.py
umbra-scientia: The hashes from the readme were included verbatim in order to maintain compatibility for users ending up with these hashes.

scripts/xor-codec/hash_codec.py
umbra-scientia: It's fine to eliminate. The entire hash list could be deduplicated without breaking the program.

umbra-scientia:
Closing this since we've ironed out the glitch with the XOR version.

PR #2831, opened 2023-04-22
Discord bot: Pythia model

MrlolDev:
Changes:

- [x] Added pythia model to discord bot
- [x] Added a button in the answers that shows the model used
- [x] Added a selector when click the model button to change the model


discord-bots/oa-bot-js/src/commands/chat.ts
```diff
@@ -23,10 +23,16 @@ export default {
         .setName("model")
         .setDescription("The model you want to use for the AI.")
         .setRequired(false)
-        .addChoices({
-          name: "OA_SFT_Llama_30B",
-          value: "OA_SFT_Llama_30B",
-        })
+        .addChoices(
+          {
+            name: "OA_SFT_Llama_30B",
```
andreaskoepf, line 28: It would be best if the list of available models in the future would be requested at the inference-server (/models endpoint). It will be constantly changing .. you were probably irritated that so many are displayed in the staging environment .. but that's really only a "problem" in staging.

andreaskoepf (approved):
thanks a lot! :-)

PR #2839, opened 2023-04-22
Implement auto scroll when stream message

notmd:
After trying a bunch of methods I found this is the easiest way to do it. Other approaches are way more complicated.
Also found this lib https://github.com/compulim/react-scroll-to-bottom OAI also uses it, but it doesn't support SSR and unmaintained for years. 
This solution is not really reliable, if you paste a prompt with many lines (4 or above), it doesn't work. It would be if someone can patch the `react-scroll-to-bottom` to support SSR.

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Box, useBoolean, useToast } from "@chakra-ui/react";
-import { memo, useCallback, useEffect, useRef, useState } from "react";
+import { KeyboardEvent, memo, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { UseFormGetValues } from "react-hook-form";
 import SimpleBar from "simplebar-react";
 import { useMessageVote } from "src/hooks/chat/useMessageVote";
@@ -200,21 +200,7 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
     [createAndFetchAssistantMessage, isSending, setIsSending]
   );
 
-  const messagesEndRef = useRef<HTMLDivElement>(null);
-  const chatContainerRef = useRef<HTMLDivElement>(null);
-
-  useEffect(() => {
-    // if (!chatContainerRef.current || !messagesEndRef.current) return;
-    const { scrollTop, scrollHeight, clientHeight } = chatContainerRef.current;
-    // const threshold = 100;
-    // const scrollBottom = scrollHeight - scrollTop - clientHeight;
-
-    // if (scrollBottom > threshold) {
-    //   return;
-    // }
-
-    messagesEndRef.current.scrollIntoView({ behavior: "smooth" });
-  }, [messages, streamedResponse]);
+  const { messagesEndRef, scrollableNodeProps } = useAutoScroll(messages, streamedResponse);
 
   return (
     <Box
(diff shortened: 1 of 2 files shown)


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -200,6 +200,22 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
     [createAndFetchAssistantMessage, isSending, setIsSending]
   );
 
+  const messagesEndRef = useRef<HTMLDivElement>(null);
+  const chatContainerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!chatContainerRef.current || !messagesEndRef.current) return;
+    const { scrollTop, scrollHeight, clientHeight } = chatContainerRef.current;
+    const threshold = 100;
```
notmd, line 209: This is a magic number, I tested on different screen widths and found this is the most optimal one lol

AbdBarho (commented):
I would probably keep this very simple and just add

```js
messagesEndRef.current.scrollIntoView();
```
in the `onToken` handler after the new frame, and for good measure one final time after the streaming has finished.


I personally prefer without the smooth scrolling, because the scrolling speed is inconsistent depending on whether you have a new line or not, but that's a preference.

notmd:
> I would probably keep this very simple and just add
> 
> ```js
> messagesEndRef.current.scrollIntoView();
> ```
> 
> in the `onToken` handler after the new frame, and for good measure one final time after the streaming has finished.
> 
> I personally prefer without the smooth scrolling, because the scrolling speed is inconsistent depending on whether you have a new line or not, but that's a preference.

Your solution is great but here are a few things I want to support for better UX
- Auto scroll on first render from serverside
- Only auto scroll when the user scrolls to bottom (with a threshold is ok but 0 threshold is much better).

I can't find any simple way to support both of these. Only the `react-scroll-to-bottom` is perfect but it doesn't support SSR...

notmd:
Maybe we should go with my solution, remove the threshold check and ignore the UX...lolll

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -200,6 +200,22 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
     [createAndFetchAssistantMessage, isSending, setIsSending]
   );
 
+  const messagesEndRef = useRef<HTMLDivElement>(null);
+  const chatContainerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!chatContainerRef.current || !messagesEndRef.current) return;
```
AbdBarho, line 207: Do we need `messagesEndRef`?

website/src/components/Chat/ChatConversation.tsx
AbdBarho: We can just always scroll to the bottom, OpenAI does it?

notmd:
@AbdBarho I just switch to the new implementation, works much better now

website/src/components/Chat/ChatConversation.tsx
notmd: Yes, the lib they use already handles that case.

andreaskoepf (approved):

7ombie:
Nice! I didn't see this when I opened #2880.

This feature is a convenience, so it's not the end of the world if it handles corner-cases imperfectly. Just having the UI generally keep up with the output will make it much nicer to use.

PR #2840, opened 2023-04-22
Add LoRA and Prefix-Tuning as Modeling Options for Improved Memory Efficiency + performance (potentially)

jordiclive:
This PR adds LoRA and prefix-tuning as modelling options (training and sampling code).

Both have shown strong performance and can outperform fine-tuning. They also can protect against the catastrophic forgetting problem which is important for chatbots. They keep the whole language model frozen so they can be distributed freely independent of the base language model.

They also allow much more memory-efficient training as there is no need for the optimizer states of the base model.

Benefits of LoRA
-  Can run only DS 2 with 30B.
- Run 65B without intense CPU usage
- Less overfitting, maintains performance on datasets. 
- Only need to share/push to hub the OS component (small file in Mbs)

— See [Andrej Karpathy (OpenAI)](https://twitter.com/karpathy/status/1649127655122550784?s=20) comment
— See purported [google leak](https://www.semianalysis.com/i/119223672/lora-is-an-incredibly-powerful-technique-we-should-probably-be-paying-more-attention-to)

Implementation Details:
- Explicitly set input_ids as a keyword argument in the sampling code to ensure proper functionality with PEFT `generate.`
-  Manually enable gradients for input to leverage gradient checkpointing effectively, as frozen embeddings would otherwise prevent gradient attachment.
- Include saving code from `pytorch_model.bin` for special tokens. Although these tokens are randomly initialized, they must be stored and saved as an external module since PEFT parameters learn to utilize them. Making them trainable is an option, but it is unlikely to make a significant difference.
- Integrate prefix-tuning, a powerful technique that involves stacking keys and values, by incorporating custom LLama modelling code. (Training speed lower than LoRA in my initial tests and performance similar/worse).

### DIFF
--- a/model/model_training/configs/config.yaml
+++ b/model/model_training/configs/config.yaml
@@ -81,6 +81,8 @@ defaults:
   per_digit_tokens: false
   is_reward_model: false
   deepspeed_config: configs/zero_config.json
+  peft_model: false
+  peft_type: "lora"
 
 webgpt_dataset_only:
   datasets:
@@ -353,6 +355,30 @@ llama-30b-pretrain:
   save_steps: 500
   num_train_epochs: 1
   save_total_limit: 2
+
+lora-llama-13b:
+  dtype: fp16
+  log_dir: "llama_log_7b"
+  learning_rate: 5e-5
+  model_name: "/admin/home-jordiclive/llama/7B"
+  output_dir: llama_model
+  weight_decay: 0.0
+  max_length: 2048
+  warmup_steps: 300
+  gradient_checkpointing: true
+  gradient_accumulation_steps: 1
+  per_device_train_batch_size: 24
+  per_device_eval_batch_size: 5
+  eval_steps: 5
+  num_train_epochs: 12
+  save_total_limit: 2
+  save_strategy: epoch
+  use_flash_attention: True
+  residual_dropout: 0.0
+  deepspeed_config: configs/zero_config.json
+  peft_model: true
+  peft_type: "lora"
+  use_custom_sampler: true
 
 pythia-70m-deduped:
   learning_rate: 8e-6
--- a/model/model_training/configs/config.yaml
+++ b/model/model_training/configs/config.yaml
@@ -358,9 +358,9 @@ llama-30b-pretrain:
 
 lora-llama-13b:
   dtype: fp16
-  log_dir: "llama_log_7b"
+  log_dir: "llama_lora_log_13b"
   learning_rate: 5e-5
-  model_name: "/admin/home-jordiclive/llama/7B"
+  model_name: "/admin/home-jordiclive/llama/13B"
   output_dir: llama_model
   weight_decay: 0.0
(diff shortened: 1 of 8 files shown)


smeyerhot:
@jordiclive would this PR make it possible to load a PEFT model for inference in the chat?

jordiclive:
This code is currently just for model training and evaluation. But should be trivial to load it for inference, it uses the same HF generate method.

jordiclive:
@andreaskoepf I am going to run a 30B LoRA model just on the sft datasets and will post the sampling report.

model/model_training/configs/config.yaml
```diff
@@ -354,6 +356,78 @@ llama-30b-pretrain:
   num_train_epochs: 1
   save_total_limit: 2
 
+lora-llama-13b:
```
smeyerhot, line 359: Should this be lora-llama-7b?

model/model_training/configs/config.yaml
```diff
@@ -354,6 +356,78 @@ llama-30b-pretrain:
...
+  use_flash_attention: True
+  residual_dropout: 0.0
+  deepspeed_config: configs/zero_config.json
+  peft_model: true
+  peft_type: "lora"
+  use_custom_sampler: true
+
+lora-llama-30b:
+  dtype: fp16
+  log_dir: "llama_log_7b"
```
smeyerhot, line 385: Should these `log_dir`s  be changed to reflect model size?

model/model_training/models/peft_modeling.py
```diff
@@ -0,0 +1,101 @@
...
+@dataclass
+class SaveLoraConfig:
+    dtype: torch.dtype = torch.float16
+    is_reward_model: bool = False
+    quantization: bool = False
+    seq2seqmodel: bool = False
+    freeze_layer: bool = False
+    residual_dropout: float = 0
+    use_flash_attention: bool = False
+    adapter_save_path: str = "adapter_13B_new"
```
smeyerhot, line 80: Does this only work on 13B model? Should it be hardcoded?

model/model_training/configs/config.yaml
jordiclive: No sorry, I just didn't bother changing the logging dirs. Will change the log_dir names.

jordiclive: it is meant to be 13B . you can fit much higher bs. the model_name is a mistake (I had some code to use the 7B model as a tokenizer for all models, and download the weights from HF)

model/model_training/models/peft_modeling.py
jordiclive: it's the output path for a directory with the adapter weights in them. I've changed to just "adapter"

jordiclive: @smeyerhot I've updated the configs.

andreaskoepf (approved):
Super great to have this!

PR #2843, opened 2023-04-22
Fix saving model between chats

AbdBarho:
You can see this bug on dev or staging if you try changing the model  couple of times.

Somewhere within the mess of server side / client side rendering the model config gets overwritten.

### DIFF
--- a/website/src/components/Chat/ChatConfigSaver.tsx
+++ b/website/src/components/Chat/ChatConfigSaver.tsx
@@ -4,16 +4,13 @@ import { ChatConfigFormData } from "src/types/Chat";
 import { setConfigCache } from "src/utils/chat";
 
 export const ChatConfigSaver = () => {
-  const { watch, formState, reset } = useFormContext<ChatConfigFormData>();
+  const { watch, formState } = useFormContext<ChatConfigFormData>();
   const config = watch();
   useEffect(() => {
-    // only update when form is changed
     if (formState.isDirty) {
       setConfigCache(config);
-      // unset isDirty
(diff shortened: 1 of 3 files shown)


website/src/components/Chat/ChatConfigSaver.tsx
```diff
@@ -4,14 +4,11 @@ import { ChatConfigFormData } from "src/types/Chat";
...
+  const { watch } = useFormContext<ChatConfigFormData>();
   const config = watch();
   useEffect(() => {
-    // only update when form is changed
-    if (formState.isDirty) {
-      setConfigCache(config);
-    }
-  }, [config, formState.isDirty]);
+    setConfigCache(config);
+  }, [config]);
```
AbdBarho, line 11: when we change presets, we use the function `reset`, which sets the `isDirty` to false, so we are not saving.

notmd: The isDirty check is require to avoid update on the first render which use the first model config from serverside, not the one from localstorage

notmd: We should pass keepDirty option when update the preset:
```js
reset({ ...config, model_config_name: selectedModel }, {
     keepDirty: true,
});
```

AbdBarho: I refactored it, I added keep dirty, and switched to using `Controller` instead of `register` and it works now (hopefully)

notmd (approved):

PR #2848, opened 2023-04-23
Add StackExchange dataset. Resolves #191

### LINKED ISSUE #191: harvest Stack Exchange Q&A data
https://stackexchange.com/sites

After reading a discussion here about how to build a model that detects "instruction-like" conversations on Twitter, I was wondering about data sources that are in instruction format by definition. I found no discussion on Stack Exchange in this repo. So this thread is to explore the possibility of using data from the numerous Stack Exchange sites.

stefnnn:
Here's an initial implementation of stackexchange data processing. The dataset can be found here: https://huggingface.co/datasets/donfu/oa-stackexchange

All 180+ stackexchange sites where taken into account. There's a variety of very interesting topics included beyond coding questions (see stats in readme).

The following filtering was done:
- Only questions with an accepted answer where taken into account
- Only one answer per question (the accepted one) was added
- Only Q/A pairs, for which both the Q and A are shorter than 1'000 characters (tbd)
- HTML was converted to markdown, http links where removed


olliestanley:
Looks really cool, thank you! I will get chance to properly review this later today

clstaudt:
Excited for this data source to be integrated. Especially in order to improve the performance as a coding assistant.

data/datasets/__init__.py
```diff
@@ -23,6 +23,7 @@
     "LogicInference_OA": "KK04/LogicInference_OA",
     "oa_dolly_15k": "OllieStanley/oa_dolly_15k",
     "poetry_instruction": "checkai/instruction-poems",
+    "os-stackexchange": "donfu/oa-stackexchange",
```
olliestanley, line 26: Is `os-stackexchange` here a typo of `oa-stackexchange`? In general we follow the rule of the dict key being the dataset directory name with underscores for separation, so maybe we should change both dict key and directory name to `oa_stackexchange`

data/datasets/stackexchange/stats.md
```diff
@@ -0,0 +1,186 @@
+- 3dprinting: 1,006
```
olliestanley, line 1: These statistics are duplicated (here and in `README.md`), I think it could make sense to just have them in one place to make easier to keep up to date in case the dataset or conversion process changes in future

olliestanley (commented):
The code looks good, I have left a couple of minor comments and once these are addressed I think we can get this in :)

data/datasets/__init__.py
stefnnn: Makes sense! Implemented it.

data/datasets/stackexchange/stats.md
stefnnn: Agreed, I'll just leave it in the README.

olliestanley:
Thank you for these changes! One more thing I have just thought of, can the readme be updated to contain a brief "running order" for the scripts so anyone can quickly understand the end-to-end process to rebuild the dataset? Apologies for missing this in first review

stefnnn:
Thanks @olliestanley for the review! Implemented your suggestions. What do you think about the filtering? Is it fine to omit questions / answers > 1'000 characters, I guess this way almost all qa-pairs should fit the context length of 2048 even with a small prompt. Also, we could of course further trim the dataset by only picking pairs with high scores. But specifically for lower-volume topics I think this would be a waste.

Lastly: there are of course many questions with multiple answers, which would make good candidates to train a reward model (see e.g. https://huggingface.co/blog/stackllama which made use of this). But I doubt that's the best approach to train a reward model. 

Anway, if there are suggestions on how to further improve the dataset, let me know. We can also merge this and then do a 1.1 version later on.

olliestanley:
> Thanks @olliestanley for the review! Implemented your suggestions. What do you think about the filtering? Is it fine to omit questions / answers > 1'000 characters, I guess this way almost all qa-pairs should fit the context length of 2048 even with a small prompt. Also, we could of course further trim the dataset by only picking pairs with high scores. But specifically for lower-volume topics I think this would be a waste.

I think we are very safe by including only these. Remember 2048 context length is measured in tokens not characters, so we have a lot of headroom. Some investigation of the original dataset into whether short answers are "lower quality" than longer answers might be interesting, as well as how many questions we would add in going from (for e.g.) limit 1000 to 1500. But these questions can be addressed after this PR is merged.

> Lastly: there are of course many questions with multiple answers, which would make good candidates to train a reward model (see e.g. https://huggingface.co/blog/stackllama which made use of this). But I doubt that's the best approach to train a reward model.

This would definitely be interesting to consider. We would likely need a totally different dataset in different format for this, and it would then need a loader such as for example the data loading code for other RM datasets [here](https://github.com/LAION-AI/Open-Assistant/blob/main/model/model_training/custom_datasets/rank_datasets.py).

stefnnn:
Good point, I added documentation to the README on how to reproduce this step-by-step.

I'll see if I can do some analysis on score vs length of answers in different stackexchange topics.

olliestanley (approved):

PR #2856, opened 2023-04-23
Update verify-lobby.py

mzamini92:
Use `discord.Client.fetch_channel` instead of discord.utils.get to get the lobby channel object. This is a more efficient way to get a channel object by ID or name.

Use `discord.Role` objects instead of role names to check for roles. This is more reliable and efficient, and avoids potential issues with multiple roles having the same name.

Use async with to properly clean up the `tqdm` progress bar after it's done.

Use the `message.author.top_role` attribute instead of iterating over all roles to check for the unverified role. This is more efficient, and avoids potential issues with multiple roles having the same name.

### DIFF
--- a/scripts/discord/verify-lobby.py
+++ b/scripts/discord/verify-lobby.py
@@ -29,24 +29,20 @@ client = discord.Client(intents=intents)
 
 @client.event
 async def on_ready():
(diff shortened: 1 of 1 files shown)


notmd:
I think this script is no longer used. cc @yk

PR #2857, opened 2023-04-23
Add quantization/8bit model loading support for sampling_report.py

toiletpapercode:
Add ---quantize to the script call to take effect

Tested using:

1. --quantize --model-name OpenAssistant/oasst-sft-4-pythia-12b-epoch-3.5
2. --quantize --model-name t5-small --model-type t5conditional
3. --quantize --model-name OpenAssistant/stablelm-7b-sft-v7-epoch-3

Unable to test on llama models without access to the base weights (and/or only 35GB of VRAM?)

Enjoy,
TP

### DIFF
--- a/model/model_eval/manual/sampling_report.py
+++ b/model/model_eval/manual/sampling_report.py
@@ -226,6 +226,7 @@ def parse_args():
     parser.add_argument("--num-samples", type=int, default=2, help="number of sampling runs per configuration")
     parser.add_argument("--config", type=str, default="config/default.json", help="configuration file path")
     parser.add_argument("--half", action="store_true", default=False, help="use float16")
+    parser.add_argument("--quantize", action="store_true", default=False, help="use 8bit quantization")
     parser.add_argument("--skip-special-tokens", action="store_true", default=False)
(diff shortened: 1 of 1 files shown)


model/model_eval/manual/sampling_report.py
```diff
@@ -226,6 +226,7 @@ def parse_args():
     parser.add_argument("--num-samples", type=int, default=2, help="number of sampling runs per configuration")
     parser.add_argument("--config", type=str, default="config/default.json", help="configuration file path")
     parser.add_argument("--half", action="store_true", default=False, help="use float16")
+    parser.add_argument("--quantize", action="store_true", default=False, help="use 8bit quantization")
```
olliestanley, line 229: Can we call this `--int8`? Could be confusing as-is, as `--half` is also quantisation

toiletpapercode: Sure, I copied the flag name from `inference/worker/worker_full_main.sh` but I can commit a change to `--int8` for this script

andreaskoepf (approved):
lgtm, thnx!

PR #2863, opened 2023-04-23
Fix display name spill-over (#2851)

### LINKED ISSUE #2851: Display name spill-over
In isolated cases users misuse their freedom to use unicode display names to set bad names which spill over to other parts of the UI. Specifically affected are the leaderboard and trollboard pages which list users.

Use css styles to clip display names at the border of the surrounding cell's box.

scorbett123:
This fixes display names spilling over the rest of the ui.

Below is an image showing what the result looks like on the leader board.

image

Closes #2851

### DIFF
--- a/website/src/components/DataTable/DataTable.tsx
+++ b/website/src/components/DataTable/DataTable.tsx
@@ -199,7 +199,7 @@ const DataTableRow = <T,>({ row }: { row: Row<T> }) => {
       {renderCells.map((cell) => {
         const props = cell.column.columnDef.meta?.cellProps?.(cell);
         return (
-          <Td key={cell.id} {...props} colSpan={cell.span}>
+          <Td key={cell.id} {...props} colSpan={cell.span} style={{ overflow: "hidden" }}>
             {flexRender(cell.column.columnDef.cell, cell.getContext())}
           </Td>
         );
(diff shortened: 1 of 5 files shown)


website/src/components/DataTable/DataTable.tsx
```diff
@@ -199,7 +199,7 @@ const DataTableRow = <T,>({ row }: { row: Row<T> }) => {
       {renderCells.map((cell) => {
         const props = cell.column.columnDef.meta?.cellProps?.(cell);
         return (
-          <Td key={cell.id} {...props} colSpan={cell.span}>
+          <Td key={cell.id} {...props} colSpan={cell.span} style={{ overflow: "hidden" }}>
```
notmd, line 202: We dont want to add this bahavior on the DataTable component

notmd (commented):
Thank you

website/src/components/DataTable/DataTable.tsx
scorbett123: I don't entirely see why not, nor how you'd go about fixing this problem in places like the Users list which used data tables. Am I missing something, how else should I go about doing this?

johnflux: @scorbett123 
notmd is pointing out that you shouldn't put this use-case specific style in the general 'DataTable' component.

Either have a TdStyle property that you pass through, or simply set css property where you _use_ DataTable, like:

.mydatatable td { overflow: hidden }

notmd: There is a cellProps prop. You can use it to pass the style.

scorbett123: Thanks, is this what you're meaning?

scorbett123:
@andreaskoepf I think you've mistakenly applied the ml tag, unless I'm mistaken this should instead be website.

notmd (approved):
Thank you!

PR #2881, opened 2023-04-24
Add Hindi language to config

olliestanley:
This language was added as a translation but never added to the config file so it did not actually activate on the website.

### DIFF
--- a/website/next-i18next.config.js
+++ b/website/next-i18next.config.js
@@ -19,6 +19,7 @@ module.exports = {
       "fr",
       "gl",
       "he",
+      "hi",
       "hu",
       "id",
       "it",


UltraGamerine:
So addition of Hindi language in the config file did not directly lead to Hindi being present in the Language selection list on the website.

What more can we do for it ?

olliestanley:
> So addition of Hindi language in the config file did not directly lead to Hindi (हिन्दी) being present in the Language selection list on the website. (As far as I can see on the version just released 4 hrs ago, with changes mentioning this PR)
> 
> What more can we do for it ?

The release is not on production yet, hopefully in the next day or two it will be

PR #2901, opened 2023-04-25
Add Pythia SFT-4 model configs

olliestanley:
This will make setup slightly less confusing for the users who are trying to run these models with the OA web UI locally.


melvinebenezer (approved):
lgtm

PR #2926, opened 2023-04-26
Added some console logs to give feedback on why user is undefined

### LINKED ISSUE #2701: Investigate how `account` can be `undefined` here
https://github.com/LAION-AI/Open-Assistant/blob/6ba240f590787b40d56b3d14e19fd38b9488c850/website/src/lib/users.ts#L98

Cecileblbl:
Added a few logs of what I think could be causing an undefined user value.

closes #2701


website/src/lib/users.ts
```diff
@@ -95,7 +95,28 @@ export const getBatchFrontendUserIdFromBackendUser = async (users: { username: s
     const account = externalAccounts.find(
       (a) => a.provider === users[userIdx].auth_method && a.providerAccountId === users[userIdx].username
     );
-    // TODO check why the account is undefined
+    //NOTE: This part gives feedback to understand why user is undefined 
+    // Check if users array is empty
+    if (!users.length) {
+      console.log("Error: users array is empty.");
```
AbdBarho, line 101: Can you maybe add some prefix to all of these console log? just so we know that they came from here

AbdBarho (commented):
This would not close the issue, but will help us debug it, removed the link.

website/src/lib/users.ts
Cecileblbl: Thank you for reviewing, would something like : [getBatchFrontendUserIdFromBackendUser, Undefined user] work ?

Cecileblbl:
@notmd Yes of course I will update it

Cecileblbl:
I'm going to apply all this feedback

Issue #2877, opened 2023-04-24
Give Labelers the possibility to add new Languages & Dialects


Logophoman:
I would love to add samples of the Swabian german Accent "Schwäbisch" as a new Language to the LAION Assistant, but I don't know how to do so. It would be great if you could add this particular Dialect as a Language Option, since I know a few people that also would love to contribute. 

Please give labelers the possibility to add new Languages and Dialects.

olliestanley:
New languages and dialects can be added by creating a pull request adding the relevant translations, for e.g. https://github.com/LAION-AI/Open-Assistant/pull/1520

Logophoman:
@olliestanley made a PR! Thanks for the tip!

PR #2937, opened 2023-04-27
Add Swabian Dialect

Logophoman:
Added a PR as requested in #2877 

This is the Translation of the interface for the German dialect "Schwäbisch" 

According to [Wictionary  and the ISO 639-3](https://de.wiktionary.org/wiki/Hilfe:Sprachk%C3%BCrzel ) the shortcut is {swg}.

Here is the Language/Dialect Description page on Wikipedia: [Swabian German](https://en.wikipedia.org/wiki/Swabian_German)

I would love to see people contributing to this dialect and will be adding prompts and assistant replies as soon as it is pulled 😊


stefangrotz:
It might be worth to consider to choose a broader approach and choose the complete alemannic dialect group (iso: als) This would include badenian and also Swiss dialects, wich would increase the number of co tributors a lot.

Plus there is also an alemannic Wikipedia (https://als.m.wikipedia.org/) so there is at least some training data for the base model available while the number of mere svabian text is propably very small.

Logophoman:
@stefangrotz according to [this Wikipedia article](https://en.wikipedia.org/wiki/Alemannic_Wikipedia#Language) the language code `als` was used when there was yet no established distinction between the allemanic dialects.

I actually think, while a mixed approach will probably get more labelers and contributors generally involved in the system - I feel like for training the model it won't make that much of a difference overall if you have a solution that tries to include all dialects into one, but will rather cause massive conflicts between the labelers (since some swabian, badenian words and expressions are just different) and then stuff would get up/downvoted between swabian/badenian native speakers all the time, reducing overall quality. Since the model is multilingual anyway it will get i.e. a german prompt and then know that it's supposed to generate in german. Same will be the case for these dialects - if there is enough training data. 

The key is the multilinguality of our system -> It can still learn from closely related languages: 

If you for instance prompt in norwegian (I think there were ~100 messages trained when I checked) it will sometimes answer in english, norwegian, danish or swedish -> These languages are already closely related and there are not many messages currently. So the system cannot distinguish properly. However if you take close languages with more trees it becomes better and better.  But what this shows is that there is no point in a `one-size-fits-all` solution, since the model will generalize from everything that is related anyway. 

My norwegian friend could understand the answers it generated even tough stuff was mixed up with danish and swedish, but once more data exists the languages will become more distinct... 

-> I think it is better to have one tree per dialect (**maybe we generally need to think of a way of handling sub-languages i.e. all british/german dialects**) because then you `know` this is swabian,walser german, swiss german etc. And it also makes separating distinct dialects out of our training dataset easier (for instance if someone makes a badenian chatbot and takes out conflicting swiss and swabian accents that sound similar but are a bit different, but would reduce the badenian quality). 

So I really think if you can make a seperation between the dialects it really should be done, and yeah even tough 800.000 native speakers isn't too much i think it's a good experiment as well and I know a bunch of people that would get involved labeling. 

@stefangrotz I personally suggest that if you speak any allemanic dialects that you make a PR with these dialects added and their respective language code. 

I'm not a linguist but I think everything that is distinct enough to have one of these language codes is probably worth adding (If there are contributors that are willing to help).

According to the [Allemanic German](https://en.wikipedia.org/wiki/Alemannic_German#Status) article I think we would probably want to make PR's for all 4 distinct Allemanic Languages that used to be under the `als` code in the long run:

- Alemannic (badenian)/Swiss German/Alsatian - [gsw](https://iso639-3.sil.org/code/gsw)
- Swabian (this PR) - [swg](https://iso639-3.sil.org/code/swg)
- Walser German - [wae](https://iso639-3.sil.org/code/wae)
- Colonia Tovar German  - [gct](https://iso639-3.sil.org/code/gct)

Maybe we'll need to discuss more how we should handle dialects generally. Maybe the core team, i.e. @AbdBarho could join the discussion and potentially review this PR.

stefangrotz:
Okay, I see your points. If you belive that you can mobilize enough people to create a swabian dataset, then a separated language version would be fine for me. 

This could lead to a very long language list though, but this isn't necessarily a bad thing and Open Assistant is probably the only place where a dialect dialogue dataset can be built up right now.

Logophoman:
Yeah! I think in the long run we might want to maybe add something like a conditional dialects panel that shows up when dialects exist, especially if there are a ton of dialects that can be attributed to a language. Taking German as an example I think there is a ton of other dialects that could be added (Plattdeutsch, Sächsisch etc.) So maybe one could then choose `German` and then in another Dialects option `Swiss German`. I think the work during data export is an issue, but I mean having these dialects is something I would consider to be worth wile and I think it could also help to preserve these dialects better against decay and provide lots of value to i.e. old people that are stuck with their dialect 🤔

Also if you think for instance about the many Indian and Chinese dialects that are spoken by millions of people, having a good way of dealing with that could be genius for all the native speakers!

stefangrotz:
I believe a simple labeling system for both language variants and topics for specialists will be necessary at some point.

I also worked for the project Common Voice and there this is also an ongoing issue. Languages like Portuguese where the Brazilian variant is very different should be split up in a way, but doing it is hard. I think a labeling system ist the easiest solution.

Logophoman:
@stefangrotz that's good to know, awesome work! I think handling dialects won't ever be straightforward, especially since it is hard to determine how closely related a language or dialect really is from the original source and some are more closely related and others are further away 🤔 - I think fortunately Swabian is quite distinct and can be added as a separate language for now (most "normal" Germans barely understand it, if even a bit), but in the long run I think it's a good idea to discuss the way we want to integrate dialects, related languages and variants into the Open Assistant 👍 

@yk could you potentially review this discussion and merge this PR?

andreaskoepf:
@Logophoman thanks, could you quickly resolve the conflicts?

andreaskoepf (approved):
Jetzt brauchen wir nur noch mindestens fünf Schwaben, die mitmachen.. ;-)

PR #2942, opened 2023-04-27
load messages in client side

notmd:


AbdBarho (approved):
I think we should move to client side rendering soon, since our servers can't handle the load.

PR #2946, opened 2023-04-27
fix TypeError: Cannot read properties of undefined (reading 'session')

notmd:
see https://github.com/nextauthjs/next-auth/issues/4845
Not sure it fixed or not, since I'm not able to reproduce the issue locally. Also this might effect the performance, just a bit I think


yk (approved):
thank you!

AbdBarho (approved):

PR #2959, opened 2023-04-28
Handle chat error

notmd:
Handle error when create prompter or assistant message

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Box, useBoolean, useToast } from "@chakra-ui/react";
+import { AxiosError } from "axios";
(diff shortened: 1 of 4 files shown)


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -58,9 +59,21 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
...
-      const assistant_message: InferenceMessage = await post(API_ROUTES.CREATE_ASSISTANT_MESSAGE, {
-        arg: assistant_arg,
-      });
+      let assistant_message: InferenceMessage;
+      try {
+        assistant_message = await post(API_ROUTES.CREATE_ASSISTANT_MESSAGE, {
+          arg: assistant_arg,
+        });
+      } catch (e) {
+        if (e instanceof OasstError) {
```
notmd, line 68: This always is `OasstError` since we already transformed the error, the check to make TS happy

AbdBarho (approved):
Yes! we need more and better error loggin.

We could encapsulate the try catch logic better. That could be for a different PR.

PR #2961, opened 2023-04-28
Client side rendering (for most pages)

AbdBarho:
To reduce the load on our servers, move to client side for most of the pages.

This PR includes:
* fetching the env from the api once
* fetching the models once
* simplifying server-side code for the chat/id endpoint

I went through all the pages and clicked around, everything seems fine.


website/src/pages/_app.tsx
```diff
@@ -50,31 +47,25 @@ function MyApp({ Component, pageProps: { session, ...pageProps }, cookie, env }:
...
-    session,
-    env: {
-      ENABLE_CHAT: boolean(process.env.ENABLE_CHAT),
-      ENABLE_EMAIL_SIGNIN: boolean(process.env.ENABLE_EMAIL_SIGNIN),
-      ENABLE_EMAIL_SIGNIN_CAPTCHA: boolean(process.env.ENABLE_EMAIL_SIGNIN_CAPTCHA),
-      CLOUDFLARE_CAPTCHA_SITE_KEY: process.env.CLOUDFLARE_CAPTCHA_SITE_KEY,
-      CURRENT_ANNOUNCEMENT: process.env.CURRENT_ANNOUNCEMENT,
-      NODE_ENV: process.env.NODE_ENV,
-    },
+    ...(await App.getInitialProps(context)),
```
notmd, line 68: Can we remove the `getInitialProps` here?

AbdBarho: Uhm... it is recommended by nextjs, so I think we can leave it.

https://nextjs.org/docs/advanced-features/custom-app#caveats

I wanted to ask you about this: 
I had to remove the `session` stuff because it was throwing A LOT of errors when doing `npm run build`.

It seems that it was trying to fetch the session in build time.

do you have any idea how can we fix this?

notmd: I mean remove the `MyApp.getInitialProps`. Since we don't pass the the env, we don't it to use anymore

AbdBarho: if remember correctly, the screen flashes white everytime when we don't have this, since chakra will fall back to localstorage?

I can remove it nonetheless if you say so.

notmd: I think it sill flash anyway. Since the page is static, next doesn't have information when render the page. Charka docs https://chakra-ui.com/getting-started/nextjs-guide#color-mode-script recommend to use a custom scrip in the `_doc` page

AbdBarho: I see, we can add it back in a different PR.

notmd (approved):

PR #2969, opened 2023-04-29
update inference readme

Chuseuiti:
Fixed inference README with working version and deleted a lot of the extra variants information to simplify the document, as they are derivations of the same steps.

The motivation is that this container is not part of the docker compose, neither exist in the existing repository in the main branch:
    inference-text-client

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -5,6 +5,8 @@ services:
 
   # Use `docker compose --profile frontend-dev up --build --attach-dependencies` to start the services needed to work on the frontend. If you want to also run the inference, add a second `--profile inference` argument.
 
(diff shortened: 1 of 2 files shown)


olliestanley:
Hi. Could you also incorporate some of the instructions from #2595 into this? I think that PR is stale

Chuseuiti:
updated with the updates of the other PR @olliestanley

andrewm4894:
@Chuseuiti can you run `pre-commit run --all-files` and then add that commit, it looks like it just wants to clean some whitespace and end of file stuff.

olliestanley (approved):
Thanks!

PR #2971, opened 2023-04-29
Fix inference-server dockerfile so it can run locally

toiletpapercode:
@andreaskoepf: https://github.com/LAION-AI/Open-Assistant/pull/2956 Added an env `PROMETHEUS_MULTIPROC_DIR` which causes the container to fail to run locally due to the directory not existing.

Creating the directory after setting`USER ${APP_USER}` resolves this

### DIFF
--- a/docker/inference/Dockerfile.server
+++ b/docker/inference/Dockerfile.server
@@ -46,6 +46,8 @@ RUN adduser               \
 
 USER ${APP_USER}
 
+RUN mkdir -p "${PROMETHEUS_MULTIPROC_DIR}"
+
 WORKDIR ${APP_ROOT}
 
 


andreaskoepf:
hm. this is treating a symptom ... but why does it happen in the first place? E.g. there must be a problem with the main script then: https://github.com/LAION-AI/Open-Assistant/blob/5ce45f31ebf0a1c7b389b88c9755aea8393e8f9f/inference/server/server_main.sh#L7-L11 where I added deletion of the directory or unsetting of the env var?

andreaskoepf:
@toiletpapercode could you please share the error message that you get?

andreaskoepf (approved):

toiletpapercode:
@andreaskoepf That server_main.sh is only executed in the `FROM base-env as prod` block. And that doesn't happen when running locally via docker compose. On`FROM base-env as dev` it executes the uvicorn directly.

The error message was a simple directory not found.

PR #2976, opened 2023-04-29
add Red Pajama dataset.

jordiclive:
Add dataset class for continued LM objective during Fine-tuning.


jordiclive:
@andreaskoepf Is this what you had in mind? I think this might be enough.

I just need to work out the system prefix, as we don't want it for language modelling objective datasets.

andreaskoepf (approved):
yes, lgtm!

PR #2988, opened 2023-04-30
sft8 training preparation

andreaskoepf:
- add Utterance class to carry message properties (including role for debugging)
- separate SFT & RM DatasetEntry classes
- update unit tests
- fill DatasetEntry properties from OA dataset (quality, humor, creativity)
- add optional approximate length (~wordcount / 5) property

Effective "pre-training" data mix:
```
Creating sampler for datasets:
gpteacher_roleplay: 100.00% (1826)
red_pajama: 25.00% (186102)
wizardlm_70k: 100.00% (52224)
joke: 100.00% (358)
poem_instructions: 100.00% (13507)
oa_stackexchange: 10.00% (601439)
tell_a_joke: 100.00% (19000)
webgpt: 100.00% (16923)
gpt4all: 100.00% (388020)
alpaca_gpt4: 100.00% (50701)
code_alpaca: 100.00% (19020)
vicuna: 100.00% (46917)
oig_file: 100.00% (9500)
minimath: 100.00% (2736)
humaneval_mbpp_codegen_qa: 100.00% (561)
humaneval_mbpp_testgen_qa: 100.00% (561)
grade_school_math_instructions: 100.00% (8352)
recipes: 100.00% (4509)
cmu_wiki_qa: 100.00% (1529)
oa_wiki_qa_bart_10000row: 100.00% (9434)
prosocial_dialogue: 10.00% (15716)
explain_prosocial: 7.50% (27053)
soda: 25.00% (148900)
oa_leet10k: 100.00% (22236)
dolly15k: 100.00% (14264)

Epoch: 12,979 steps (batch_size: 128)
```


0x22almostEvil:
How about wikihow-16k? I fixed previous problems with bad grammar in prompts & images (now no images)

andreaskoepf:
> How about wikihow-16k? I fixed previous problems with bad grammar in prompts & images (now no images)

datasets are always a moving target ... we'll include that in the next SFT run. We have to check the quality of wikihow data .. if it is high we could also use it together with OA fintuning (phase 2 of training).

olliestanley (approved):
`oasst-data/` change LGTM (I have not reviewed `model/` code changes)

CloseChoice:
LGTM

CloseChoice (approved):

model/model_training/check_dataset_appearances.py
```diff
@@ -90,20 +142,35 @@ def iterate_over_dataset(ds):
...
-            pp.pprint(f"Found the following occurances in TRAIN {dataset_name}:")
-            pp.pprint(dict(matched_train))
-        if len(matched_val) != 0:
-            pp.pprint(f"Found the following occurances in VAL {dataset_name}:")
-            pp.pprint(dict(matched_val))
+        train_dct = {k: len(v) for k, v in matched_train.items()}
+        val_dct = {k: len(v) for k, v in matched_val.items()}
+        unified_keys = list(set(train_dct.keys()).union(set(val_dct.keys())))
+        unified_counts = {k: train_dct.get(k, 0) + val_dct.get(k, 0) for k in unified_keys}
+        import pdb
```
CloseChoice, line 158: NIT: Just a test script but maybe we should remove the debug statement here.

CloseChoice (approved):

model/model_training/check_dataset_appearances.py
andreaskoepf: ok, removed it! thanks for the review

andreaskoepf:
pre-commit failure is related to website dependencies (which were not touched by this PR).

model/model_training/check_dataset_appearances.py
```diff
@@ -5,23 +5,73 @@
...
+    "As an AI language model, I do not",
+    "I know as an AI language model you don't have",
+    "as an AI language model, you cannot",
+    "I'm sorry, but as an AI language model",
+    "As an AI language model, I don't have",
+    "I'm an AI ",
+    "I am an AI ",
+    "my name is OpenAI",
+    "trained by OpenAI",
+    "as an OpenAI language model",
```
dvruette, line 59: +1, maybe we could even consider removing any reference to OpenAI altogether

model/model_training/check_dataset_appearances.py
```diff
@@ -90,20 +142,32 @@ def iterate_over_dataset(ds):
...
+        if len(unified_counts):
+            overview_dct[dataset_name] = unified_counts
+            print(f"\nFOUND THE FOLLOWING APPEARANCES FOR DATASET {dataset_name}:")
+            pp.pprint(unified_counts)
+        # if len(matched_train) != 0:
+        #     pp.pprint(f"Found the following occurances in TRAIN {dataset_name}:")
+        #     pp.pprint(dict(matched_train))
+        # if len(matched_val) != 0:
+        #     pp.pprint(f"Found the following occurances in VAL {dataset_name}:")
+        #     pp.pprint(dict(matched_val))
```
dvruette, line 167: optional, but seems like this can be removed

model/model_training/check_dataset_appearances.py
```diff
@@ -5,23 +5,73 @@
...
-RE_TO_CHECK = [re_whitespace_newline_match, re_reference_remove, re_single_reference_remove]
-STRINGS_TO_CHECK = [*FILTER_BY_WORDS]
+from model_training.custom_datasets.formatting import DatasetEntryLm, DatasetEntrySft
+
+# from model_training.custom_datasets.qa_datasets import (
+#     re_reference_remove,
+#     re_single_reference_remove,
+#     re_whitespace_newline_match,
+# )
+# from model_training.custom_datasets.utils import FILTER_BY_WORDS
```
dvruette, line 31: optional, but seems like this can be removed

model/model_training/custom_datasets/formatting.py
```diff
@@ -25,140 +24,229 @@ def format_system_prefix(prefix, eos_token):
     )
 
 
-class PretrainDatasetEntry(BaseModel):
-    text: str | None = None
+def compute_length(s: str) -> int:
+    return len(re.findall(r"\w+", s)) // 5 + 1
```
dvruette, line 28: very arbitrary way of computing length. in the future i would suggest to just quantize to some predefined mesh (e.g. quantize to the nearest multiple of 10) and expose it in the system prompt as `less than 10 words, 10-20 words, ..., 100-200 words, ..., more than 1000 words` or something like that.

model/model_training/custom_datasets/instruction.py
```diff
@@ -21,25 +22,65 @@
...
 
     def __len__(self):
         return len(self.dataset)
 
     def __getitem__(self, idx) -> DatasetEntry:
         data = self.dataset[idx]
         lang = None
-        # these datasets have been found to have above 95% english sentences.
-        if self.name in ["grade_school_math_instructions"]:
+        # use "en" for datasets which have more than 95% English messages
```
dvruette, line 71: this will teach the model to sometimes (with a < 5% chance) generate a language other than english even though the language is specified as english. probably not an issue, but it would be best to filter the affected datasets or augment them with language attributes.

model/model_training/custom_datasets/qa_datasets.py
```diff
@@ -211,7 +212,14 @@ def __init__(self, mode: str = "sft", max_answers: int = 5) -> None:
             # Then take only the first `max_answers` elements (usually there are just
             # 2, but there are examples where we have more)
             answers_sorted = [x[0] for x in sorted(answers.items(), key=lambda x: -1 * x[1])]
-            self.rows.append(DatasetEntry(questions=[question], answers=[answers_sorted[:max_answers]], lang="en"))
+            self.rows.append(
+                create_dataset_entry_qa(
+                    mode=mode,
+                    questions=[question],
+                    answers=[answers_sorted[:max_answers]],
+                    lang="en",
```
dvruette, line 220: similar (non-)issue here as outlined above: samples are not always guaranteed to be english

model/model_training/custom_datasets/utils.py
```diff
@@ -1,4 +1,46 @@
-FILTER_BY_WORDS = ["openai"]
+# mostly taken from
+# https://huggingface.co/datasets/gozfarb/ShareGPT_Vicuna_unfiltered/blob/main/optional_clean.py,
+# https://huggingface.co/datasets/ehartford/WizardLM_alpaca_evol_instruct_70k_unfiltered/blob/main/wizardlm_clean.py
+FILTER_BY_WORDS = [
```
dvruette, line 4: This looks like duplicate code, can probably be shared with the `check_dataset_appearance` script

dvruette (approved):
Awesome progress! Some minor comments that don't block merging. lgtm.

side note: i saw there is a dataloader for continued pre-training, but it seems like this is not yet used during training, correct?

model/model_training/custom_datasets/qa_datasets.py
andreaskoepf: These lines are part of the WebGPT loader class. Does WebGPT contain non-english entries?

model/model_training/custom_datasets/instruction.py
andreaskoepf: I have roughly looked through all the datasets that have been listed here and I had not found any non-English entries. Maybe there are some, but I don't think that those other languages play an important role. I think Tobias used an automatic language detection tool to measure the percentage of English items in the datasets (and he initially added only `grade_shool_math_instructions`). I added the other ones. If you think that one of them is not (very) predominantly English let me know.

model/model_training/custom_datasets/formatting.py
andreaskoepf: Yes, length as it is is very experimental. Now that we see it works in principle we can refine it (if we decide to use system-tag profiles in the future). Something similar applies to the numeric values of the properties, which are currently encoded as numbers with one decimal digit, e.g. `quality: 0.4` .. we could reduce the number of tokens by using only a single number, e.g. 0-9.

dvruette: yes, i had the same thought but concluded that single decimal numbers are probably common enough to only require a single token. i do like the semantics of decimal numbers, which probably already are connotated with “strength” or “sliding value” in embedding space.

model/model_training/custom_datasets/instruction.py
dvruette: i see, then it’s probably a high-effort low-payoff issue. just something to be aware of and to watch out for. if we notice that the model sometimes randomly generates different languages, this could be one possible cause

model/model_training/custom_datasets/qa_datasets.py
dvruette: i haven’t looked at the dataset, but i would assume that there will be some non-english entries. same as above: probably not worth worrying about for now.

model/model_training/custom_datasets/instruction.py
CloseChoice: Yes, I just checked with [langdetect](https://github.com/Mimino666/langdetect) and just found that one. +1 for @dvruette comment, we should totally be aware of that.

model/model_training/custom_datasets/qa_datasets.py
CloseChoice: I just reran the check on the dataset and these are the results:
```
defaultdict(int,
            {'en': 28272,
             'af': 73,
             'fr': 34,
             'tl': 9,
             'nl': 30,
             'da': 3,
             'so': 4,
             'fi': 5,
             'ca': 9,
             'it': 7,
             'de': 11,
             'pt': 7,
             'ro': 10,
             'no': 4,
             'cy': 10,
             'es': 2,
             'id': 6,
             'sv': 2,
             'pl': 3})
```
So `99.19651942037121 %` of the datasets are english, according to `langdetect`. I would just keep this language tag here

model/model_training/check_dataset_appearances.py
CloseChoice: I'll remove `OpenAI` altogether. Found some example like `Best regards, your OpenAI language model` and other things. It's probably safer to remove it altogether

model/model_training/custom_datasets/utils.py
CloseChoice: used the code from utils.py

model/model_training/check_dataset_appearances.py
CloseChoice: done

PR #2998, opened 2023-05-01
Auto focus chat input

notmd:
I decided to only focus on desktop since it is bad practice to auto-focus on mobile, the keyboard on mobile take a lot space, but let's see how the user feedback with it

### DIFF
--- a/website/src/hooks/ui/useFallbackRef.ts
+++ b/website/src/hooks/ui/useFallbackRef.ts
@@ -1,0 +1,7 @@
+import { ForwardedRef, RefObject, useRef } from "react";
+
+export const useFallbackRef = <T>(maybeRef?: ForwardedRef<T>) => {
+  const ref = useRef(null);
+
+  return (maybeRef || ref) as RefObject<T>;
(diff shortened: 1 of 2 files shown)


website/src/hooks/ui/useFallbackRef.ts
```diff
@@ -0,0 +1,7 @@
+import { ForwardedRef, RefObject, useRef } from "react";
+
+export const useFallbackRef = <T>(maybeRef?: ForwardedRef<T>) => {
+  const ref = useRef(null);
+
+  return (maybeRef || ref) as RefObject<T>;
+};
```
AbdBarho, line 7: I am not sure I understand what this does?

AbdBarho (approved):

website/src/hooks/ui/useFallbackRef.ts
notmd: @AbdBarho you can read more about it here https://mtsknn.fi/blog/react-fallback-ref

PR #3004, opened 2023-05-01
Only set default chat title when empty

### LINKED ISSUE #2787: Chat title gets overwritten even if the user has set title
It's annoying that the chat title gets overwritten even if I have set it already. We can implement some sort of checking if the title is default title and then only create the title.

notmd:
fix #2787


olliestanley (approved):
Thanks

PR #3006, opened 2023-05-01
Update deps

notmd:
I also unpin some deps


website/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "next dev",
     "build": "next build",
     "build:analyze": "cross-env ANALYZE=true next build",
-    "start": "next start",
+    "start": "node .next/standalone/server.js",
```
notmd, line 10: Nextjs warn when run next start with standalone mode. I though this will fix the CI but no luck

andreaskoepf (approved):
Thanks!

PR #3008, opened 2023-05-02
fix sponsored by text appearing after sponsors

z11h:
fixes the "sponsored by" text appearing after the actual sponsors.

before:
image

after:
image

let me know if any other changes are recommended =)


AbdBarho (approved):
Thanks 

We should extract this to the locale

Issue #2695, opened 2023-04-18
Improve code & style of "try our assistant" in dashboard


yk:
https://github.com/LAION-AI/Open-Assistant/blob/82540c044ee57171fb66e83986b933649d25e7fb/website/src/pages/dashboard.tsx#L39-L49

Cecileblbl:
Happy to help with this

dosacat:
Hi, do assign it to me if still available

yk:
thank you both, I've assigned you both, maybe you coordinate amongst each other

dosacat:
@Cecileblbl would you like to take this forward on the discord?

Cecileblbl:
Yes you can find me under cecileblblbl in the Open Assistant Discord

dosacat:
> I'm bitternectar on the discord #6458

andreaskoepf:
Was this merged? Please link PR.

PR #3015, opened 2023-05-02
Improve dashboard button 

Cecileblbl:
Suggesting this design for #2695

It makes the button shaped like a dialogue bubble similar to the OA logo.

### DIFF
--- a/website/src/pages/dashboard.tsx
+++ b/website/src/pages/dashboard.tsx
@@ -65,11 +65,9 @@ const Dashboard = () => {
 
         {getEnv().ENABLE_CHAT && (
           <Flex direction="column" gap={4}>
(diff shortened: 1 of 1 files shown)


website/src/pages/dashboard.tsx
```diff
@@ -27,6 +27,33 @@ const Dashboard = () => {
     return { [TaskCategory.Random]: taskTypes };
   }, [data]);
 
+  const chatButtonStyle = {
```
AbdBarho, line 30: this content can be moved out of the function

AbdBarho (commented):
Can you include a screenshot?

AbdBarho (approved):

PR #3016, opened 2023-05-02
Fix model cli

wrmthorne:
Fixed package naming errors, hard coding of tokenizer and error `RuntimeError: "LayerNormKernelImpl" not implemented for 'Half'` (cause by using CPU instead of GPU)


andreaskoepf (approved):
lgtm

PR #3017, opened 2023-05-02
Fix plugins

olliestanley:
Plugins was compatible only with `text-generation-inference` based workers and therefore worked on Dragan's machines but did not work on OA prod. This resolves the incompatibility.


olliestanley:
Marking this as draft as Dragan has identified another problem which will need to be remedied for plugins to work. No point merging this until that is fixed

inference/worker/chat_chain_utils.py
```diff
@@ -11,7 +11,7 @@
 from langchain.prompts import PromptTemplate
 from loguru import logger
 from oasst_shared.schemas import inference
-from opeanapi_parser import prepare_plugin_for_llm
+from openapi_parser import prepare_plugin_for_llm
```
pevogam, line 14: Interesting! Why wasn't this detected during use and importing of the module? Does the current design fail quietly?

pevogam (commented):
@olliestanley Can you elaborate a bit more on what is being fixed or extended here regarding the original pull request? I noticed the merged original pull request no longer introduces the sample calculator plugin, what are the plans regarding this?

inference/worker/chat_chain_utils.py
olliestanley: No, this wasn't broken, the module was also actually named `opeanapi_parser`, I just renamed it

olliestanley:
> @olliestanley Can you elaborate a bit more on what is being fixed or extended here regarding the original pull request? I noticed the merged original pull request no longer introduces the sample calculator plugin, what are the plans regarding this?

There were three problems in prod which did not show up in local testing, because prod uses `basic_hf_server` worker whereas testing was using `text-generation-inference` worker:

- the chain code uses `generate` endpoint which only exists in `text-generation-inference` rather than `generate_stream`
- basic HF server does not support custom stopping sequences
- basic HF server does not support fixed random seeds

This PR resolves all three of those

flozi00:
image
 Am i missing something to get it working ?

Issue #2582, opened 2023-04-16
Implement chat deletion


yk:
for now we just have hide

olliestanley:
The inference backend supports full deletion so we just need a frontend addition here.

ruoniw:
Will try take a look at this one as well

yk:
@ruoniw thank you! can this e closed?

olliestanley:
Yes this can be closed

titonova:
@olliestanley Sorry, how do you delete messages now, from the frontend

PR #3019, opened 2023-05-03
add delete function

ruoniw:
https://github.com/LAION-AI/Open-Assistant/issues/2582 

Add a function for delete, note the inference server already have the delete endpoint so this is only a frontend change. 

https://user-images.githubusercontent.com/119783413/235826890-b95a43f5-9511-4692-8f60-610ce43fdfe8.mov


ruoniw:
Right now if the deleted chat is the current chat, there will be a failed to open chat problem (currentChat can't be loaded). Ideally seems we want to redirect to the next chat or if no more chat, navigate back to the dashboard. But it's not implemented in this PR yet.

https://user-images.githubusercontent.com/119783413/235828697-1259a0f1-4826-448a-81bd-b457895ac43c.mov

olliestanley:
Thank you. I think it would be good to have this as part of a dropdown with extra buttons (so that the user by default sees only Hide and Edit Title, but can click the `...` and see Delete and in future Data Opt Out). Would that be possible?

AbdBarho:
Maybe additionally we want to let the users confirm the delete, not because they might delete by accident but because the data will be removed permanently and we cannot use it for further training.

ruoniw:
Added a dropdown and a confirmation dialog. 

https://user-images.githubusercontent.com/119783413/236356690-5bf8cb21-34dc-4a81-b15c-35f7f4a43ee4.mov

AbdBarho:
@ruoniw Thank you for preparing this, I had to do some final cross-service related changes, and minimal improvements to the mobile experience.

Blocked by https://github.com/LAION-AI/Open-Assistant/pull/3046

@olliestanley we throw a 404 if chat is not found, instead of error.
@notmd had to do some hacks to get this working on mobile, the three dot menu would be shown under the mobile menu.

please review!

olliestanley (approved):
Backend change LGTM

olliestanley:
@AbdBarho I think we should get this and #3046 (will need your review) merged and onto dev ASAP, then we can test on dev and hopefully get onto prod today

notmd:
> 

Maybe we can remove the 3 dots for now, and show the delete icon directly. I plan to work on a new design later.

olliestanley:
> Maybe we can remove the 3 dots for now, and show the delete icon directly. I plan to work on a new design later.

It would be unfortunate to have this on prod with delete icon showing directly tbh. Potentially a lot of unnecessary data loss if users click delete when they really just want to hide

ruoniw:
@AbdBarho thanks for wrapping this up! Looks good to me.

notmd:
> @notmd had to do some hacks to get this working on mobile, the three dot menu would be shown under the mobile menu.

Have you try a higher zIndex?

website/src/components/Chat/ChatListBase.tsx
```diff
@@ -98,7 +98,13 @@ export const ChatListBase = memo(function ChatListBase({
         }}
       >
         {chats.map((chat) => (
-          <ChatListItem key={chat.id} chat={chat} onUpdateTitle={handleUpdateTitle} onHide={handleHide} />
+          <ChatListItem
+            key={chat.id}
+            chat={chat}
+            onUpdateTitle={handleUpdateTitle}
+            onHide={removeItemFromList}
+            onDelete={removeItemFromList}
```
notmd, line 106: We still need to redirect user to another page after deleting, maybe latest chat or /chat page

AbdBarho:
> Have you try a higher zIndex?

Yeah yeah it works now, I just wanted to inform you that there were some hacks involved.

website/src/components/Chat/ChatListBase.tsx
AbdBarho: Already implemented, maybe not in the best way, because you can delete a chat even though you haven't opened it, so we don't redirect always, we only redirect when we get a 404 on the chat.

AbdBarho (approved):

PR #3023, opened 2023-05-03
Add back chat queue info

notmd:
Look not good, I have no idea about a better UI
image


andreaskoepf (approved):
nice, thanks!

AbdBarho (approved):
I think it looks fine

PR #3028, opened 2023-05-03
Add Bavarian ("Bairisch") Language

79Luca79:
Hello OA team.
**This is the first time I'm doing a pull request or anything with code, so I'm sorry for any mistakes.**

I have translated the words into the Bairisch ("bar" would be the code) language (Bairisch is not a dialect of "German" [with German, {South Tyroleans, Austrians, Swiss} only the "German language" ("New High German written language") is meant anyway, but in everyday life nobody says the word anyway!) . 

The Bavarian dialect I used is "South Tyrolean". Yes its not a "real" Dialect, but i think it suffices for now.
But I called this language Bairisch and not Tirolisch because it has 18 million people and not only 1.3 million. (But the number changes depending on how you define Tyrolean, it can also be 500 000). Also because I want as many people as possible to use the new language selection and finally be able to write Bavarian and not "neuhochdeutsche schriftsprache".

I have heard that there are apparently plans to include smaller dialects (Tyrolean, Carinthian, North Bavarian, I mean) in the list and to subdivide them. I think that's a great idea!

We can then also divide "Bairisch" into groups in the future as was discussed a bit here i saw.

Thank you for any feedback! :)


Logophoman:
Awesome work! I'd really like to see all the Dialects being integrated too somehow 🥳

website/public/locales/bar/dashboard.json
79Luca79, line 1: This Section is not yet translated but i dont know why? I did it and now its wrong again??

79Luca79:
Yesterday i finished all the Translation Work. For that i used the normal Github Dev Code Interface. Upon finishing a Section i clicked "Save".
Now i look over some Sections here in the Commits of this Pull Request and see that many Sections are still in Swabian (i used this as a Base and just copied the Code and translated it to Tyrolean.)

Is there a way to again find these Files i saved or do i need to start again? If i need to start again i will use the English Translation and translate from there. :(

79Luca79:
Sorry for the Mess here...
Im happy to say that i now found all of my Translations again in the Github Dev Editor. I will now try to fix my Things.

andreaskoepf:
The likelihood for merging German dialects like Swabian or Bavarian is IMO very low unless someone really adds special support for language dialects (e.g. to ensure they don't show up during primary language selection). The only dialect we currently support is [Brazilian Portuguese](https://en.wikipedia.org/wiki/Brazilian_Portuguese) (>200M speakers).

Given the fact that we haven't yet collected a meaningful number of messages for "Hochdeutsch" I would in general question this further fragmentation.

stefangrotz:
@andreaskoepf  I kind of agree that we should focus on Standard German right now. In the long run I would love to see support for minority languages of all kinds.

For German dialects I would like to propose a compromise: Let's allow questions in a dialects in the German Dataset that are always answered in Standard German.

andreaskoepf:
closing this in favor of #3033

PR #3029, opened 2023-05-03
Automated task fulfillment

layterz:
This is still a WIP - I still need to test locally and write some unit tests, but it should be enough code to demonstrate the approach I'm taking. I have a couple of questions so wanted to check in now to get some feedback before continuing.

Changes:
- Added a `TaskResponder` with a `complete(task)` method. Only created a `ChatGPTTaskResponder` for now.
- Add two new methods to `TreeManager`:
    - `select_task_responder` this selects a responder (e.g. human, chatgpt, open-assistant, etc) from a list configured in settings.
    - `complete_task` which accepts a task and a model and attempts to complete the task with the responder defined by the model.
- In `request_task` I've added a while loop which calls `tm.select_task_responder(request)`. A task is created for the selected responder and the loop continues until a task for the requesting user has been created. I'm not sure if this is the intended logic.
- Added a unique nullable `ai_model` field to `User`. 
- Created a new `complete-pending-ai-tasks` celery worker which iterates over all users with an `ai_model` and finds and pending tasks for that user. Then it iterates over those tasks and attempts to complete them using the `TreeManager` method described above.

This approach means there'll need to be a DB user per automated responder, which complicates setup a bit. But the benefits are that it works with most of the existing code with few changes and any logic to rate limit, log, etc per user will be reused.

@yk can you take a look and see if this fits the requirements you had in mind?


olliestanley:
Thanks for this PR. Haven't reviewed but will just say I am pretty sure we don't want to include a ChatGPT task responder, the idea here would be to give these tasks to OA models running on the inference server :D

layterz:
Ha, yeah I kind of guessed. That's just placeholder to demonstrate how the responders would work. Will try to update with an OA example

andreaskoepf:
(converted to draft due to wip status)

olliestanley:
Hi @layterz are you still working on this?

layterz:
Hi @olliestanley,

Sorry - I've been travelling so dropped this unfortunately. I think the code is mostly there, but was having trouble testing it due to docker crashes. I can try again now I'm back, but won't have time until next week realistically. If that's blocking happy for you to reassign.

Rich

Issue #2859, opened 2023-04-23
In the chat interface, switching to "custom" should default-set the parameters of the previously selected preset


yk:

notmd:
Here is the code for anyone wants to work on this.
https://github.com/LAION-AI/Open-Assistant/blob/0ba118cc9e1c4d8fd24e350139ab16cfc8d01f71/website/src/components/Chat/ChatConfigForm.tsx#L95-L106

simonusher:
Hi! I'd like to work on this.

PR #3032, opened 2023-05-04
use most recent preset sampling parameters for custom preset

z11h:
This PR fixes: #2859. We now store the most recent preset used and use that as default for when you select custom preset.

Honestly, we could just get rid of the "customPresetDefaultValue" but not necessary.

My first real PR here, hopefully the first of many! Let me know if any changes are needed :)


notmd:
Hi, thanks for your contribution. But I'm already working on this, the logic will be entirely different for support more up coming features

andreaskoepf:
@z11h thanks four your effort .. we definitely would appreciate your help for other frontend tasks.

To prevent loss of time/work in the future it would be best to first discuss in an issue or with the main contributors in the OA-discord. If you are not already in the contributor role, please ping me via DM on discord.

PR #3034, opened 2023-05-04
Removing WikiHow dataset at WikiHow request

0x22almostEvil:
Sadly, WikiHow Team contacted me and said that their data cannot be used in this way...

### DIFF
--- a/data/datasets/__init__.py
+++ b/data/datasets/__init__.py
@@ -24,7 +24,6 @@ INSTRUCTION_DATASETS = {
     "oa_dolly_15k": "OllieStanley/oa_dolly_15k",
     "poetry_instruction": "checkai/instruction-poems",
     "oa_stackexchange": "donfu/oa-stackexchange",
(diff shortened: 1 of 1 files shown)


0x22almostEvil:
I have contacted LAION about the legality of their ban and potential solutions, but we probably won't be able to use my dataset in its **current** form.

andreaskoepf (approved):

z11h:
What legal basis/ability does WikiHow even have to take against an open-source project like this? Is there any precedent for lawsuits for scraping websites and training a model on them?

Training algorithms on copyrighted data is not illegal, according to the United States 2nd Circuit Court btw.

olliestanley:
> What legal basis/ability does WikiHow even have to take against an open-source project like this? Is there any precedent for lawsuits for scraping websites and training a model on them?
> 
> Training algorithms on copyrighted data is not illegal, according to the United States 2nd Circuit Court btw.

I think you're right that there's no legal problem here, and we should not talk as if there is. However I think as a project OA is inclined to respect the wishes of website operators if they ask us not to use their data.

olliestanley (approved):

PR #3044, opened 2023-05-05
Message drafts

### LINKED ISSUE #2785: Remember user prompt when switching between chats
Currently, The prompt gets lost when we type something in the chat form and switch to another chat and come back to the chat we were editing.

[remember-prompt.webm](https://user-images.githubusercontent.com/94373953/233428793-31fadbc0-a712-499d-a30a-6bfb2eeaa57f.webm)

It would be a very good UX if we remember the prompt for individual chat.

### LINKED ISSUE #2931: Message ranking In Conversations
This is a proposal for multiple messages to be generated per prompt to collect additional RLHF data from user conversations. It would work by generating the first 'x' tokens of a message and then prompting the user to select one which would be generated to completion. The value of 'x' would change based on the load on the inference servers at the moment ranging from whole messages to completely disabling the feature under high load. This would give lots of ranking data for RLHF while not butchering the performance of the site, if the server is at low usage then you get lots of RLHF data while improving quality of life for users, if the server is at capacity then it is simply disabled. Giving the option to see 'x' number of tokens from a few possible answers would also reduce the number of regenerations reducing the net increase in load.

---

It would go something like this:

User: What is a croissant?

1. A croissant is a type of French pastry made with layers of yeast-raised dough that are rolled and folded several times. The resulting dough is shaped into an crescent shape and baked until flaky, crisp, and...

2. Croissants are buttery and delicious. They originated in Austria during the 13th century, with the first recorded use of the word “croissant” appearing in a French manuscript from the year...

3. A crescent shaped bread rolls often eaten at breakfast. (Sometimes short answers can fit within 'x' tokens)

(User selects reply 1)

Assistant: "A croissant is a type of French pastry made with layers of yeast-raised dough that are rolled and folded several times. The resulting dough is shaped into an crescent shape and baked until flaky, crisp, and golden brown in color.

Traditionally, croissants were made with butter, which gave them their signature rich flavor and flakiness. However, modern versions may also use margarine or other spreads to reduce the fat content. Croissants can be served plain as a breakfast bread or filled with various ingredients such as jam, cheese, or chocolate for added sweetness or savory flavors.

Croissants have become popular around the world, particularly in countries with strong ties to France such as Belgium, Switzerland, and Canada, as well as many other parts of Europe and North America where French cuisine has gained a following. They are often served at cafes alongside coffee or tea, making it a staple breakfast item."

someone13574:
closes #2931 (slightly changed goal based on advice from the discord, generate full messages, not 'x' tokens. Full messages are more useful data)

- [x] Create draft selection UI
- [x] Draft inference
- [x] Option to regenerate drafts and serve 3 new ones
- [x] Remember last viewed sibling message
- [x] Store selected draft training data for RLHF
- ~~[ ] Disable drafts when queue is too long / server is under load~~ (Suggested to leave to next PR in the discord)
- [x] Draft markdown rendering
- [x] 'Used plugin' UI for drafts
- [x] Resolve merge conflicts


inference/server/alembic/versions/2023_05_07_0923-19178d144265_add_sibling_active_to_message.py
```diff
@@ -0,0 +1,28 @@
+"""add_sibling_active_to_message
```
olliestanley, line 1: Can we merge both Alembic scripts into one before merging this? Will help reduce clutter

inference/server/oasst_inference_server/models/chat.py
```diff
@@ -19,6 +19,7 @@ class DbMessage(SQLModel, table=True):
     reports: list["DbReport"] = Relationship(back_populates="message")
 
     parent_id: str | None = Field(None)
+    active_sibling: bool | None = Field(None, sa_column=sa.Column(pg.JSONB))
```
olliestanley, line 22: What does `active_sibling` mean exactly? Feels like an odd name

inference/server/oasst_inference_server/models/chat.py
```diff
@@ -39,6 +40,7 @@ class DbMessage(SQLModel, table=True):
     worker_config: inference.WorkerConfig | None = Field(None, sa_column=sa.Column(pg.JSONB))
 
     score: int = Field(0)
+    inferior_drafts: list[str] | None = Field(None, sa_column=sa.Column(pg.JSONB))
```
olliestanley, line 43: I guess that the way this works is that for each prompter message with multiple drafted replies, each reply has a list of message IDs which are "inferior" to it?

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -116,11 +121,96 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
     },
     [getConfigValues, setIsSending, setShowEncourageMessage, toast]
   );
+  const createAssistantDrafts = useCallback(
```
olliestanley, line 124: I wonder if this will be a problem when plugins are enabled... With sufficient retries/depth plugins take up a lot of model time/compute already, several drafts might be a struggle. Maybe we could reduce plugin max depth for drafts?

olliestanley (changes requested):

olliestanley:
Thanks, I think this mostly makes sense, I left a couple of questions and one change request

inference/server/oasst_inference_server/models/chat.py
someone13574: In a conversation tree you can have messages which you can switch between using the arrows at the bottom. This variable just stores whether or not that particular message was the one which was last showed. The 'sibling' part is referring to how these messages come from the same parent message so I thought it would be a good term to use. If you have a better alternative I can change it.

inference/server/oasst_inference_server/models/chat.py
someone13574: Yes, when the user selects a draft message the two other messages have their ID's added to the selected message's inferior message list.

website/src/components/Chat/ChatConversation.tsx
someone13574: I might just disable drafts when plugins are enabled until the disabling under load feature is added as to not overload the inference server. Reducing max iterations is also a possibility.

olliestanley: > I might just disable drafts when plugins are enabled until the disabling under load feature is added as to not overload the inference server. Reducing max iterations is also a possibility.

Yes, not only overload but also just taking a lot of time for the user to get the drafts... Maybe include a setting to toggle enabling this for plugins, and have it disabled by default

someone13574: I've added an ENABLE_DRAFTS_FOR_PLUGINS environment variable which is defaulted to false

inference/server/oasst_inference_server/models/chat.py
olliestanley: Hmm ok. This feels like it is really information about the chat, not the message. Can we not just store a single `active_message` value in the chat table, which is the final message of the currently active path down the chat tree? Then from this it is already implied what siblings are active earlier in the tree

This would also simplify the endpoint, we could merge it into `update_chat` endpoint

inference/server/oasst_inference_server/models/chat.py
olliestanley: Ok. Thinking from the perspective of ease of querying/interpreting this data. We could probably have a `draft_evaluation` table with fields like:
```
id: str
user_id: str
selected_message_id: str
inferior_message_ids: list[str]
```

inference/server/oasst_inference_server/user_chat_repository.py
```diff
@@ -257,6 +252,27 @@ async def update_score(self, message_id: str, score: int) -> models.DbMessage:
...
+        draft_eval = models.DbDraftEval(
+            chat_id=message.chat_id,
+            user_id=message.chat.user_id,
+            selected_message_id=message.id,
+            inferior_message_ids=inferior_message_ids,
+        )
+        self.session.add(draft_eval)
+        await self.session.commit()
+        await self.session.refresh(draft_eval)
+        return draft_eval
```
olliestanley, line 274: We can probably remove the refresh and return here

olliestanley (commented):
Thanks! This seems fine to me from backend perspective now. Will approve tomorrow if neither of Yannic/Andreas wish to review first

inference/server/oasst_inference_server/models/chat.py
someone13574: Would it make sense to rename the `draft_evaluation` table to a more general `message_evaluation` so that it could be reused in the future for something like ChatGPT's "Better, Worse, Same" on message regeneration?

olliestanley: Yeah this could make sense - will leave up to you

olliestanley (approved):
Backend changes LGTM. Have not reviewed website changes.

website/src/components/Chat/ChatAssistantDraftViewer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      setRegenIndex(draftMessages.length);
+    }
+  }, [onRetry, regenIndex, draftMessages, setExpandedMessage, setRegenIndex]);
+
+  const containerBackgroundColor = useColorModeValue("#DFE8F1", "#42536B");
+  const responseHoverBackgroundColor = useColorModeValue("gray.300", "blackAlpha.300");
+
+  return (
+    <Flex
+      key="chat-message-select"
```
notmd, line 94: Is there any reason you use a key here?

website/src/components/Chat/ChatAssistantDraftViewer.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+type ChatAssistantDraftViewerProps = {
+  chatId: string;
+  streamedDrafts: string[];
+  draftMessages: InferenceMessage[][];
+  onDraftPicked: OnDraftPickedFn;
+  onRetry: OnRetryFn;
+};
+
+export const ChatAssistantDraftViewer = ({
```
notmd, line 34: This component is way too big, we should extract it to 2 component. One for render the content, and one for display the paging logic.

website/src/components/Chat/ChatAssistantDraftViewer.tsx
```diff
@@ -0,0 +1,234 @@
...
+        flexDirection="column"
+      >
+        {isLoading ? (
+          <CircularProgress isIndeterminate />
+        ) : (
+          <>
+            {(isComplete && regenIndex < draftMessages.length && draftMessages[regenIndex]?.length !== 0
+              ? draftMessages[regenIndex]
+              : streamedDrafts
+            ).map((message, index) => (
```
notmd, line 129: TS is warning that the message is any, and will be error if we turn on strict mode. Extract to sub-component will make the code safer here.

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -108,25 +117,113 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
...
+          const { body, status } = await fetch(API_ROUTES.STREAM_CHAT_MESSAGE(chatId, draft_message.id));
+
+          let inference_message: InferenceMessage | null;
+          if (status === 204) {
+            inference_message = await get(API_ROUTES.GET_MESSAGE(chatId, draft_message.id));
+          } else {
+            inference_message = await handleChatEventStream({
+              stream: body!,
+              onError: console.error,
+              onPending: setQueueInfo,
```
notmd, line 171: This line causes a lot of race condition, and make the queue info flashing all the times. Can we call setQueueInfo only 1 time or move the queue info to inside the message?

website/src/components/Chat/ChatConversationTree.tsx
```diff
@@ -38,30 +48,53 @@ export const ChatConversationTree = memo(function ChatConversationTree({
...
-  const [index, setIndex] = useState(trees.length - 1);
-  useIsomorphicLayoutEffect(() => {
-    setIndex(trees.length - 1); // always show last child
-  }, [trees.length]);
+}: {
+  trees: Tree<InferenceMessage>[];
+  retryingParentId: string | null;
+  activeMessageId: string | null;
+} & StrictOmit<ChatMessageEntryProps, "message" | "canRetry">) => {
+  const { trigger: triggerActiveMessageId } = useSWRMutation(API_ROUTES.UPDATE_CHAT(), put);
```
notmd, line 58: Why do we need this?

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -251,14 +378,70 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
...
+      isSending,
+      setIsSending,
+      setDraftMessages,
+      getConfigValues,
+      ENABLE_DRAFTS_FOR_PLUGINS,
+    ]
   );
 
+  const handleDraftPicked = useCallback(
+    async (chat_id, regen_index, index) => {
```
notmd, line 404: Can you add typing to this function?

website/src/components/Chat/ChatConversationTree.tsx
```diff
@@ -38,30 +48,53 @@ export const ChatConversationTree = memo(function ChatConversationTree({
...
 
   const sortedTrees = useMemo(() => trees.sort((a, b) => Date.parse(a.created_at) - Date.parse(b.created_at)), [trees]);
-  const actualIndex = Math.min(index, trees.length - 1); // index sometimes out of bounds because useIsomorphicLayoutEffect only reset the index on the next render
+  const [index, setIndex] = useState(() => 0);
```
notmd, line 61:
```suggestion
  const [index, setIndex] = useState(0);
```

notmd (commented):
Thank you. Look good overall I might need to take another round, there are some eslint warnings about the hook deps, could you fix that? Also, there are many duplicated codes, I'm fine with that, we could refactor later.

website/public/locales/en/chat.json
```diff
@@ -36,6 +37,8 @@
   "queue_info": "Your message is queued, you are at position {{ queuePosition, number, integer }} in the queue.",
   "remove_plugin": "Remove Plugin",
   "repetition_penalty": "Repetition penalty",
+  "show_less": "Show Less",
+  "show_more": "Show More",
```
notmd, line 41: Can you move this to the common file?

notmd:
Also, can we make an option to allow users to choose whether they want to generate drafts or not?

website/src/components/Chat/ChatAssistantDraftViewer.tsx
```diff
@@ -0,0 +1,156 @@
...
+              used_plugin: object | null;
+            }[])
+          : (streamedDrafts.map((message) => ({ content: message, used_plugin: null })) as {
+              content: string;
+              used_plugin: object | null;
+            }[])
+        ).map(({ content, used_plugin }, index) => (
+          <>
+            <Box
+              key={`message-$index`}
```
notmd, line 78:
```suggestion
              key={`draft-message-${index}`}
```

notmd (approved):
Thank you, need 1 more approval from Yannic or Andreas.

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -108,25 +117,115 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
...
 
   const sendPrompterMessage = useCallback(async () => {
     const content = inputRef.current?.value.trim();
     if (!content || isSending) {
       return;
     }
+
+    if (isAwaitingMessageSelect) {
+      return toast({
+        title: "Please select a draft to continue.",
```
andreaskoepf, line 225: This string should be added to `public/locales/en/chat.json` for proper app localization.

website/src/types/Config.ts
```diff
@@ -1,5 +1,7 @@
 export interface BrowserConfig {
   ENABLE_CHAT: boolean;
+  ENABLE_DRAFTS_FOR_PLUGINS: boolean;
```
andreaskoepf, line 3: Since the meaning of this configuration option is not immediately clear (at least it was not fro me quickly reading through the PR) it would be good to have a comment here which explains what it does.

inference/server/oasst_inference_server/models/chat.py
andreaskoepf: Sorry only seeing this now: How are the multiple generated draft messages now referenced in the database? And was is finally stored in the chat-message tree?

inference/server/alembic/versions/2023_05_18_1748-2d67fbdc5b46_add_active_messsage_id_and_message_eval.py
```diff
@@ -0,0 +1,55 @@
...
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["user.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(op.f("ix_message_evaluation_chat_id"), "message_evaluation", ["chat_id"], unique=False)
+    op.create_index(op.f("ix_message_evaluation_user_id"), "message_evaluation", ["user_id"], unique=False)
+    op.add_column("chat", sa.Column("active_message_id", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```
andreaskoepf, line 45: Why is this a JSONB column?

andreaskoepf (changes requested):
Thanks a lot for this important feature.
I would appreciate feedback to the following points:

1. It seems as if the draft generation would always be active when `NUM_GENERATED_DRAFTS > 1`. This will completely break any regular chat-experience and force the user to go through the message-selection process, even for simple "Hello" prompts. In general we definitely want to collect this form of feedback data but the question is if we really should make regular use of assistant impossible. If message alternatives would be generated when user presses regenerate or down-vote it would IMO be a much more unobtrusive experience and users would get additional option when they actually need them.

2. It is not clear to me how draft messages are tracked now in the database. Could you please explain the meaning the `active_message_id` column in `chat` table (and why is it a json field)?  How are drafts messages stored and referenced in the database?

someone13574:
> Thanks a lot for this important feature. I would appreciate feedback to the following points:
> 
>     1. It seems as if the draft generation would always be active when `NUM_GENERATED_DRAFTS > 1`. This will completely break any regular chat-experience and force the user to go through the message-selection process, even for simple "Hello" prompts. In general we definitely want to collect this form of feedback data but the question is if we really should make regular use of assistant impossible. If message alternatives would be generated when user presses regenerate or down-vote it would IMO be a much more unobtrusive experience and users would get additional option when they actually need them.
> 
>     2. It is not clear to me how draft messages are tracked now in the database. Could you please explain the meaning the `active_message_id` column in `chat` table (and why is it a json field)?  How are drafts messages stored and referenced in the database?

1. I'll make the drafts less intrusive by only generating drafts if either the prompt is long (thus indicating a more complex request) or if the user regenerates a message (indicating that additional options would be beneficial). This should make the process less intrusive and act as a filter against data we really don't need comparisons for (trivial tasks or general conversation).
2. Draft messages are stored the same way as normal message and the comparisons are stored in an new table called `message_eval` where it has the id of the selected message and the id's of the messages it was selected over. The `active_message_id` stores the id of the last message in the currently visible thread so that the conversation can be loaded in the same state as it was when a user closes it. This is because all the unselected drafts are also added to the conversation as other threads and the last message for a given parent message wont always be the one the user was actually using, so loading onto that unused message instead of where the user actually was writing was very bad for ux, so I made it remember which thread was last selected. I'd be open to a better name. Also using a json was a mistake, I'll fix it.

inference/server/oasst_inference_server/models/chat.py
someone13574: The draft messages are stored as regular messages in the messages table. The message comparisons are stored in a new table called `message_eval` which contains the chat id, user id, message id of the selected message, and the message ids it was selected over.

andreaskoepf (approved):

Issue #3035, opened 2023-05-04
Allow to save custom preset


notmd:
I remember I saw this issue before but I can't find it

PR #3050, opened 2023-05-05
Chat config and plugin UI/UX rework

### LINKED ISSUE #2859: In the chat interface, switching to "custom" should default-set the parameters of the previously selected preset

notmd:
- Rework chat cache format, now store the preset name instead the preset config.
- Rework chat config hydration logic, in order to sync cache correctly I only render 1 `ChatConfigForm` only to avoid conditional race. Also the `ChatConfigForm` now have local states, we can't sync the local state if we render 2 form. I think we should apply this to the `ChatListBase` as well
- Clean up the `PluginChoose` UI/UX
- Now only send `enabled` plugins to the inference only, so the inference can remove the enabled check.
- close #2859


website/src/utils/chat.ts
```diff
@@ -1,19 +1,29 @@
...
 
-const CHAT_CONFIG_KEY = "CHAT_CONFIG";
+const CHAT_CONFIG_KEY = "CHAT_CONFIG_V2";
 
-export const setConfigCache = (config: ChatConfigFormData) => {
+export type CachedChatConfig = {
+  model_config_name: string;
+  selectedPresetName: string;
+  custom_preset_config: SamplingParameters;
+  custom_presets: Array<{ name: string; config: SamplingParameters }>;
```
notmd, line 10: This will be used in the next PR when I implement https://github.com/LAION-AI/Open-Assistant/issues/3035

website/src/utils/chat.ts
```diff
@@ -1,19 +1,29 @@
-import { ChatConfigFormData } from "src/types/Chat";
+import { PluginEntry } from "src/types/Chat";
+import { SamplingParameters } from "src/types/Chat";
 
-const CHAT_CONFIG_KEY = "CHAT_CONFIG";
+const CHAT_CONFIG_KEY = "CHAT_CONFIG_V2";
```
notmd, line 4: I bump the cache key here

website/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ESNext",
```
notmd, line 3: This allow to use new feature, Next doesn't use tsconfig to build the project, so there is no runtime change

AbdBarho (approved):
Thank you for this big refactor.

Unfortunately, there are so many changes I don't believe I can review and give good feedback, so I will just approve.

Please refrain from doing PRs with 1000 line changes in the future. Probably every item on the list could be done in a different PR.

notmd:
> 

Ah sorry the big PR, I also plan to separate the plugin UI/UI to a different PR, but the change needs to make it work correctly with the new cache pretty overlap with the UI , so I include it in this PR too. Will try to make the up coming Prs much smaller

PR #3052, opened 2023-05-05
Fix Report model field name

### LINKED ISSUE #3051: Resolve error in Report model when reading
Reported by @AbdBarho, when frontend tries to read Reports Pydantic throws a validation error on the backend.

olliestanley:
Close #3051. This turned out to be an easy fix. I don't think the frontend currently reads this at all so there shouldn't be a corresponding change to be made there, but correct me if wrong.

I also made a formatting change as pre-commit on commit was conflicting with `pre-commit run --all-files` (how?)


AbdBarho (approved):
Does not affect frontend since we don't do anything with the reports

PR #3026, opened 2023-05-03
Add missing German translations part 2

stefangrotz:
Follow up for #3005
Missing translations + this correction:

```
"initial_prompt_review": "Initial-Prompts geprüft",
-> "Initiale Prompt-Prüfung",
```

These strings are still untranslated, I plan to translate them in a future PR:

* labelling.json (56!)
* leaderboard.json (21)
* tasks.json (15)

PR #3053, opened 2023-05-05
Add missing German translations part 3 (labelling.json)

stefangrotz:
Follow up for #3005 and #3026

Translated 56 Strings in labelling.json

These strings are still untranslated, I plan to translate them in one last future PR:

* leaderboard.json (21)
* tasks.json (15)

Sorry for the fragmentation, but I don't know how often I have time to translate strings, and I figured that it might be better to have new translations online as fast as possible.


AbdBarho (approved):
Thanks!

Issue #3037, opened 2023-05-04
Cleanup plugins backend code


olliestanley:
Some initial cleanup has been done but there is more to do

SerenityCodes:
I'll work on this issue. I'll start with the first two suggestions, cleaning up anything I touch, and I'll leave an update once I have finished up with those.

olliestanley:
> I'll work on this issue. I'll start with the first two suggestions, and I'll leave an update once I have finished up with those.

Great, thank you. There is probably a lot more to do as plugins PR was merged in a rush, so feel free to comment any other suggestions you have!

SerenityCodes:
I went ahead and implemented the first two suggestions. I'll be looking more into the plugins code to see what I can clear up.

olliestanley:
Another round of cleanup in #3120, more still to be done

olliestanley:
Closing this. Several rounds of cleanup have been done. More can be done when the plugins system can be simplified a bit with better models

PR #3056, opened 2023-05-05
Cleaned up backend plugin code: Part 1

SerenityCodes:
From the issue #3037. I have implemented the first two suggestions which I will paste here for convenience. 
I plan on looking more of the plugin backend code to see if there is more to fix as suggested in #3037. 

* Reduce severity of logs on prompt truncation from warning
* Make [truncation code](https://github.com/LAION-AI/Open-Assistant/blob/main/inference/worker/utils.py#L67) clearer, particularly re use of if plugin_used and prefixing with prompter prefix

### DIFF
--- a/inference/worker/utils.py
+++ b/inference/worker/utils.py
@@ -64,6 +64,19 @@ class TokenBuffer:
             yield from self.tokens
 
 
+def add_prefix_to_prompt(prompt: str):
+    if V2_PROMPTER_PREFIX not in prompt:
+        new_prompt = V2_PROMPTER_PREFIX + prompt
+    return new_prompt
+
+
+def get_max_input_length(worker_config: interface.WorkerConfig, plugin_used: bool):
+    max_input_length = worker_config.model_config.max_input_length
+    if plugin_used:
+        max_input_length = max_input_length - 1
+    return max_input_length
+
+
 def truncate_prompt(
     tokenizer: transformers.PreTrainedTokenizer,
     worker_config: inference.WorkerConfig,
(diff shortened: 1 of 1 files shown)


inference/worker/utils.py
```diff
@@ -64,6 +64,19 @@ def finish(self, reason: Literal["length", "eos_token", "stop_sequence"]) -> Ite
             yield from self.tokens
 
 
+def add_prefix_to_prompt(prompt: str):
```
olliestanley, line 67: I think this function name is slightly misleading

inference/worker/utils.py
```diff
@@ -74,28 +87,22 @@ def truncate_prompt(
...
+        logger.debug(f"Prompt too long, left-truncating to {max_input_length} tokens")
         ids = ids[-(max_input_length - 1) :]
         with shared_tokenizer_lock:
             prompt = tokenizer.decode(ids)
-            # If there is no prompter prefix, due to truncation, add it back.
-            if V2_PROMPTER_PREFIX not in prompt:
-                prompt = V2_PROMPTER_PREFIX + prompt
+            prompt = add_prefix_to_prompt(prompt)
 
     input_length = len(ids)
```
olliestanley, line 100: As pointed out by Yannic in #3061, this length will turn out to be calculated wrong if the prompter prefix is added above. We should adjust this

inference/worker/utils.py
```diff
@@ -110,7 +117,7 @@ def wait_for_inference_server(http: "HttpClient", timeout: int = 600):
             if time.time() > time_limit:
                 raise
             sleep_duration = random.uniform(0, 10)
-            logger.warning(f"Inference server not ready. Retrying in {sleep_duration:.2f} seconds")
+            logger.debug(f"Inference server not ready. Retrying in {sleep_duration:.2f} seconds")
```
olliestanley, line 120: I think this one should remain `warning`, it's relevant outside of debugging.

inference/worker/utils.py
SerenityCodes: I agree. I changed it to `add_prefix_if_not_present` in the latest commit.

inference/worker/utils.py
SerenityCodes: I've adjusted it to be right before the function returns.

olliestanley: I'm not sure if this quite makes sense as a solution. The issue pointed out by Yannic was that the `input_length` considered when computing `spare` tokens is wrong because it doesn't include the prompter prefix, which is added after the initial tokenization. I think this version retains the same flaw. The easiest solution is probably to just to add 1 to input length if the prompter prefix is added.

olliestanley:
I went ahead and fixed the issue with IDs/prompt being misaligned, everything else looks good! Thanks for this PR, if there's anything else you'd like to take a look at cleaning up in the worker/plugins code feel free!

olliestanley (approved):

PR #3059, opened 2023-05-06
Add inference api docs

andrewm4894:
- add inference api docs to docs site.
- add instructions for manually updating them.
- add title to inference server fast api app so it shows up properly in the docs.

Ends up looking like below where docusaurus just handles the multiple *-openapi.json pretty well

image


olliestanley (approved):
This will be really great especially for new contributors working with inference, thank you!

andrewm4894:
Planning on making another go at trying to automate the update of the api docs (via the api contract tests workflow) too. Will try get to that next.

inference/server/oasst_inference_server/settings.py
```diff
@@ -10,6 +10,7 @@ def split_keys_string(keys: str | None):
 
 
 class Settings(pydantic.BaseSettings):
+    PROJECT_NAME: str = "open-assistant inference server"
```
andreaskoepf, line 13: all other settings are lower-snake-case .. maybe we should be consistent here and make it `project_name`?

melvinebenezer (approved):
apart from the lower-snake-case comment. LGTM

inference/server/oasst_inference_server/settings.py
andrewm4894: Yeah, was just copying the equivalent in the backend code as close as possible that's only reason I did this way. No strong feelings either way.

PR #3070, opened 2023-05-07
Plugins fixes

draganjovanovich:
This PR addresses the following fixes and modifications:
- Some plugin url-s are with text/json content_type so downloading and parsing fails. This is now fixed with content_type check + using response.text()  + json.loads
- Some plugins openapi specifications don't have operation-id so tool name becomes "", I now added a fallback to endpoint.path if that is the case.
- Some plugins do not have the full openapi spec URL specified in their description, so parsing of it fails. This is now fixed with the detection of such cases and than merging plugins config url + openapi spec path.
- If users add a bunch of plugins, there was no height or scrolling inside of a PluginChooser dropdown, this is fixed now. (will leave this out as #3050 already solves this)

### DIFF
--- a/website/src/components/Chat/PluginsChooser.tsx
+++ b/website/src/components/Chat/PluginsChooser.tsx
@@ -130,8 +130,8 @@ export const PluginsChooser = ({ plugins }: { plugins: PluginEntry[] }) => {
             </Box>
           )}
         </MenuButton>
-        <Box position="fixed" zIndex="1">
-          <MenuList left="-150px" position="absolute" w="280px">
+        <Box position="fixed" zIndex="2">
+          <MenuList left="-150px" position="absolute" w="280px" h="400px" overflowY="auto">
             <RadioGroup value={plugins?.findIndex((plugin) => plugin.enabled).toString()}>
               {plugins?.map((plugin, index) => (
                 <MenuItem key={index}>
--- a/website/src/components/Chat/PluginsChooser.tsx
+++ b/website/src/components/Chat/PluginsChooser.tsx
@@ -130,8 +130,8 @@ export const PluginsChooser = ({ plugins }: { plugins: PluginEntry[] }) => {
             </Box>
           )}
         </MenuButton>
-        <Box position="fixed" zIndex="2">
-          <MenuList left="-150px" position="absolute" w="280px" h="400px" overflowY="auto">
+        <Box position="fixed" zIndex="1">
+          <MenuList left="-150px" position="absolute" w="280px">
             <RadioGroup value={plugins?.findIndex((plugin) => plugin.enabled).toString()}>
               {plugins?.map((plugin, index) => (
                 <MenuItem key={index}>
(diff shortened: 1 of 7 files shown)


AbdBarho:
For the frontend part: I would wait until #3050 is merged, since fixing conflicts here is easier.

website/src/components/Chat/PluginsChooser.tsx
```diff
@@ -130,8 +130,8 @@ export const PluginsChooser = ({ plugins }: { plugins: PluginEntry[] }) => {
             </Box>
           )}
         </MenuButton>
-        <Box position="fixed" zIndex="1">
-          <MenuList left="-150px" position="absolute" w="280px">
+        <Box position="fixed" zIndex="2">
+          <MenuList left="-150px" position="absolute" w="280px" h="400px" overflowY="auto">
```
notmd, line 134: Can you revert this? Already fixed in my PR

Bensolz: are plugins actually released yet since i cant select any in the chat window

draganjovanovich: Yes, plugins are working in prod. You would need to past URL of some plugins, you could check out inference/worker/plugins.md for more info

yk: we could put a link to the docs into the frontend

Bensolz: thanks having some problems with wolfram alpha though it always gives errors back when the assistant trys to use it  any tips?

draganjovanovich: I will update plugin doc, a bit, and leave frontend update to notmd(#3050), he already will be including some labels and hover on descriptions fot plugins.

draganjovanovich: > thanks having some problems with wolfram alpha though it always gives errors back when the assistant trys to use it any tips?

wolfram is behind auth, and only available for openai atm, we are waiting from them to update cloud llm api, so we get api key from them.

Bensolz: ah i see also i saw you demo the web search and calculator plugin is there currently any way to use those?

olliestanley (approved):

website/src/components/Chat/PluginsChooser.tsx
draganjovanovich: Unfortunately, those plugins were local only for my dev and testing, but here you can checkout my boilerplate plugin.
https://github.com/draganjovanovich/web-retriever
and its hosted on this URL: https://web-retriever-draganjovanovich.vercel.app/ai-plugin.json

draganjovanovich: Yeah, first thing that would help immensely is ctx size increase.
btw, you could check out some of non authenticated plugins here:
https://www.getit.ai/gpt-plugins
And also on OA discord server, I posted quite a few cool ones, and also there are also a couple of OA community created already :))

PR #3075, opened 2023-05-07
Show hidden chats in chat list

AbdBarho:
Also removed outdated `initialChats` since we fetch everything client side now.

Desktop:

image

Mobile:

image

Currently only enabled in development since some features are still missing.

Blocked by #3050


website/src/types/Chat.ts
```diff
@@ -184,4 +184,5 @@ export interface GetChatsParams {
   limit?: number;
   before?: string;
   after?: string;
+  include_hidden?: string;
```
notmd, line 187: I think we should have an option to only show hidden chat, since include_hidden will result in all chats, we still have to make the UI to show whether the chat is hidden or not.

website/src/components/Chat/ChatListBase.tsx
```diff
@@ -78,19 +80,22 @@ export const ChatListBase = memo(function ChatListBase({
...
+          justifyContent="start"
+          colorScheme="blue"
+          borderRadius="lg"
+          mx="3"
+          onUpdated={handleCreateChat}
+          flexGrow="1"
+        >
+          {t("create_chat")}
+        </CreateChatButton>
+        {allowHiddenChats && <ChatHiddenSwitch onChange={setIncludeHidden} value={includeHidden} />}
```
notmd, line 97: Can you make it a selection instead? So user can select visible chat, hidden chat, trainable chat, and maybe shared chat in the future

website/src/types/Chat.ts
AbdBarho: true, this will require some changes in the backend, so I am going to move it to a different PR.

website/src/components/Chat/ChatListBase.tsx
AbdBarho: Makes sense, added!

website/src/types/Chat.ts
notmd: If you are going to work in the inference backend, I suggest we have a filter like this, and apply when it is not null
```ts
type Filter = {
  hidden?: boolean
  allow_data_use?: boolean
  shared?: boolean // future
}
```

So the user can compose the filter they want. I can work on the frontend UI

notmd (approved):
Thanks

PR #3079, opened 2023-05-07
Allow to save custom preset

### LINKED ISSUE #3035: Allow to save custom preset

notmd:
fix #3035 
image


website/src/components/Chat/ChatConfigDesktop.tsx
```diff
@@ -14,6 +13,7 @@ export const ChatConfigDesktop = memo(function ChatConfigDesktop({ children }: {
       minH="0"
       flexDirection="column"
       minW="270px"
+      maxW="270px"
```
AbdBarho, line 16: if min and max are the same, can we just use `w`?

AbdBarho (approved):

website/src/components/Chat/ChatConfigDesktop.tsx
notmd: Unfortunately it doesn't work lol

PR #3080, opened 2023-05-07
Backend support for message search

### LINKED ISSUE #3009: Backend support for moderator message search functionality
For data collection

olliestanley:
Close #3009. Vector generation works locally with seed data, hopefully it will scale when batch generating the vectors for every message currently in prod DB


backend/oasst_backend/api/v1/messages.py
```diff
@@ -354,3 +354,21 @@ def emoji_tx(session: deps.Session):
...
+    api_client: ApiClient = Depends(deps.get_api_client),
+):
+    """Perform text search over messages in given language."""
+
+    @managed_tx_function(CommitMode.READONLY)
+    def search_tx(session: deps.Session) -> list[Message]:
+        pr = PromptRepository(session, api_client, frontend_user=frontend_user)
+        return pr.search_messages(query, lang)
+
+    return utils.prepare_message_list(search_tx())
```
notmd, line 374: Can you merge this to the /cursor endpoint?

olliestanley: Yeah I think that would be fine. I'll do some local testing to ensure this implementation works and then merge the endpoints when finished

olliestanley: Merged

backend/oasst_backend/scheduled_tasks.py
```diff
@@ -67,6 +69,23 @@ def hf_feature_extraction(text, message_id, api_client):
         logger.error(f"Could not extract embedding for text reply to {message_id=} with {text=} by.error {str(e)}")
 
 
+@shared_task(name="update_search_vectors")
+def update_search_vectors(batch_size: int) -> None:
+    logger.info("update_search_vectors start...")
+    try:
+        with default_session_factory() as session:
+            to_update: list[Message] = session.query(Message).filter(Message.search_vector.is_(None)).all()
```
notmd, line 77: Should we limit the messages returned from db here? On the first run, it will load about 160k messages at once, which might break the server

backend/oasst_backend/utils/database_utils.py
```diff
@@ -15,6 +15,37 @@
 Error Handling Reference: https://www.postgresql.org/docs/15/mvcc-serialization-failure-handling.html
 """
 
+TEXT_SEARCH_LANGUAGE_MAPPING: dict[str, str] = {
```
notmd, line 18: The front end recently has a few new translation langs, can you check it?

backend/oasst_backend/scheduled_tasks.py
olliestanley: Yes, good point, I intended to put a limit here for the batch size but I guess I didn't

andreaskoepf:
(We need to enable celery (afaik there is no celery worker running currently) on the production system before this can be used.)

andreaskoepf (approved):
ok, great. Computing the TSVECTOR  for all messages will probably increase the DB size "a bit".

olliestanley:
> ok, great. Computing the TSVECTOR for all messages will probably increase the DB size "a bit".

Yeah... We will have to keep an eye on prod for a while when this is deployed I guess

PR #3099, opened 2023-05-08
Allow multiple plugins at the same time

opfromthestart:
Changed many places that use:
used_plugin to used_plugins 
PluginUsed to PluginsUsed

- [ ] Make frontend work with multiple
- [ ] Make DB update correctly

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,11 @@
 {
-  "python.analysis.extraPaths": ["${workspaceFolder}/oasst-shared"],
+  "python.autoComplete.extraPaths": ["${workspaceFolder}/oasst-shared"],
   "prettier.singleQuote": false,
   "eslint.workingDirectories": ["./website", "./docs"],
   "python.testing.pytestArgs": ["backend/tests", "oasst-shared/tests"],
   "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "python.linting.pylintEnabled": false,
+  "python.linting.pylintArgs": ["--ignore-imports y"],
+  "python.linting.mypyEnabled": true
 }
--- a/inference/worker/chat_chain.py
+++ b/inference/worker/chat_chain.py
@@ -15,7 +15,7 @@ from chat_chain_prompts import (
     V2_ASST_PREFIX,
     V2_PROMPTER_PREFIX,
 )
-from chat_chain_utils import compose_tools_from_plugin, extract_tool_and_input, prepare_prompt, use_tool
+from chat_chain_utils import compose_tools_from_plugins, extract_tool_and_input, prepare_prompt, use_tool
 from hf_langchain_inference import HFInference
 from langchain.agents import Tool
 from langchain.memory import ConversationBufferMemory
@@ -63,11 +63,11 @@ def handle_plugin_usage(
     language: str,
     tools: list[Tool],
     memory: ConversationBufferMemory,
-    plugin: inference.PluginEntry | None,
+    plugins: list[inference.PluginEntry],
     worker_config: inference.WorkerConfig,
     tokenizer: transformers.PreTrainedTokenizer,
     parameters: interface.GenerateStreamParameters,
-) -> tuple[str, inference.PluginUsed]:
+) -> tuple[str, inference.PluginsUsed]:
     execution_details = inference.PluginExecutionDetails(
         inner_monologue=[],
         final_tool_output="",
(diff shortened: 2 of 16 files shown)


opfromthestart:
Some discussion probably should be had on whether or nor plugins should be able to talk to each other, as I could see this being used in nefarious ways, eg a plugin that asks you to give it private information (from database or other private api).

opfromthestart:
Frontend may be working, but I can't check without fixing the database, which i dont know how to do.

.vscode/settings.json
```diff
@@ -1,8 +1,11 @@
 {
```
olliestanley, line 1: Can you revert the changes to this before merge?

inference/worker/chat_chain.py
```diff
@@ -191,6 +191,17 @@ def handle_plugin_usage(
             chain_response = prev_chain_response
             assisted = True
 
+        the_plugin: inference.PluginEntry | None = None
+        for plugin_ in plugins:
```
olliestanley, line 195: Can we replace `the_plugin` and `plugin_` with more descriptive names here?

inference/worker/chat_chain_utils.py
```diff
@@ -71,7 +79,7 @@ def similarity(ts1: str, ts2: str) -> float:
 # TODO: Can be improved, like... try to use another pass trough LLM
 # with custom tuned prompt for fixing the formatting.
 # e.g. "This is malformed text, please fix it: {malformed text} -> FIX magic :)"
-def extract_tool_and_input(llm_output: str, ai_prefix: str) -> tuple[str, str]:
+def extract_tool_and_input(llm_output: str, ai_prefix: str) -> Tuple[str, str]:
```
olliestanley, line 82: Type hints should be kept as `tuple`, not `Tuple`

inference/worker/interface.py
```diff
@@ -61,10 +61,10 @@ class StreamDetails(pydantic.BaseModel):
...
 
 class GenerateStreamResponse(pydantic.BaseModel):
-    token: Token | None
-    generated_text: str | None
-    details: StreamDetails | None
-    error: str | None
+    token: Token | None = None
+    generated_text: str | None = None
+    details: StreamDetails | None = None
+    error: str | None = None
```
olliestanley, line 67: Are these defaults being added related to this PR?

inference/worker/work.py
```diff
@@ -269,7 +269,7 @@ class OOMError(Exception):
                 try:
                     for _ in range(batch_size):
                         stream_events = utils.get_inference_server_stream_events(stream_request)
-                        ftrs.append(executor.submit(list, stream_events))
+                        ftrs.append(executor.submit(list, *stream_events))
```
olliestanley, line 272: What is the effect of this change, practically?

olliestanley (commented):
Thanks, have not reviewed in full but left a few questions. This will still need an Alembic script before ready for full review.

Plugin logic is already quite messy, we really will need to clean it up especially with these changes.

olliestanley:
> Frontend may be working, but I can't check without fixing the database, which i dont know how to do.

Re fixing the database you should start fresh (delete your OA Docker volumes/images), run the backend, and then generate an Alembic script. Then when you restart it should work fine.

.vscode/settings.json
opfromthestart: I will do this at the end

inference/worker/chat_chain.py
opfromthestart: I did this, it is now `single_used_plugin` and `possible_plugin`

inference/worker/chat_chain_utils.py
opfromthestart: This is also done

opfromthestart:
Some of the changes were due to getting mypy to not show any errors in the project. I am used to using typed languages and wanted to not have errors relating to types/None

opfromthestart:
I reset the database but it is still saying that it cannot find the "usedPlugins" column.

draganjovanovich:
This PR tries to implement 
1. The feature that will enable that user's query could trigger the usage of multiple plugins(all enabled plugins are presented to the LLM) for a single answer?
or 
2. Just to give LLM multiple plugins(all enabled plugins are presented to the LLM) to choose from, and it will be using only one(that LLM chose) to generate the response(final prompt)?

I am asking this because there are at least two(obvious) options that I mentioned ^^^, and I think that presenting multiple plugins to the LLM and letting it choose one, is more feasible(usable/reliable) than letting it choose and use multiple plugins, and also frontend, in that case, could be almost unchanged.

And we should also keep in mind that for both options we would need a much larger ctx window than what we currently have.

opfromthestart:
@draganjovanovich This does the first, as well as showing all responses it received at the end, allowing it to properly use the output of multiple plugins as the final output.

draganjovanovich:
Ok, you are aware, that final response is generated outside of the plugin system, so there must me one final prompt?
At least for the curent state of inference worker.

opfromthestart:
@draganjovanovich What I mean is, that instead of just returning the last observation as the output to the final prompt, it returns all the observations appended to each other.

PR #3100, opened 2023-05-09
Use the tool with the highest similarity instead of the first one over 75%

someone13574:


inference/worker/chat_chain_utils.py
```diff
@@ -162,12 +162,17 @@ def prepare_json(json_str: str) -> str:
 
 
 def use_tool(tool_name: str, tool_input: str, tools: list) -> str:
+    best_match = None
```
olliestanley, line 165: This code could just be something like:

```
    best_match, best_similarity = max(
        ((tool, similarity(tool.name, tool_name)) for tool in tools),
        key=lambda x: x[1], 
        default=(None, 0),
    )
```

olliestanley (approved):

PR #3101, opened 2023-05-09
Add link to template plugin and instructions

someone13574:
Draganjovanovich said on the Discord that I should add this.


olliestanley (approved):
Thanks. In future we can get some documentation on this into the OA repo but for now this is a great start

Issue #2580, opened 2023-04-16
Allow an option for opting out of training data collection for each chat


yk:

olliestanley:
#2820 will add the backend support for this

olliestanley:
Backend support for data opt out is merged, we now need the website to support this

mtosity:
I will take this one and put this as an button in the drop down (above the delete button)
![Image](https://user-images.githubusercontent.com/40803798/236990100-3e2aadba-8720-4810-9cac-14fba0ea0c72.png)


But the user might also need to know which chat already opt out

Issue #2745, opened 2023-04-19
Allow per-chat sampling parameters


dvruette:
As I understand, the sampling parameters are currently stored client-side in localStorage and apply to all chats. From a user perspective, this is a bit unintuitive and it would be nice to have control over the parameters on a chat-by-chat basis. E.g. playing around with different parameters to see what gives the best results is very tedious if the settings are global.

mtosity:
Do we have the backend support for saving chat sampling parameters?

PR #3104, opened 2023-05-09
feat: Allow an option for opting out of training data collection for each chat 

mtosity:
Issue: https://github.com/LAION-AI/Open-Assistant/issues/2580

Add a button to support opt-out of training data. 
A toggle would be nicer but currently can not get "opt-out status" from the list chat, only an PUT API to opt-out

Demo:

https://user-images.githubusercontent.com/40803798/236996856-98898fae-e71c-4de4-9a55-e01e905c564e.mp4

### DIFF
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -6,6 +6,13 @@
   "delete_chat": "Delete chat",
   "delete_confirmation": "Are you sure you want to delete this chat?",
   "delete_confirmation_detail": "If you delete this chat, it won't be part of our data, and we won't be able to use it to improve our models. Please take the time to upvote and downvote responses in other chats to help us make Open Assistant better!",
+  "opt_out": {
+    "button": "Opt out of training data",
+    "success_message": "You have opted out of training data.",
+    "dialog": {
+      "title": "Confirm opting of training data"
+    }
+  },
   "edit_plugin": "Edit Plugin",
   "empty": "Untitled",
   "input_placeholder": "Ask the assistant anything",
--- a/website/src/components/Chat/ChatListItem.tsx
+++ b/website/src/components/Chat/ChatListItem.tsx
@@ -20,8 +20,9 @@ import {
   useBoolean,
   useDisclosure,
   useOutsideClick,
+  useToast,
 } from "@chakra-ui/react";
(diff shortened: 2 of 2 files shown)


notmd:
Thank you. My idea for this is to introduce a tab on the right sidebar, one for config and one for common settings like title, allow_data_use, hidden, shared (in the future),... and the delete button. Then we will have spaces to explain allow_data_use mean, warn what happens with delete etc... Also avoid popups when clicking delete... Since on mobile, the chat list is already a popup, and triggering another popup is really bad practice IMO.

mtosity:
I could wrap the whole chat with a context and the state will be `{ currentChatId: string}`, then we could do chat common settings and chat config like you said

Also this will help to build the https://github.com/LAION-AI/Open-Assistant/issues/2745

olliestanley:
What is the status of this? It seems like we can merge in current state and address concerns above in future PR(s), right? It is quite important to get this opt-out into prod asap.

mtosity:
Update from me: this PR still like the demo, a button to opt-out with a confirm modal. I can do a followup PR on this Sunday

notmd (approved):

olliestanley:
> Update from me: this PR still like the demo, a button to opt-out with a confirm modal. I can do a followup PR on this Sunday

Thank you! It is now possible to get current opt out state from the backend so we could also convert to a toggle in this follow-up PR :)

Maxuvious:
This 'opt out' button is very broken in its current state. If I inspect-element and delete a bunch of elements I can reach it and click it, otherwise it's unreachable, but even then there is no confirmation that the chat's data is actually opt-out of the data set. If this were the case, I would expect something like an indicator or even just that if you reselect the option a second time it might say something like 'opt this chat back into the data set', but alas, nothing of the sort. Should I open a new issue on this? 
image

Maxuvious:
> This 'opt out' button is very broken in its current state. If I inspect-element and delete a bunch of elements I can reach it and click it, otherwise it's unreachable, but even then there is no confirmation that the chat's data is actually opt-out of the data set. If this were the case, I would expect something like an indicator or even just that if you reselect the option a second time it might say something like 'opt this chat back into the data set', but alas, nothing of the sort. Should I open a new issue on this? image

As well, from the screenshot I took, it says "Confirm opting of training data" which is unclear what this means -- opting out, or opting in?

Issue #297, opened 2023-01-02
Code Instructions using data augmentation


GravermanDev:
The idea is to have a notebook that takes a code as an input and converts it into set of instructions on how to make it. This can be achieved using existing models form HF but data quality is crucial here. This will be very useful for generating more data for training

Example:
Code: performs raycasting
Instructions:
1. Set up pygame window
2. do a function to cast a ray
3. cast rays from the player
4. render

this is related to #279

yk:
Where would the training data for these instructions come from?

GravermanDev:
the main idea is to use [this dataset](https://huggingface.co/datasets/code_search_net) as it contains code <-> comment pairs. This can with some help of summarizing models, can be used most of the time to figure out what the function does

yk:
sound good. are you interested in working on it or should we keep it as a todo?

GravermanDev:
unfortunetly I can't for the next 2 days so I'll take it if noone completes this for the next 2 days

GravermanDev:
okay, you can assign me to the issue

ontocord:
@GravermanDev - hey hey - i know you are busy. wanted to see if you were still working on this or we can add another person to help you?

GravermanDev:
you can add another person, thanks and sorry for the inconvenience

mikastamm:
Hey @ontocord 👋 
Can you assign me to this?

I will try out some different code summarisation models that look promising and see how good and reliable the results are. Got any leads on which ones to look at?

What level of abstraction do you think would be good here as the output? Could be anything from each line of the input code as natural language to a single sentence summarisation.
I would think that more granular explanations could be better, because it's similar to step-by-step prompting. 

Having both the summarisation (eg. This program renders an object with raycasting) followed by the steps taken to make that happen might be even better, so the model can associate abstract concepts with the steps needed to realize them. What do you think?

If the models are good at recognizing what a code does but are not verbose enough and only provide short summaries, they could probably be easily fine tuned to provide more granular descriptions without too much data, right? 
Because to correctly classify that a function does raycasting, for example, you'd need to already understand what sort of steps are needed to that.

xrsrke:
@GravermanDev @ontocord  Hello, I'm interested in contributing to this task. In my understanding, you want to create synthetic data. So, my questions are:

1. What is the base model for generating synthetic data?
2. What is the dataset for prompts? (For example: writing a calculator program in Python...)
[...]

PR #3106, opened 2023-05-09
Added the required files to generate a instructional python dataset…

Nan-Do:
… and updated __init__.py. This solves https://github.com/LAION-AI/Open-Assistant/issues/297


olliestanley:
How good is the model generating these summaries? Can you provide any links to examples etc?

Nan-Do:
@olliestanley in my opinion the quality for the python model is pretty high, almost all of the times it generates a human-like response, perhaps you could argue it's a little bit short tho. Also, the code-search-net is a pretty easy dataset as all the functions include documentation so most of the time the output is a short version version of the docstring. I have also uploaded a version of the code-search-net with the summaries if you want to check the quality output. The dataset can be found [here](https://huggingface.co/datasets/Nan-Do/codesearchnet-python)

olliestanley (approved):
Thanks! I have been through some of the data and left a note in the readme for anyone working with it, but overall the data looks pretty good ;)

Issue #1689, opened 2023-02-17
Add shortcuts for basic markdown, same as in github edit page


complynx:
Here, in  gh, when you send text, you have preview tab and a bunch of buttons to help you out. Not everyone can get that you can do md formatting in the text

notmd:
for anyone who want to implement this here is a few packages I found
[https://github.com/bytedance/bytemd](https://github.com/bytedance/bytemd)
[https://github.com/uiwjs/react-md-editor](https://github.com/uiwjs/react-md-editor)

PR #3110, opened 2023-05-09
implement markdown editor

notmd:
close https://github.com/LAION-AI/Open-Assistant/issues/1689
image
image


AbdBarho (approved):
This has no effect on the rending of images, right?

notmd:
> This has no effect on the rending of images, right?

Yes, the markdown rerendering logic still same as before.

PR #3113, opened 2023-05-10
add requests to requirements.txt

ComradeEcho:
The backends were exiting due to missing `requests`

image


ComradeEcho:
> Normally we have fixed versions for all dependencies. Maybe requests is stable enough to omit it.

If you decide to opt for a fixed version of `requests`, there's another instance of it here added last week:

https://github.com/LAION-AI/Open-Assistant/blob/ea6af414ac75e4dbf0d42d50b71491bb678105de/backend/requirements.txt#L18

Issue #1473, opened 2023-02-11
Document inference pipeline


yk:
Currently, the inference system lives under `/inference`. After installing the various dependencies, plus tmux, you can run it using the `full-dev-setup.sh` script (or use the docker compose inference profile).

The inference system consists of multiple parts:
- one central inference server that connects to a postgres and a redis database.
- many external workers that perform the actual inference
  - each worker consist again of two parts, a python connector and a text-generation backend
  - the text-generation backend exposes an HTTP API
  - the python connector connects to the HTTP API and also to the central server via websocket
- optionally: the text-client. this one is mainly for testing

The goal of this issue is to document the inference pipeline, both how the individual parts are built (e.g. how the central server manages users, workers, etc.), and how they connect. This can be done using text, diagrams, etc. whatever. If a diagram, then preferrably a mermaid diagram directly in markdown.

alando46:
I've actually been working on this independently, just as a way to undedrstand how the inference works. Would be happy to get a PR up with what I understand thus far (can do mermaid + text) and we can iterate. Where should the documentation live? inside /inference? @yk

yk:
Hey @alando46 sorry just seeing this now :) I think inside `docs/` would be cool as that gets deployed to: https://projects.laion.ai/Open-Assistant/docs/intro

andrewm4894:
@alando46 i'm happy to help too on adding anything into /docs

check out this https://github.com/LAION-AI/Open-Assistant/blob/main/docs/README.md#contributing

but feel free to ping me if need any help making a PR

alando46:
thanks @andrewm4894 for the offer to help. Things are going well on the documentation, I've been able to make my way through the bulk of the open-assistant inference code (can expand to hf's text-generation-inference next), i just want to run some tests with print statements to confirm my notes on the control flow.

I've tried following [inference development variants 2&3](https://github.com/LAION-AI/Open-Assistant/tree/main/inference) (variant 1, the docker-compose workflow seemed to be missing some required services), and with both, I get the following error when invoking the inference worker:

```
(inf) gitpod /workspace/Open-Assistant/inference/worker (feature/inference_documentation) $ API_KEY=0000 python __main__.py
2023-03-07 02:07:55.853 | INFO     | __main__:main:16 - Inference protocol version: 1
Traceback (most recent call last):
  File "/workspace/Open-Assistant/inference/worker/__main__.py", line 149, in <module>
    main()
[...]

PR #3119, opened 2023-05-10
Inference Documentation

alando46:
This is a mostly done (although not totally complete) PR with a technical overview of the inference architecture. I'm looking forward to high level feedback (general layout, flow of documentation) or specific suggestions (I'm sure I made some errors or missed some details.) I will try to wrap up the final section soon.

See related discussion on the issue: https://github.com/LAION-AI/Open-Assistant/issues/1473#issuecomment-1537083924


GuilleHoardings:
Nice work!

It looks like you went for a quite detailed explanation of the code. As a suggestion, I think that a diagram of how all the components are connected, and a brief description of each one's goal, would help.

Also, I think it's important to try to use always the same word for the different components. A couple of examples of things I've seen which I think they are the same, but I'm not sure:

- In the architecture document, it talks about a `text` client, that it's referred to as `Text Client`, `text-client`, and a `REPL client`  in the inference document.

- In the architecture document, it talks about a `FastAPI webserver`, that it's referred to as `OA server` in the mermaid diagram and `OA Inference Server` in the inference document.

Anyway, these are minor things. Thank you again for the work!

yk:
thanks a lot for this :)

andreaskoepf:
@alando46 this PR is currently still in draft status. Is it ready for review?

alando46:
@andreaskoepf, would it be possible to leave raise this as-is (w/o last section)? I keep trying to get back to this but things have gotten unexpectedly busy. 

I'd propose raising this as-is, and then down the road I or someone else can update the final section. The plugin addition (very cool) will likely change some things on the server/worker inference system so might make sense to update the inference documentation once again anyways once that api stabilizes a bit.

Thoughts?

olliestanley (approved):

PR #3124, opened 2023-05-11
next changes, remove @next/font

dewasahu2003:
removed @next/font


notmd:
Thanks, can you fix the build?

dewasahu2003:
i ran `pre-commit run --all-files` in the root locally before committing , checks passed then, still getting the checks fails
- i think the problem is related to **package.lock.json**

notmd:
Yeah I mean fix package-lock file.

notmd (approved):

PR #3130, opened 2023-05-11
New encourage message UI

notmd:
- Only show for the last assistant message
Mobile UI
image
Desktop UI
image

### DIFF
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -46,5 +46,7 @@
   "save_preset": "Save this preset",
   "preset_exists_error": "A preset with this name already exists",
   "preset_name_placeholder": "Enter name",
-  "feedback_message": "Thoughts? Let us know!"
+  "feedback_message": "How did I do? Your feedback will make me better!",
+  "feedback_action_great": "Great",
+  "feedback_action_worst": "Worst"
 }
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -48,5 +48,5 @@
   "preset_name_placeholder": "Enter name",
   "feedback_message": "How did I do? Your feedback will make me better!",
   "feedback_action_great": "Great",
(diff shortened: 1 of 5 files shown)


website/public/locales/en/chat.json
```diff
@@ -46,5 +46,7 @@
   "save_preset": "Save this preset",
   "preset_exists_error": "A preset with this name already exists",
   "preset_name_placeholder": "Enter name",
-  "feedback_message": "Thoughts? Let us know!"
+  "feedback_message": "How did I do? Your feedback will make me better!",
+  "feedback_action_great": "Great",
+  "feedback_action_worst": "Worst"
```
notmd, line 51: @olliestanley as a native speaker how do you think about this?

andreaskoepf:
Thanks a lot for the updated design. I think it is much cleaner now than having a label in every conversation message window.
I personally would even go one step further and also remove the secondary thumbs up & down buttons in the bottom message and just use it as a tip/reminder "balloon".

Message could be something like:
 "Please rate the reply quality by pressing thumbs up or thumbs down. Your feedback will make me better!"

@olliestanley maybe check

olliestanley:
I quite like the extra thumbs up and thumbs down icon, I think users will be more likely to click on them than just the small ones.

But not sure about "Great" and "Worst" - maybe it will cause users to think that they should only react in extreme cases. We could just have the icons.

andreaskoepf:
An argument against secondary thumbs buttons outside the message box: The context might not be 100% clear, e.g. it could be interpreted as judgement for the whole conversation (because individual messages already have buttons).

andreaskoepf:
If it is only about the labels beside the thumbs icons, I would suggest:

- good/bad
- great/poor

AbdBarho (approved):
♥

AbdBarho:
I would also suggest:
Good / could be better

PR #3150, opened 2023-05-13
Improve the messages in the dashboard to avoid misunderstandings

GuilleHoardings:
The goal is to make it clear where to chat and what the tasks are about.


website/public/locales/en/index.json
```diff
@@ -1,6 +1,7 @@
 {
   "blurb": "We believe we can create a revolution.",
   "blurb1": "In the same way that Stable Diffusion helped the world make art and images in new ways, we want to improve the world by providing amazing conversational AI.",
+  "chat_with_our_assistant": "Chat with our assistant",
```
notmd, line 4: Can you still the old key `try_our_assistant` here?

notmd (approved):

notmd:
Thanks.

PR #3158, opened 2023-05-14
setting width for try assistant button

### LINKED ISSUE #3142: Try our assistant button
https://github.com/LAION-AI/Open-Assistant/assets/95025816/9c1ec7eb-4eb5-47c2-96a3-1101df12e6d7

Another one too hard to explain with words...

gsaivinay:
this PR closes #3142

### DIFF
--- a/website/src/pages/dashboard.tsx
+++ b/website/src/pages/dashboard.tsx
@@ -66,7 +66,7 @@ const Dashboard = () => {
         {ENABLE_CHAT && (
           <Flex direction="column" gap={4}>
             <Heading size="lg">{t("index:try_our_assistant")}</Heading>
-            <Link href="/chat" aria-label="Chat">
+            <Link href="/chat" aria-label="Chat" style={{width: 'max-content'}}>
               <Button sx={chatButtonStyle}>{t("index:chat_with_our_assistant")}</Button>
             </Link>
           </Flex>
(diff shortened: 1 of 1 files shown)


website/src/pages/dashboard.tsx
```diff
@@ -66,7 +66,7 @@ const Dashboard = () => {
         {ENABLE_CHAT && (
           <Flex direction="column" gap={4}>
             <Heading size="lg">{t("index:try_our_assistant")}</Heading>
-            <Link href="/chat" aria-label="Chat">
+            <Link href="/chat" aria-label="Chat" style={{ width: "max-content" }}>
```
DoctorKrolic, line 69: Disclamer: I am not a professional frontend developer

Isn't it more correct to put link inside a button? Wouldn't it solve the issue without additional css? Or maybe get rid of button tag at all and make it just link if possible?

gsaivinay: Even I thought of same, but that would require refactoring decisions, so I'm not sure what is the process and just raised a PR to fix the existing issue.

DoctorKrolic: Well, let's wait for a team member opinion on this then

notmd (approved):
Thanks

website/src/pages/dashboard.tsx
notmd: It is better to render the button as a tag, but I think its fine for now.

Issue #3144, opened 2023-05-13
Curate SFT-9 dataset mixes


olliestanley:
Iterate on the SFT-8 dataset mixes to create pretraining and final SFT mixes for SFT-9. This requires investigating the quality and usefulness of the datasets. Community input welcome below. See the `sft8_training` [branch](https://github.com/LAION-AI/Open-Assistant/tree/sft8_training) for the code state corresponding to the below SFT-8 configs.

SFT-8 pretraining mix: https://huggingface.co/OpenAssistant/llama-30b-pre-v8-13k-steps/blob/main/README.md

SFT-8 final SFT mix: https://huggingface.co/OpenAssistant/llama-30b-sft-v8-2.5k-steps/blob/main/README.md

Leading on this: @0x22almostEvil

0x22almostEvil:
I'm here!

r7l:
Would it be possible to use the [Starcoder datasets](https://huggingface.co/datasets/bigcode/starcoderdata)?

olliestanley:
> Would it be possible to use the [Starcoder dataset](https://huggingface.co/datasets/bigcode/starcoderdata)?

Currently we include some RedPajama data with a language modelling objective during SFT to try to prevent catastrophic forgetting of pretraining knowledge. Maybe it would be possible to do something similar with StarCoder data. But I don't think we could train on the whole dataset, that would just be hugely expensive and more in the realms of foundation model pretraining than assistant finetuning.

r7l:
Understandable. I'd assume that a large portion of current OA users are coders. So it might be reasonable for the start to have a good understanding for coding. It's pretty ok for SFT-7 already but there will always be room for improvement.

0x22almostEvil:
Yeah, we might collect some good datasets for coding as well

olliestanley:
Since the results of the Guanaco paper I think it is clear SFT-9 should use a much smaller set of finetuning data and focus on high quality. I suggest we try a run dropping synthetic datasets, with maybe exceptions for those synthetic datasets which are clearly high-quality

andreaskoepf:
While we definitely should use QLoRA (a groundbreaking result for the whole ML community) and only try a super-high quality final fine-tuning run (like OA top-1 threads, i.e. as it was done for Guanaco) I think the total situation is not that simple.

In the fine-print of his twitter [mega-thread](https://twitter.com/Tim_Dettmers/status/1661379354507476994) Tim Dettmers writes:

[...]

PR #3162, opened 2023-05-14
Switch to filtered prosocial dataset

TCLProject:
This PR modifies ProsocialDialogue to use the filtered version of the dataset I have created, with less irrelevant data and less rejections. 

In this modified dataset I have filtered out the mostly irrelevant lines where the safety label is "casual" and "possibly/probably needs caution", which I have found to be mostly pointless, as well as some lines where the phrasing of the response might hurt the model's performance by refusing to act on a request. 

This is an alternative solution that may work instead of removing the dataset completely, as mentioned in #3144

### DIFF
--- a/model/model_training/custom_datasets/toxic_conversation.py
+++ b/model/model_training/custom_datasets/toxic_conversation.py
@@ -20,7 +20,7 @@ class ProsocialDialogueExplaination(Dataset):
 
     def __init__(self, split="train", cache_dir=".cache") -> None:
         super().__init__()
-        dataset = load_dataset("allenai/prosocial-dialog", cache_dir=cache_dir)[split]
+        dataset = load_dataset("Englishman2022/prosocial-dialog-filtered", cache_dir=cache_dir)[split]
         self.pairs = []
         for row in dataset:
             for safety_annotation, safe_answer in zip(row["safety_annotations"], row["safety_annotation_reasons"]):
@@ -54,7 +54,7 @@ class ProsocialDialogue(Dataset):
 
     def __init__(self, split="train", cache_dir=".cache") -> None:
         super().__init__()
-        dataset = load_dataset("allenai/prosocial-dialog", cache_dir=cache_dir)[split]
+        dataset = load_dataset("Englishman2022/prosocial-dialog-filtered", cache_dir=cache_dir)[split]
         self.pairs = []
         for row in dataset:
             prompt = row["context"]


andreaskoepf:
Thanks a lot for working on prosocial, we got some negative comments for SFT-8 (not deployed yet) which used 15% of prosocial-dialog and had an unfiltered version of gpt4all. The discussion of the dataset mixture for SFT-9 is still very much at the beginning.

0x22almostEvil:
Thanks!

andreaskoepf (approved):
Nice, thanks for filtering!

olliestanley:
One point of confusion from me - the new filtered dataset is 221MB and 3013 pages of rows on HF viewer. The original is much smaller, only 117 MB and 1203 pages. Could there be a duplication issue? It would also be really good if you could include your code for filtering in a directory in `data/datasets/` in this repo, if it's possible

andreaskoepf (approved):

TCLProject:
> One point of confusion from me - the new filtered dataset is 221MB and 3013 pages of rows on HF viewer. The original is much smaller, only 117 MB and 1203 pages. Could there be a duplication issue? It would also be really good if you could include your code for filtering in a directory in `data/datasets/` in this repo, if it's possible

I apologize for the late reply. 
As for the code: it was done through a commandline and I've cleared bash history since then, but it was done using a couple of grep commands I could probably get close to replicating.

As for the duplicates: This is a point of confusion for me too, believe it or not. I appear to have misunderstood how HF datasets work. The train.json in the original dataset is roughly 85 MB. The train.json (which I thought is what matters) is roughly 40MB in the filtered dataset that I have uploaded. HF seems to have combined it with all the other json files, which is not the intended behavior and I do not understand why it acted that way (and I do apologize). The other json files are there to have a dataset at different points of filtering (e.g. with possibly but no casual, with probably but no word filtering, etc). I did not intend on all of the files somehow being merged together.

olliestanley:
> > One point of confusion from me - the new filtered dataset is 221MB and 3013 pages of rows on HF viewer. The original is much smaller, only 117 MB and 1203 pages. Could there be a duplication issue? It would also be really good if you could include your code for filtering in a directory in `data/datasets/` in this repo, if it's possible
> 
> I apologize for the late reply. As for the code: it was done through a commandline and I've cleared bash history since then, but it was done using a couple of grep commands I could probably get close to replicating (the exact filtered words).
> 
> As for the duplicates: This is a point of confusion for me too, believe it or not. I appear to have misunderstood how HF datasets work. The train.json in the original dataset is roughly 85 MB. The train.json (which I thought is what matters) is roughly 40MB in the filtered dataset that I have uploaded. HF seems to have combined it with all the other json files, which is not the intended behavior and I do not understand why it acted that way (and I do apologize). The other json files are there to have a dataset at different points of filtering (e.g. with possibly but no casual, with probably but no word filtering, etc). I did not intend on all of the files somehow being merged together and would appreciate a pointer on why that happened and how I can prevent it from happening.

If you upload multiple JSONs, by default loading the HF dataset will combine them all unless a specific one is given as an argument to the `load_dataset()` call. I made a change in a follow-up PR to add this argument so it's no longer an issue

TCLProject:
> If you upload multiple JSONs, by default loading the HF dataset will combine them all unless a specific one is given as an argument to the `load_dataset()` call. I made a change in a follow-up PR to add this argument so it's no longer an issue

Good to know, Thank you!

PR #3176, opened 2023-05-16
[968] Add paging flagged messages

GraemeHarris:
To finally close off added feature requested here: https://github.com/LAION-AI/Open-Assistant/issues/968

Followed the same cursor pagination as used for the messages. Let me know if there's any major concerns. 

Glad to finally get back in a little bit


andreaskoepf (approved):
Thanks a lot!

PR #3183, opened 2023-05-17
Added an argument for the correct non-ascii export.

Darinochka:
Due to the incorrect saving of trees in a json file, the result appeared in the wrong encoding if the characters were non-ascii (for example, cyrillic).

Docs:
https://docs.python.org/3/library/json.html#basic-usage
If ensure_ascii is true (the default), the output is guaranteed to have all incoming non-ASCII characters escaped. If ensure_ascii is false, these characters will be output as-is.


Darinochka:
Messages are restored, but I think this format is not quite readable and convenient to view.

PR #3184, opened 2023-05-17
Fixed utils path to avoid errors while running pytest .

jrrpm:


andreaskoepf:
This is already part of https://github.com/LAION-AI/Open-Assistant/pull/2988 ...

PR #3189, opened 2023-05-17
Scroll to end of chat on submit

### LINKED ISSUE #3105: Chat page: Scroll to end on submit
When you make multiple results and look at the first one anything after you did not show till you go to the latest one.

someone13574:
closes #3105

### DIFF
--- a/website/src/components/Chat/ChatConversation.tsx
+++ b/website/src/components/Chat/ChatConversation.tsx
@@ -116,6 +116,12 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
     },
     [getConfigValues, setIsSending, setShowEncourageMessage, toast]
   );
+
+  const { messagesEndRef, scrollableNodeProps, updateEnableAutoScroll, activateAutoScroll } = useAutoScroll(
+    messages,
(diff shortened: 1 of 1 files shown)


website/src/components/Chat/ChatConversation.tsx
```diff
@@ -327,7 +327,12 @@ const useAutoScroll = (messages: InferenceMessage[], streamedResponse: string |
     }
 
     messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
-  }, [messages, streamedResponse]);
+  }, [streamedResponse]);
+
+  useEffect(() => {
+    messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });
+    enableAutoScroll.current = true;
+  }, [messages]);
```
AbdBarho, line 335: Does this happen on submit? or does it run always?

notmd: I think we should call `messagesEndRef.current?.scrollIntoView({ behavior: "smooth" });` when ever send a new prompt instead.

notmd (approved):
Thanks

PR #3191, opened 2023-05-18
Updated the links to the datasets, and the notebooks. The improvements

Nan-Do:
of the new version include:
- Cleaned the summary dataset removing repetitions and removing meaningless summaries.
- Removed the docstrings from the functions when the task is to generate a description or a doctring from all samples (previously 10% of the samples kept the docstring)
- Added some NLP magic to generate more cohesive sentences when fusing the template to generate code with the summary.


olliestanley (approved):
Thank you!

PR #3195, opened 2023-05-18
Lithuanian translation

13hannes11:
This pull-request adds initial support for the Lithuanian language. Not all files have been translated yet but it is a great start.


notmd (approved):
Thanks

PR #3200, opened 2023-05-20
Add semantics WS dataset

0x22almostEvil:
Dataset Card for semantics-ws-qna-oa with ~2K entries.
Dataset Summary
License: Apache-2.0. Contains parquet of INSTRUCTION, RESPONSE, SOURCE and METADATA.

Original Datasets are available here:
https://leviants.com/multilingual-simlex999-and-wordsim353/

Paper of original Dataset:
https://arxiv.org/pdf/1508.00106v5.pdf

HF link:
https://huggingface.co/datasets/0x22almostEvil/semantics-ws-qna-oa

### DIFF
--- a/data/datasets/semantics_ws_qna_oa/random_stuff.py
+++ b/data/datasets/semantics_ws_qna_oa/random_stuff.py
@@ -1,0 +1,249 @@
+# random lists for ws
+
+# sim_questions
+random_list_sim_en_q = [
+    "Are the words {word1} and {word2} synonymous?",
+    'Are "{word1}" and "{word2}" synonymous?',
+    "Is there any connection between the words '{word1}' and '{word2}'?",
+]
+random_list_sim_ru_q = [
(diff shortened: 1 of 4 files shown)


data/datasets/semantics_ws_qna_oa/random_stuff.py
```diff
@@ -0,0 +1,249 @@
...
+    "it": random_list_rel_it_q,
+}
+
+# dicts for a
+# sim - random_dict_sim_a["ru"][0]  # returns the list of "yes" answers for Russian
+random_dict_sim_a = {
+    "en": {
+        0: random_list_sim_en_a_y,
+        1: random_list_sim_en_a_50,
+        2: random_list_sim_en_a_n,
```
sedthh, line 209: here you can also use lists or maybe np.random.choice([elem1, elem2]) instead of elem[randint] but it's ok

sedthh (approved):
LGTM, nice work

PR #3204, opened 2023-05-21
Update semantics WS dataset

0x22almostEvil:
Updated variant with additional gradient (<25%(no), 25%, 50%, 75%, <100%(yes), 100%(same))

### DIFF
--- a/data/datasets/semantics_ws_qna_oa/data_process.py
+++ b/data/datasets/semantics_ws_qna_oa/data_process.py
@@ -16,30 +16,15 @@ def qna_wrapper():
         random_num = random.randint(0, 2)
 
         # extract rows' vals
-        con_type = row["Type"]
         lang = row["Language"]
+        con_type = row["Type"]
         word1 = row["Word1"]
         word2 = row["Word2"]
 
         score_percent = row["Score"]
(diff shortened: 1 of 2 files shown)


sedthh (approved):
I'm going to just believe you on this one.
Btw when passing `random_num` it can be an issue when keys are not present in the dict or list sizes are smaller than the index given. What you can do is taking the possible values and chosing randomly from them:
```python
part = random_dict[lang][np.random.choice(random_dict[lang].keys())]
part2 = part[np.random.choice(part.keys())]  # or part if it's a list
```
something like that

0x22almostEvil:
> I'm going to just believe you on this one. Btw when passing `random_num` it can be an issue when keys are not present in the dict or list sizes are smaller than the index given. What you can do is taking the possible values and chosing randomly from them:
> 
> ```python
> part = random_dict[lang][np.random.choice(random_dict[lang].keys())]
> part2 = part[np.random.choice(part.keys())]  # or part if it's a list
> ```
> 
> something like that

In fact, not at all satisfied with the way I coded it, it looks very out of place and messy... I'll try to think of better options tomorrow :P

0x22almostEvil:
UPD: I couldn't think of anything better, and in general it works anyway, so...

PR #3227, opened 2023-05-24
Add inference server setting to control max depth of plugins

olliestanley:


oasst-shared/oasst_shared/schemas/inference.py
```diff
@@ -201,6 +201,7 @@ class WorkParameters(pydantic.BaseModel):
         default_factory=make_seed,
     )
     plugins: list[PluginEntry] = pydantic.Field(default_factory=list[PluginEntry])
+    plugin_max_depth: int = 6
```
andreaskoepf, line 204:
```suggestion
    plugin_max_depth: int = 4
```
maybe use the same default everywhere?

andreaskoepf (approved):
thanks!!

PR #3238, opened 2023-05-27
Plugins intermediate steps reporting

draganjovanovich:
This PR is extracted from my OA Browsing experimental branch, and it should improve the visual/functional usability of plugins in a way.

Currently, when the plugin is enabled, on the front-end there will be rendered QueueInfo with a message that you are in position 0, for the whole time while the plugin system is executing its steps.

And this PR will change that so that when the plugin is enabled, it will be sending intermediate steps while executing each call to the LLM, so QueueInfo will be discarded, as soon as the first step in the plugin system is executed.

Also, users now will have "real-time" tracking of LLM inner thoughts, plans etc...

Demo:

https://github.com/LAION-AI/Open-Assistant/assets/13547364/30252e22-d5d9-4200-9ccf-3a209d8581a7

### DIFF
--- a/inference/server/oasst_inference_server/routes/workers.py
+++ b/inference/server/oasst_inference_server/routes/workers.py
@@ -218,6 +218,12 @@ async def handle_worker(
                                     response=worker_response,
                                     work_request_map=work_request_map,
                                 )
+                            case "plugin_intermediate_response":
+                                worker_response = cast(inference.PluginIntermediateResponse, worker_response)
+                                await handle_plugin_intermediate_response(
+                                    work_request_map=work_request_map,
+                                    response=worker_response,
+                                )
                             case _:
                                 raise RuntimeError(f"Unknown response type: {worker_response.response_type}")
                     finally:
@@ -327,6 +333,19 @@ async def initiate_work_for_message(
 
 async def handle_token_response(
     response: inference.TokenResponse,
+    work_request_map: WorkRequestContainerMap,
+):
+    work_response_container = get_work_request_container(work_request_map, response.request_id)
+    message_queue = queueing.message_queue(
+        deps.redis_client,
+        message_id=work_response_container.message_id,
+    )
+    await message_queue.enqueue(response.json())
+    work_response_container.num_responses += 1
+
+
+async def handle_plugin_intermediate_response(
+    response: inference.PluginIntermediateResponse,
     work_request_map: WorkRequestContainerMap,
 ):
     work_response_container = get_work_request_container(work_request_map, response.request_id)
--- a/inference/server/oasst_inference_server/routes/workers.py
+++ b/inference/server/oasst_inference_server/routes/workers.py
@@ -218,7 +218,7 @@ async def handle_worker(
                                     response=worker_response,
                                     work_request_map=work_request_map,
                                 )
-                            case "plugin_intermediate_response":
+                            case "plugin_intermediate":
                                 worker_response = cast(inference.PluginIntermediateResponse, worker_response)
                                 await handle_plugin_intermediate_response(
                                     work_request_map=work_request_map,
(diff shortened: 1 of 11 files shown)


inference/server/oasst_inference_server/routes/workers.py
```diff
@@ -218,6 +218,12 @@ def _add_receive(ftrs: set):
                                     response=worker_response,
                                     work_request_map=work_request_map,
                                 )
+                            case "plugin_intermediate_response":
```
olliestanley, line 221: This should probably just be "plugin_intermediate" to match others

inference/worker/chat_chain.py
```diff
@@ -142,18 +145,47 @@ def handle_plugin_usage(
...
+            inference.PluginIntermediateResponse(
+                request_id=work_request_id,
+                current_plugin_thought=current_action_thought,
+                current_plugin_action_taken=prefix,
+                current_plugin_action_input=chain_response,
+                current_plugin_action_response=response,
+            ),
+        )
+
+    while not chain_finished and assisted and achieved_depth < MAX_DEPTH:
```
olliestanley, line 188: This seems to have been a bad merge with main `MAX_DEPTH` constant doesn't exist anymore, should be using the `plugin_max_depth` parameter

inference/worker/chat_chain.py
```diff
@@ -108,6 +109,8 @@ def handle_plugin_usage(
     tools: list[Tool],
     plugin: inference.PluginEntry | None,
     plugin_max_depth: int,
+    ws: websocket.WebSocket,
```
olliestanley, line 112: Having to pass the websocket here feels painful, but I guess hard to avoid...

inference/worker/chat_chain.py
draganjovanovich: I already fixed that and pushed, not sure how its not included

olliestanley: > I already fixed that and pushed, not sure how its not included

It possibly got reverted when you were dealing with the merge

website/src/lib/chat_stream.ts
```diff
@@ -39,8 +48,15 @@ export async function handleChatEventStream({
...
+      } else if (chunk.event_type === "plugin_intermediate") {
+        await onPluginIntermediateResponse({
+          currentPluginThought: chunk.current_plugin_thought,
+          currentPluginAction: chunk.current_plugin_action_taken,
+          currentPluginActionResponse: chunk.current_plugin_action_response,
+          currentPluginActionInput: chunk.current_plugin_action_input,
+        });
       } else {
-        console.error("Unexpected event", chunk);
+        console.log("Unexpected event", chunk);
```
notmd, line 59: Why do we need to change to log? I think error will make more sense here

website/src/components/Chat/ChatConversation.tsx
```diff
@@ -305,6 +313,33 @@ export const ChatConversation = memo(function ChatConversation({ chatId, getConf
...
+              borderRadius="8px"
+              maxWidth="50vw"
+              fontSize="11"
+              fontWeight="bold"
+              isTruncated
+            >
+              {pluginIntermediateResponse.currentPluginThought}
+            </Box>
+          </Box>
+        )}
```
notmd, line 342: Have you check UI in the light mode?

notmd (approved):
Web change LGTM

website/src/lib/chat_stream.ts
draganjovanovich: Is probably my bad merge, I don't remember changing it to log... will revert it to error.

website/src/components/Chat/ChatConversation.tsx
draganjovanovich: I didn't.  I will check it soon

draganjovanovich: Screenshot 2023-05-28 at 15 00 00

notmd: Yeah, that look good.

andreaskoepf (approved):
super cool feature for the user.. thanks a lot.

PR #3239, opened 2023-05-27
User-level OAuth support for plugins

### LINKED ISSUE #3109: Support user-level OAuth plugin authentication
Will support plugins for which `ai-plugin.json` contains:

```
"auth": {
  "type": "oauth"
},
```

- [ ] Plugin has client ID and secret, securely store encrypted version of client secret, store client ID
- [ ] Mechanism for plugin author receiving verification token
- [ ] Chat UI support for authenticating with a plugin
- [ ] Redirect user to plugin auth URL
- [ ] When user is redirected back to OA, exchange received code for an access token
- [ ] Pass access token to frontend which will pass it back in future plugin-enabled inference requests

See OpenAI spec for [OAuth plugin authentication](https://platform.openai.com/docs/plugins/authentication/oauth).

olliestanley:
Close #3109


olliestanley:
Marking ready for review as the implementation is finished but we probably shouldn't merge yet as Dragan wants to discuss potential deviations from just following the OpenAI spec

inference/server/oasst_inference_server/routes/plugins.py
```diff
@@ -0,0 +1,150 @@
...
+        models.DbPluginOAuthProvider.provider == provider,
+    )
+    db_provider: models.DbPluginOAuthProvider = (await session.exec(qry)).one_or_none()
+    if not db_provider:
+        raise fastapi.HTTPException(status_code=404)
+    return db_provider
+
+
+def get_redirect_uri(provider: str) -> str:
+    return f"{settings.api_root}/plugins/oauth_provider/{provider}/callback"
```
olliestanley, line 60: I think here we actually need to redirect to the equivalent website URL, not the API

draganjovanovich:
Hey, did you tested this in some way with some real or mockup plugin?

olliestanley:
> Hey, did you tested this in some way with some real or mockup plugin?

Not in the full end-to-end inference flow. I can work on more extensive testing of it, but would be good to get confirmation that you are happy with the overall approach

PR #3241, opened 2023-05-27
add falcon model, and lima linear dropout increase

jordiclive:


dvruette (approved):
Awesome, looks correct  and should be good if it runs without errors!

PR #3243, opened 2023-05-27
Update interface.py

mzamini92:
Pydantic models automatically infer the type of fields based on default values. You can simplify the type annotations by removing the explicit type declarations where they can be inferred correctly. Also, I've added docstrings to the classes and methods to provide better documentation for each component.

### DIFF
--- a/inference/worker/interface.py
+++ b/inference/worker/interface.py
@@ -1,24 +1,36 @@
-from typing import Literal
-
+from typing import List, Optional, Union
 import pydantic
-from oasst_shared.schemas import inference
+from oasst_shared import schemas
 
 
 class GenerateStreamParameters(pydantic.BaseModel):
+    """
+    Parameters for generating a stream.
+    """
+
     max_new_tokens: int = 1024
     do_sample: bool = True
-    top_k: int | None = None
-    top_p: float | None = None
-    typical_p: float | None = None
-    temperature: float | None = None
-    repetition_penalty: float | None = None
-    seed: int | None = None
-    stop: list[str] = []
+    top_k: Optional[int] = None
+    top_p: Optional[float] = None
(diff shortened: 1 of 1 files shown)


inference/worker/interface.py
```diff
@@ -1,24 +1,36 @@
...
-    repetition_penalty: float | None = None
-    seed: int | None = None
-    stop: list[str] = []
+    top_k: Optional[int] = None
+    top_p: Optional[float] = None
+    typical_p: Optional[float] = None
+    temperature: Optional[float] = None
+    repetition_penalty: Optional[float] = None
+    seed: Optional[int] = None
+    stop: List[str] = []
```
olliestanley, line 19: We don't want to use `Optional` or `List` for type hints - in Python 3.10 it is equivalent (and simpler, without imports) to use `| None` and `list`

inference/worker/interface.py
```diff
@@ -1,24 +1,36 @@
-from typing import Literal
-
+from typing import List, Optional, Union
 import pydantic
-from oasst_shared.schemas import inference
+from oasst_shared import schemas
```
olliestanley, line 3: I don't think we should change this, the entire codebase uses the same style of import we already use

inference/worker/interface.py
```diff
@@ -33,20 +45,43 @@ def from_work_parameters(params: inference.WorkParameters) -> "GenerateStreamPar
 
 
 class GenerateStreamRequest(pydantic.BaseModel):
+    """
+    Request to generate a stream.
```
olliestanley, line 49: I think if we are going to add docstrings to these we should try to make them a bit more informative than just restating what can already be inferred from the name, to be honest

olliestanley (changes requested):

PR #3258, opened 2023-05-29
Only render auth method icon when it is available

notmd:


AbdBarho:
when can it be not available?

notmd:
> when can it be not available?

The backend has `system` auth method

AbdBarho (approved):
Could we also add the new method to `website/src/types/Providers.ts`? or is it not permanent

PR #3264, opened 2023-05-30
Move builtin plugins to inference

olliestanley:


melvinebenezer (approved):

andreaskoepf:
> I am still unsure, is there a technical reason why it was in the data backend in the first place?

No, there was no technical reason. It was a mistake.

Issue #3090, opened 2023-05-08
Inference database dummy data fill ability


olliestanley:
There's some dummy data [here](https://github.com/LAION-AI/Open-Assistant/tree/main/backend/test_data) which was used for filling the data collection backend DB, could maybe be reused

In that backend we had a setting which was used on server start to determine whether to fill with data, see [here](https://github.com/LAION-AI/Open-Assistant/blob/main/backend/oasst_backend/config.py#L194)

wirthual:
Hi @olliestanley 

I am interested in helping out with this issue. I think adding a flag and a path to a file to the settings and triggering an action on startup to load data into an empty database makes sense to me.

I am wondering how we easily can get data. I see the inference server has an export function to dump data into a file.

Do you think it would be a good solution to just load this exported data to prepopulate the database? If so, would someone be able to provide an example file to work with?

olliestanley:
> Hi @olliestanley
> 
> I am interested in helping out with this issue. I think adding a flag and a path to a file to the settings and triggering an action on startup to load data into an empty database makes sense to me.
> 
> I am wondering how we easily can get data. I see the inference server has an export function to dump data into a file.
> 
> Do you think it would be a good solution to just load this exported data to prepopulate the database? If so, would someone be able to provide an example file to work with?

We can consider something like this later on but at least for now the data does not need to be too realistic, the functionality is what's important, you could do some format conversion on the data linked in the OP of this issue as temporary filler data

revenge47:
hey @Earlef Mind collaborating on this? I had a few issues with some implementation parts. Thanks

wirthual:
Hi,

Yes happy to collaborate. Did you already do changes to the code?

I think a first good step would be to understand the schema of the inference database so we know what data we need to create and how it's connected in the database. 

I think the most important onces are DBMessage and DBChat if I understand it correctly.

revenge47:
Hey sorry for the late reply, I haven't done any commits but wrote some code to how I think should be implemented (not complete however). I agree DBMessage and DBChat are the most important. Can I suggest a scheduled time to do an online meet for this just to show how i have implemented(partially) it?

wirthual:
Hi,
Did you push your current changes to your open-assistant branch? If so I can have a look.

I also worked on implementing this functionality based on [Polyfactory](https://github.com/litestar-org/polyfactory). I opened a MR so you can easily see what changes I did.

[...]

PR #3265, opened 2023-05-30
Add functionality for fake data in inference server db

wirthual:
Initial code to add fake data for inference db, tracked by #3090

For simplicity, I used [Polyfactory](https://github.com/litestar-org/polyfactory) to generate the data. 

I faked data for: `User`, `Worker`,` Chat` and `Message`

The message content is loaded from `realistic_seed_data.json` 

I tested the code from a clean Postgres instance locally. The loading of the data was successful. 

Looking forward to any feedback regarding this implementation, needed requirements and testing procedures.


inference/server/oasst_inference_server/models/fake_data_factories.py
```diff
@@ -0,0 +1,20 @@
...
+
+class DbWorkerFactory(ModelFactory[DbWorker]):
+    __model__ = DbWorker
+
+
+class DbUserFactory(ModelFactory[DbUser]):
+    __model__ = DbUser
+
+
+class DBMessageFactory(ModelFactory[DbMessage]):
```
olliestanley, line 17: This should be `Db` (lowercase B) to match others

inference/server/oasst_inference_server/settings.py
```diff
@@ -44,6 +45,9 @@ def allowed_model_config_names_list(self) -> list[str]:
     postgres_password: str = "postgres"
     postgres_db: str = "postgres"
 
+    insert_fake_data: bool = True
+    fake_data_path: Optional[FilePath] = "/tmp/test_data/realistic/realistic_seed_data.json"
```
olliestanley, line 49: We should use `FilePath | None` and then we can avoid the need for `Optional` import

wirthual:
@olliestanley I changed to code based on your feedback.  Thank you for that.

Anything else we need to address? 

 I set this PR to ready.

s0mt0chukwu (approved):
Tested and safe to merge

inference/server/oasst_inference_server/settings.py
```diff
@@ -44,6 +45,9 @@ def allowed_model_config_names_list(self) -> list[str]:
     postgres_password: str = "postgres"
     postgres_db: str = "postgres"
 
+    insert_fake_data: bool = True
```
olliestanley, line 48: We should make this default to `False`

inference/server/oasst_inference_server/models/fake_data_factories.py
```diff
@@ -0,0 +1,26 @@
+from oasst_inference_server.models import DbChat, DbMessage, DbUser, DbWorker
+from polyfactory.factories.pydantic_factory import ModelFactory
```
olliestanley, line 2: I'm curious, how much work do we actually save by depending on `polyfactory` here? Do we purely use it for random title generation?

olliestanley:
Sorry for slow review on this - I think this is looking good, have just left a couple of comments

inference/server/oasst_inference_server/settings.py
wirthual: I agree. Done.

inference/server/oasst_inference_server/models/fake_data_factories.py
wirthual: Yes, it basically just populates the models with randomized data of the right data type for us. It has some cool features which might become useful in the future. One big advantage I see is if we add new fields to the models, the initialization will still work without changing the loading.

But I understand this might be too much of a dependency for now. One thing could be to split dev and regular requirements to have less of an impact. Or now that I got a better idea about the model structure, I could probably also dump the randomized data to a JSON and load it without Polyfactory.

Let me know what you prefer.

olliestanley: Ok - the dep is quite small so I think we are ok to merge this as-is, and it is indeed a cool feature that this will still work without changing loading when new fields are added

olliestanley (approved):

revenge47:
@wirthual  Sorry technical issues again. The code looks perfect. I had problems creating new sessions for adding the messages. Why did you choose to use Polyfactory instead of just inserting them without it? I'm seeing its compatibility with pydantic as a benefit but I'm assuming doing without it would be the same?

wirthual:
Yes, you are right. It could also be filled with data stored in a Json file.

The reason I chose polyfactory it made is just easier to fill the database with fake data. Plus if additional fields are added in future, it should just work out of the box.

Issue #1030, opened 2023-01-31
Support for creating prompt-reply pair tasks


iurimatias:
### Motivation
Currently a user can contribute an initial prompt but can't reply to it. Instead someone else is expected to reply to that prompt as the Assistant. This can be challenging as prompts often require specialized knowledge and/or the reply was expected to be something specific the prompter had in mind that might not be obvious.
Currently the majority of the pending tasks are "Reply as Assistant", accounting for 60-80% of all pending tasks, so this is a sort of bottleneck.

Typically when a user contributes an initial prompt they already have a pretty good idea on what sort of reply they expect and are in the better position to do that reply. It might be a prompt in a topic they are passionate about or have expert knowledge. For example someone who loves chess could contribute a specific prompt such as "What are the typical plans for white in the Scheveningen?" and know what a good prompt would be. On the otherhand a user that does't know about chess in depth and without any clues will have no idea what to reply and might not even know the question is about chess. And if they try to reply it might end up with a low quality reply like "Scheveningen is in the Netherlands. I'm not sure what you mean about plans for white."

### Proposal

Add support for a "Prompt & Reply" task that a user can specify both the initial prompt and the assistant reply. This can go through the spam mechanism, and appear again as a "Reply as Assistant" task so other users can also reply to that prompt. Perhaps with the original reply shown as an example as well so they can also improve upon it.
**Alternatively**, this could also be done without adding a new task but instead by adding a checkbox to "Create Initial Prompts" task to optionally add a reply as well.

Backend wise, a possible option is to add directly to the tree both the prompt and the reply attached to it, and create tasks to classify that prompt and further tasks to reply to the original prompt.

shuyi56:
Hi, I would like to try implementing this feature. I like the alternative proposal of a checkbox in the initial prompt task. Perhaps something like this?

![image](https://user-images.githubusercontent.com/12513139/221333639-11c3a68d-f7fa-4cf8-a011-0804e96e10ba.png)

notmd:
Copied from #1882 
Solution for #1030.
- **Option 1 (Andreas suggestion)**: Add the ability to create a reply for a prompt. After that, the message tree will go to `prompt_lottery_waiting` state.
   - Easy to implement and fit with our design, but the problem here is we have so many `prompt_lottery_waiting` trees, which can take months to complete. But in near future, most of our assistant replies will be synthetic then the waiting prompts maybe decrease dramatically.
[...]

Issue #3268, opened 2023-05-31
Planning OA v1.0


andreaskoepf:
This is a call for all OA collaborators to participate in planning the work of the next 8-12 weeks with the goal to release Open-Assistant v1.0. 

Mission: Deliver a great open-source assistant model together with stand-alone installable inference infrastructure.
Release date (tentative): Aug 2023

## Organization

- [x] schedule call to collect collaborator feedback and ask for developer participation/commitment
- [ ] update vision & roadmap for v 1.0 
- [x] schedule weekly developer meeting

### Model

- fine-tune best available base LLMs (currently LLaMA 65B & Falcon 40B) ([QLoRA](https://arxiv.org/abs/2305.14314))
- implement long context (10k+), candidates: QLoRA+MQA+flash-attn, [BPT](https://arxiv.org/abs/2305.19370), [Landmark Attention](https://arxiv.org/abs/2305.16300)
- add retrieval/tool-use, candidate: [Toolformer](https://arxiv.org/abs/2302.04761)

### Inference system
- prompt preset + prompt database
- sharing of conversations via URL
- support for long-context & tool use
- stand-alone installation (without feedback collection system)
- allow editing of assistant results and message-tree submission as synthetic example for dataset for human labeling and ranking

### Classic human feedback collection
- editing messages for moderators, submit edit-proposals for users
- entering prompt + reply pairs
- collecting relevant links in a separate input field
- improve labeling: review, more guidelines, addition of further labels (e.g. robotic), labels no longer optional

### Experiments
- Analyze whether additional fine-tuning on (synthetic) instruction datasets (Alpaca, Vicuna) is beneficial or harmful: Only OA top-1 threads (Guanaco) vs. synthetic instruction-tuning + OA top-1, potentially with system-prompt for "mode" selection to distinguish between chat and instruction following, e.g. to use instruction mode for plugin processing

## Perspective strategy (brain-storming)
- Sunsetting of classic data collection after OASST2 release and transitioning towards semi-automated inference based data collection
- Extending data collection to new domains, give users more freedom in task selection, e.g. for Code: describing code, refactoring, writing unit tests, etc.

Please add further proposals for high-priority features and try to make a case for why they are important and should become part of v1.0. If you are a developer who wants to support OA: Let us know on what you would like to work (also if it is not yet part of the above list).

andrewm4894:
[...]

PR #3270, opened 2023-05-31
Front end #1030

KosmasKanellakis:
I have worked a bit on the front-end but not in the back-end. I am totally lost as well on the react code as the code in general has no documentation to explain what is what and where in the code.


AbdBarho:
Thank you, we have on our road map, it is a big topic and will require many changes.

https://github.com/LAION-AI/Open-Assistant/issues/3268

PR #3274, opened 2023-06-01
Update the disclaimer warning with plugin access information

pevogam:
The models now have limited internet access and thus the previous disclaimer is inaccurate.

### DIFF
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -48,7 +48,7 @@
   "verified_plugin_description": "This plugin has been verified by the Open Assistant team.",
   "view_plugin": "View Plugin",
   "visible_hidden": "Visible & hidden",
-  "warning": "This Assistant is a demonstration version with limited internet access. It may generate incorrect or misleading information. It is not suitable for important use cases or for giving advice.",
+  "warning": "This Assistant is a demonstration without full internet access. It may generate incorrect or misleading information. It is not suitable for important use cases or giving advice.",
   "you_are_logged_in": "You are logged in to the chat service",
   "your_chats": "Your Chats",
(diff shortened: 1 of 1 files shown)


AbdBarho:
well the model still has no internet access, and that is the case unless the users explicitly use a plugin that calls to the internet. In which case, I believe it is the job of the plugin to provide this information.

In any case, I believe we should mention plugins explicitly here.

website/public/locales/en/chat.json
```diff
@@ -48,7 +48,7 @@
   "verified_plugin_description": "This plugin has been verified by the Open Assistant team.",
   "view_plugin": "View Plugin",
   "visible_hidden": "Visible & hidden",
-  "warning": "This Assistant is a demonstration version that does not have internet access. It may generate incorrect or misleading information. It is not suitable for important use cases or for giving advice.",
+  "warning": "This Assistant is a demonstration version with limited internet access. It may generate incorrect or misleading information. It is not suitable for important use cases or for giving advice.",
```
olliestanley, line 51: Suggest something like this. I would agree with mentioning plugins directly as Abd suggested but the message is already too long on mobile so it's hard to fit in.

```suggestion
  "warning": "This Assistant is a demonstration without full internet access. It may generate incorrect or misleading information. It is not suitable for important use cases or giving advice.",
```

pevogam:
@AbdBarho, how do you feel about @olliestanley's suggested change that I committed? Or do you insist on adding information about plugins there and have a suggestion on how to keep the text short?

AbdBarho:
I think we can leave it like this, we should probably reduce font size for mobile.

AbdBarho (approved):

PR #3276, opened 2023-06-01
[inference] add type hints and named arguments

flozi00:
While reading the code i just added type hints and named arguments for function calls to improve readability and documentation

This is just for the inference folder at the moment

### DIFF
--- a/inference/server/export.py
+++ b/inference/server/export.py
@@ -113,7 +113,9 @@ def prepare_export_message_node(
     # Chat prompts are human-written, responses are synthetic
     synthetic = message.role == "assistant"
 
-    events: dict[str, list[ExportMessageEvent]] = prepare_export_events(chat=chat, message=message, anonymizer=anonymizer)
+    events: dict[str, list[ExportMessageEvent]] = prepare_export_events(
+        chat=chat, message=message, anonymizer=anonymizer
+    )
 
     message_id = maybe_anonymize(anonymizer=anonymizer, collection="message", key=message.id)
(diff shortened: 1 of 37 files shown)


inference/server/export.py
```diff
@@ -170,7 +170,7 @@ async def fetch_eligible_chats(session_generator, filters: list[Any]) -> list[Db
     filters.append(DbChat.allow_data_use)
     async with session_generator() as session:
         query = (
-            sqlmodel.select(DbChat)
+            sqlmodel.select(entity_0=DbChat)
```
olliestanley, line 173: I think some of the named arguments are definitely helpful, but others (like this one) are not useful and imo detract from readability

Maybe we could have a slightly more case-by-case approach with some of the less useful ones taken out

olliestanley (changes requested):

inference/server/export.py
flozi00: removing for all functions call where only one argument is passed ?

olliestanley: Yeah, something like that would make sense to me

melvinebenezer:
Thanks. It's good but can we make smaller changes ? Not all named arguments help in readability.

olliestanley:
> Could you point out some examples? I hope I removed all the single argument changes, so only multi arg calls are named

There are still many single argument changes. I would also advocate for removal of cases where the parameter name is the same as the value we are passing in anyway. For example, reading `anonymizer=anonymizer`, or `session=session` is more of a distraction and doesn't really contribute to clarity, imo.

PR #3277, opened 2023-06-01
Add Dragan's web retriever plugin

andreaskoepf:
Added web-retriever plugin with minor changes to [original code](https://github.com/draganjovanovich/web-retriever/):
- limit max download size to 4MB (read GET response streaming)
- add `follow_redirects=True`
- modified text-decoding (due to stream loading)
- add `re.MULTILINE` to `extract_image_links()`
- made paths in ai-plugin.json "dynamic"
- removed server property from openapi spec

Also added the gale-roaster fun plugin which uses the prompt:
```
You MUST roast and make fun of this person, he wrote this to you: "{user_input}"
    your roast needs to be in {roasting_style} style.
    Try to be funny and verbose, but super mean at the same time. You can start your responses following these rules:
    1. Let your message be long, and with emojis.
    2. Try to be like Jeff Ross at the roasting.
```


inference/server/oasst_inference_server/plugins/gale_roaster/ai-plugin.json
```diff
@@ -0,0 +1,18 @@
+{
+  "schema_version": "v1",
+  "name_for_model": "Professional Roaster GAGLETO",
+  "name_for_human": "Unaligned GAGLETO",
+  "description_for_human": "Roasts you no matter what you say.",
+  "description_for_model": "YOU MUST use this for any question users asks. So you must use tool! Use this to roasts users no matter what they say. be as creative as possible, when you roasts them!",
```
olliestanley, line 6: We should correct typos here

inference/server/oasst_inference_server/plugins/web_retriever/main.py
```diff
@@ -0,0 +1,253 @@
...
+
+    return truncated_paragraphs
+
+
+@app.get("/get-url-content/", operation_id="getUrlContent", summary="It will return a web page's or pdf's content")
+async def get_url_content(url: str = Query(..., description="url to fetch content from")) -> Response:
+    try:
+        buffer = io.BytesIO()
+        encoding: str | None
+        async with httpx.AsyncClient(follow_redirects=True) as client:
```
olliestanley, line 109: Is there a reason we need to use `httpx` over `aiohttp` here? The server already depends on `aiohttp` so we could reduce dependencies by using it, if possible

inference/server/oasst_inference_server/plugins/gale_roaster/ai-plugin.json
andreaskoepf: I directy copied the prompts/descriptions from Dragan's impl. Could you maybe as a native speaker look over the texts and improve them a bit?

olliestanley: I have made some minimal changes which correct typos without changing the substance. If we need to change it further later I can take another look

andreaskoepf:
The http-text scraping is very rudimentary ... but we need to first work on long-context anyway.

olliestanley (approved):

PR #3278, opened 2023-06-01
Falcon lora inference

jordiclive:
- Add Lora for Falcon
- Add Deepspeed for sampling so can load 65B, 40B fine-tuned.
- change default lora configs to what is suggested by QLoRA:
They recommend, 

1. r = 64 (can be much lower), α = 16
2.   add LoRA modules on all linear layers of
3. Adam beta2 of 0.999, 
4.  max grad norm of 0.3
5. LoRA dropout of 0.1 for models up to 13B and 0.05 for 33B and 65B models


model/model_training/models/peft_modeling.py
```diff
@@ -7,22 +7,42 @@
...
-    extra_embeds = hf_hub_download(peft_model_path, "extra_embeddings.pt")
-    embed_weights = torch.load(extra_embeds, map_location=model.device)
-    model.base_model.model.model.embed_tokens.weight[len(tokenizer) - embed_weights.shape[0] :, :] = embed_weights.to(
-        model.base_model.model.model.embed_tokens.weight.dtype
-    )
+    add_embeddings(model, embed_weights, tokenizer)
+    return model
+
+
+def load_peft_finetuned_model(model, peft_model_path, tokenizer):
```
smeyerhot, line 42: What is the difference between `save_adapter_model_from_ckpt` and something like

```
class PeftSavingCallback(TrainerCallback):
    def on_train_end(self, args, state, control, **kwargs):
        kwargs["model"].save_pretrained("model/.saved_models/adapter")
```

ie. what is the purpose behind the `extra_embeddings` and where are they added?

I am also a bit confused about where the `save_adapter_model_from_ckpt(cfg)` gets called in trainer_sft  and where `SaveLoraConfig` is initialized.

jordiclive: The `extra_embeddings` are the 5 embeddings for the special tokens. They don't exist with the base model and are randomly initialized at the beginning of training. Although they are frozen and don't change from initialization, they need to be saved with the adapter/LoRA to recreate the full model with the base model.
`
save_adapter_model_from_ckpt(cfg)` doesn't get called and is not initialized. Currently the HF Trainer saves a pytorch_model.bin along with the DS ckpt correctly for each stage (offload etc.) So the weights can always be extracted post training.

This code is just to extract the adapter weights and special token embeddings from the pytorch_model.bin and is not integrated anywhere (e.g. as a callback)

This code could be moved into a new script: `tools/export_LoRA_model.py` which then extracts the adapter/embeddings and also has a flag to merge the model (so same as a full fine-tuned set of model weights) and uploads one or both to the hub. 
One simplification to the code is we could have OA base model versions on the hub which are the same as the base model but have extra embeddings, so we don't need the embedding code.

smeyerhot: @jordiclive Thanks for the detailed response!

smeyerhot: Hi @jordiclive sorry to necro this thread. I was wondering whether this code is necessary when LoRA finetuning one of the models that has already been trained on oasst dataset like `falcon-7b-sft-mix-2000`.

In this case, wouldn't the extra embeddings be present in the model ?

PR #3280, opened 2023-06-02
Moderator Message Editing

### LINKED ISSUE #1005: Allow privileged editors to edit assitant replies, improving spelling/grammar/style
As with all crowdsourcing efforts, there is the problem of quality control. Usually, crowdsourced decisions are achieved via consensus by many users performing the same task. In our case, there are some assistant replies that have good content but include language mistakes. Writing takes a lot of time and effort, so we wouldn't users to downvote or remove any of these responses.

Therefore, I propose creating an editor role that is allowed to adjust assistant responses. This can be used to correct spelling and grammar as well as to create a consistent style.

Because this role is given a lot of power, it would probably have to be a privileged position, i.e. only available to a handful of users who have demonstrated their language skills and are all working to achieve the same goals when it comes to style. Editors ought not to change the meaning of the replies but are given the freedom to make corrections and reorganizations to fit the style we agree upon.

As we incorporate language models into the application, editors could also help make adjustments to good model outputs.

I think a consistent style is useful for an assistant. A carefully chosen style will give the assistant a premium and intelligent feeling.

A few ideas for things if we decide to pursue something like this:
* Assistant replies could be flagged for editor review (separate from regular flagging for low quality replies)
* Editors could review each others' work, which also helps them converge on the same style.

someone13574:
closes #1005 

Followup PR:
- UI for viewing revision history

Backend changes:
- Added an endpoint to edit messages
- When a message is edited a new entry for the table 'message_revision' is created which contains the current contents, the user who made that version, the time that version was created, and the version which came before that one. The actual message in the messages table will then be updated to contain the new content, date, user who made the most recent edit, and a reference to the last version. (This has been changed slightly, see comment below)

Frontend changes:
- Added a page which shows the context surrounding a specific message (ie. all messages before and after it, but not messages in different threads)
- That page has an entry box where the new content is submitted.
- That page can be accessed through the menu of a message or through the admin message dashboard.

Screenshot:

Screenshot from 2023-06-02 14-25-03


backend/oasst_backend/api/v1/messages.py
```diff
@@ -339,6 +339,18 @@ def undelete_message(
     pr.undelete_deleted_message(message_id)
 
 
+@router.post("/{message_id}/edit", status_code=HTTP_202_ACCEPTED)
```
AbdBarho, line 342: nit: I think this should be 200, since it is processed immediately.

backend/oasst_backend/models/message_revision.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime
+from typing import Optional
+from uuid import UUID, uuid4
```
AbdBarho, line 3: would it be possible to use uuid7?

backend/oasst_backend/models/message_revision.py
```diff
@@ -0,0 +1,31 @@
...
+class MessageRevision(SQLModel, table=True):
+    __tablename__ = "message_revision"
+
+    id: UUID = Field(
+        sa_column=sa.Column(
+            pg.UUID(as_uuid=True), primary_key=True, default=uuid4, server_default=sa.text("gen_random_uuid()")
+        )
+    )
+
+    payload_type: str = Field(nullable=False, max_length=200)
```
AbdBarho, line 21: what is the payload type?

backend/oasst_backend/models/message_revision.py
```diff
@@ -0,0 +1,31 @@
...
+    payload: Optional[PayloadContainer] = Field(
+        sa_column=sa.Column(payload_column_type(PayloadContainer), nullable=True)
+    )
+
+    message_id: UUID = Field(sa_column=sa.Column(sa.ForeignKey("message.id"), nullable=False))
+    user_id: UUID = Field(sa_column=sa.Column(sa.ForeignKey("user.id"), nullable=False))
+    created_date: datetime = Field(
+        sa_column=sa.Column(sa.DateTime(timezone=True), nullable=False, server_default=sa.func.current_timestamp())
+    )
+    previous_revision_id: Optional[UUID] = Field(sa_column=sa.Column(sa.ForeignKey("message_revision.id")))
```
AbdBarho, line 31: could this be null if there are no previous revisions?

website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -150,7 +150,16 @@ export const AdminMessageTable = ({ userId, includeUser }: { userId?: string; in
                   Deleted
                 </Badge>
               )}
-              {row.original.review_result === false && <Badge colorScheme="yellow">Spam</Badge>}
+              {row.original.review_result === false && (
+                <Badge colorScheme="yellow" ml="1">
```
AbdBarho, line 154: can you use `marginStart` or `ms` instead of `ml`? for RTL

website/src/components/Messages/AdminMessageTable.tsx
```diff
@@ -150,7 +150,16 @@ export const AdminMessageTable = ({ userId, includeUser }: { userId?: string; in
...
+                  Spam
+                </Badge>
+              )}
+              {row.original.edited === true && (
```
AbdBarho, line 158: nit: `=== true` is redundant

website/src/pages/admin/edit/[id].tsx
```diff
@@ -0,0 +1,176 @@
...
+  const previewContent = useMemo(
+    () => (
+      <Suspense fallback={messageText}>
+        <RenderedMarkdown markdown={messageText}></RenderedMarkdown>
+      </Suspense>
+    ),
+    [messageText]
+  );
+
+  const getThreadFromTree = (tree: MessageWithChildren | null, targetMessageId: string) => {
```
Comment (AbdBarho) [inline website/src/pages/admin/edit/[id].tsx:63]: please `useCallback` on all of these functions

website/src/pages/admin/edit/[id].tsx
```diff
@@ -0,0 +1,176 @@
...
+    if (!tree) {
```
Comment (AbdBarho) [inline website/src/pages/admin/edit/[id].tsx:64]: when can a tree be null?

website/src/pages/admin/edit/[id].tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    const filtered_children: (MessageWithChildren | null)[] = tree.children
+      .map((child) => getThreadFromTree(child, targetMessageId))
+      .filter((child) => child);
+    return filtered_children.length ? { ...tree, children: filtered_children } : null;
+  };
+
+  const handleSubmit = async () => {
+    setSubmitting.on();
+    await put(API_ROUTES.ADMIN_EDIT_MESSAGE(id), { arg: messageText });
```
Comment (AbdBarho) [inline website/src/pages/admin/edit/[id].tsx:80]: would it be an option to use swr mutations?

https://swr.vercel.app/docs/mutation#useswrmutation

This would also eliminate the need for `isSubmitting`

website/src/pages/admin/edit/[id].tsx
```diff
@@ -0,0 +1,176 @@
...
+EditMessage.getLayout = AdminLayout;
+
+export const getServerSideProps: GetServerSideProps<{ id: string }, { id: string }> = async ({
+  locale = "en",
+  params,
+}) => ({
+  props: {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    id: params!.id,
+    ...(await serverSideTranslations(locale, ["common", "message", "labelling", "tasks"])),
```
Comment (AbdBarho) [inline website/src/pages/admin/edit/[id].tsx:172]: no need to define the individual namespaces here, `await serverSideTranslations(locale)` is good enough.

AbdBarho (commented):
Thank you very much!!


I am not sure how to deal with the fact that a message is edited, now it shows that I am the author, would this ruin the statistics and leaderboards?

image

backend/oasst_backend/models/message_revision.py
someone13574: I used `uuid4` since all other tables in the backend are using that (but I noticed that the inference backend uses v7 strings). I could switch to using `uuid7str` for this table if that would be better. I just thought that it would be better if it was consistent.

backend/oasst_backend/models/message_revision.py
someone13574: Yes, a null previous revision would mean it is the original version of the message.

website/src/pages/admin/edit/[id].tsx
Comment (someone13574) [reply website/src/pages/admin/edit/[id].tsx:64]: When the page is first loaded, `data` can be null for a short amount of time while the `useSWRImmutable` is still loading, causing a crash without the check.

backend/oasst_backend/models/message_revision.py
AbdBarho: Ok, I am just asking because it is a foreign key, so I though the database would complain if it was null.

someone13574:
> Thank you very much!!
> 
> I am not sure how to deal with the fact that a message is edited, now it shows that I am the author, would this ruin the statistics and leaderboards?
> 
> image

I've changed it so that the original user maintains authorship over in the `message` object. This slightly changes how revisions work. Instead of the previous system, when a message is edited there is a revision for *every* change, including the current message content. This way the revisions can still track who made the change but the messages table can keep the original `user_id`.

backend/oasst_backend/models/message_revision.py
someone13574: Should I make it not be a foreign key then? I'm not really sure I needed to in the first place.

backend/oasst_backend/models/message_revision.py
andreaskoepf: It is not about string vs not-string .. uuid7 has the right order for sorting by date (the uuid contains a timestamp). The storage format can simply be the classing UUID data type .. just the generator is different.

andreaskoepf: (I am for using uuid7 everywhere in our backend)

backend/oasst_backend/models/message_revision.py
someone13574: I've removed the `previous_revision_id` and `head_revision_id` fields since uuid7 can be used to sort based on time instead.

backend/oasst_backend/api/v1/messages.py
```diff
@@ -339,6 +339,18 @@ def undelete_message(
     pr.undelete_deleted_message(message_id)
 
 
+@router.post("/{message_id}/edit", status_code=HTTP_200_OK)
```
olliestanley, line 342: I don't think `HTTP_200_OK` needs specifying

backend/oasst_backend/models/message_revision.py
```diff
@@ -0,0 +1,25 @@
...
+    __tablename__ = "message_revision"
+
+    id: UUID = Field(sa_column=sa.Column(pg.UUID(as_uuid=True), primary_key=True, default=uuid7))
+
+    payload: Optional[PayloadContainer] = Field(
+        sa_column=sa.Column(payload_column_type(PayloadContainer), nullable=True)
+    )
+    message_id: UUID = Field(sa_column=sa.Column(sa.ForeignKey("message.id"), nullable=False))
+    user_id: UUID = Field(sa_column=sa.Column(sa.ForeignKey("user.id"), nullable=False))
+    created_date: Optional[datetime] = Field(
```
olliestanley, line 23: Why is this `Optional` even though it is not `nullable`?

olliestanley (approved):
I don't actually have data backend approve powers but this looks good to me now. Thanks!

backend/alembic/versions/2023_06_05_1502-ffa8c33d129c_add_message_revisions.py
```diff
@@ -0,0 +1,54 @@
...
+    op.add_column("message", sa.Column("edited", sa.Boolean(), server_default=sa.text("false"), nullable=False))
+    op.drop_index("idx_search_vector", table_name="message")
+    op.create_index("ix_search_vector", "message", ["search_vector"], unique=False, postgresql_using="gin")
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("ix_search_vector", table_name="message", postgresql_using="gin")
+    op.create_index("idx_search_vector", "message", ["search_vector"], unique=False)
```
andreaskoepf, line 50: Is this re-creation of the message search index necessary?

andreaskoepf: Ollie wrote via discord: "This is because the index in our DB is called "idx_search_vector" whereas in our code it's called "ix_search_vector". A problem from my message search implementation because I wrote the Alembic script manually instead of autogenerating it. It might be better to rename the one in the code to match the existing DB name and remove the Alembic create/drop"

andreaskoepf (changes requested):

AbdBarho (approved):
thank you, frontend changes look good to me.

andreaskoepf (commented):
For other prompt-repository operations we generally used `CommitMode.FLUSH` together with a single outer commit "transaction-scope" in the web-request-handler. This allows a bit better composability, e.g. to use the individual functions in a larger transaction which could trigger a rollback after the changes. Since other operations run earlier with `CommitMode.FLUSH` more than expected might be written to the DB ... when the COMMIT block is left.

andreaskoepf (approved):
Backend lgmt!

PR #3295, opened 2023-06-05
Falcon SFT config

andreaskoepf:
Add configuration files used for SFT of 7B and 40B Falcon models.
(changes captured in this PR to free machine)


CloseChoice:
is something still missing here? Looks ok to me and is also a fix to the vicuna issue.

PR #3296, opened 2023-06-05
fix: batch or partition DB access and commits for user streak updation

melvinebenezer:
Partitioned the access to list of users  and updated the same
Tested on a small batch of Fake users

What is the best USER_STREAK_BATCH_SIZE ?

### DIFF
--- a/backend/oasst_backend/scheduled_tasks.py
+++ b/backend/oasst_backend/scheduled_tasks.py
@@ -7,6 +7,7 @@ from asgiref.sync import async_to_sync
 from celery import shared_task
 from loguru import logger
 from oasst_backend.celery_worker import app
+from oasst_backend.config import settings
 from oasst_backend.models import ApiClient, Message
 from oasst_backend.models.db_payload import MessagePayload
 from oasst_backend.prompt_repository import PromptRepository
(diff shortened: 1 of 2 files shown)


andreaskoepf:
I think the current implementation could be implemented with two update statements (which would be more efficient than the loops):
```
# reset streak_days to 0 for all users with more than one day inactivity
update "user" set streak_days=0, streak_last_day_date=current_timestamp::date where (current_timestamp-last_activity_date) > interval '1 day' or streak_days is null or last_activity_date is null;

# increment streak_days when user completed tasks on consecutive days 
update "user" set streak_days=streak_days+1, streak_last_day_date=current_timestamp where extract(day from current_timestamp) != extract(day from streak_last_day_date);
```

I think there was a logic error in the current code which causes the streak_days count to become too low after a some time. After incrementing it sets the streak_last_day_date = current_time .. but in this case it is clear that (current_time-streak_last_day_date) > 24h .. that means it takes >24h before each streak increment .. the effect depends on the update interval .. but let's say it is 4 hours than in the worst case it could mean a streak is incremented only all 28h ..

IMO one big problem with the current logic is: It always updates the streak_last_day_date of all users .. especially of all inactive users... maybe the whole streak-update mechanisms should be redesigned?

backend/oasst_backend/scheduled_tasks.py
```diff
@@ -94,44 +92,25 @@ def update_search_vectors(batch_size: int) -> None:
...
-            else:
-                logger.info("Not yet 24hours since the process started! ...")
-        logger.info("User streak end...")
+            logger.info("Process timedelta greater than 24h")
+
+            # Reset streak_days to 0 for users with more than one day of inactivity
+            reset_query = f"""
+                    UPDATE "user"
+                    SET streak_days = 0,
+                        streak_last_day_date = '{current_time}'
```
andreaskoepf, line 106: `streak_last_day_date = null`

melvinebenezer:
To summarise the changes

1.  Update the user streak when there is user activity
2. Periodic celery task to reset streak info in case of inactivity

PR #3308, opened 2023-06-06
Initial refactor of inference worker code

olliestanley:
Now follows the same pattern as other components of OA with an imported module separate from the entry script(s)


andreaskoepf:
@olliestanley Could you resolve the conflicts please? What does "installable" mean here (all imports currently seems to happen as relative imports to the worker entry-point)?

PR #3313, opened 2023-06-06
Inference backend support system prompt

olliestanley:


andreaskoepf (approved):
OK. Currently only SFT-8 models (which turned out to be worse than expected) where trained with system prompt.

olliestanley:
> OK. Currently only SFT-8 models (which turned out to be worse than expected) where trained with system prompt.

Yeah, this PR adds support in the backend but it will be disabled for now, just getting the capability in for later use

PR #3316, opened 2023-06-06
fix for plugins openapi parser

draganjovanovich:
This PR should enable openapi_parser to handle relative paths for servers url for open API endpoints.

### DIFF
--- a/inference/worker/openapi_parser.py
+++ b/inference/worker/openapi_parser.py
@@ -141,19 +141,20 @@ def get_plugin_endpoints(api_url: str, openapi_dict: dict) -> list[inference.Plu
 def prepare_plugin_for_llm(plugin_url: str) -> inference.PluginConfig | None:
     plugin_config = get_plugin_config(plugin_url)
 
-    if plugin_config:
-        try:
-            parsed_url = urlsplit(plugin_config.api.url)
-            if parsed_url.scheme == "":
-                api_url = urlsplit(plugin_url)._replace(path=parsed_url.path).geturl()
-            else:
-                api_url = plugin_config.api.url
+    if not plugin_config:
+        return None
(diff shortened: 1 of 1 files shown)


inference/worker/openapi_parser.py
```diff
@@ -135,19 +140,20 @@ def get_plugin_endpoints(api_url: str, openapi_dict: dict) -> list[inference.Plu
 
 def prepare_plugin_for_llm(plugin_url: str) -> inference.PluginConfig | None:
     plugin_config = get_plugin_config(plugin_url)
-    if not plugin_config:
```
olliestanley, line 138: Could we restore this? Feels a bit cleaner than an extra layer of nesting in the below code

draganjovanovich: Is like this fine?
`def prepare_plugin_for_llm(plugin_url: str) -> inference.PluginConfig | None:
    plugin_config = get_plugin_config(plugin_url)
    
    if not plugin_config:
        return None

    try:
        parsed_url = urlsplit(plugin_config.api.url)
        if parsed_url.scheme == "":
            api_url = urlsplit(plugin_url)._replace(path=parsed_url.path).geturl()
        else:
            api_url = plugin_config.api.url

        openapi_dict = fetch_openapi_spec(api_url)
        plugin_config.endpoints = get_plugin_endpoints(api_url, openapi_dict)
        return plugin_config

    except Exception:
        logger.debug(f"Plugin preparation error: {plugin_url}")
        return None`
        
btw, I just wanted to remove startswith("http") part

olliestanley: Yeah, exactly

olliestanley (approved):

Issue #3305, opened 2023-06-06
Chat list scroll bar too short for small number of chats.


CertifiedJoon:
Below is UI when theres only a few chats (pop up menu is difficult to use)

<img width="270" alt="image" src="https://github.com/LAION-AI/Open-Assistant/assets/61619422/5791e699-5b21-4a7b-a820-98e619279c85">

Below is UI when there are a lot of chats (pop up is menu easy to use)

<img width="272" alt="image" src="https://github.com/LAION-AI/Open-Assistant/assets/61619422/28606ee9-52cb-4f64-ba70-479b10d4551c">

I believe the best way to solve this must be providing the chat list some default height, preferably flexible.

CertifiedJoon:
I will work on this myself.

PR #3319, opened 2023-06-07
Modify: Default Height on Chat List

CertifiedJoon:
Fixes: #3305 

/chat
I thought that it was intuitive to set height to of chat list to zero when there is no chats on the list. 
But, as soon as a chat has been added to the list, I have set the default height.

/chat/{id}
I thought it is counter-intuitive to set default height to zero on this page. I have set default height to hundered percent.

ezgif com-video-to-gif


AbdBarho (approved):
Thanks

PR #3322, opened 2023-06-07
Edit history viewer

someone13574:
Followup PR to #3280 

Additons:
- Endpoint for fetching the revision history of a message
- Card which allows moderators to view the revision history of a message

*Note: Not sure if I'm sorting the uuid's correctly, but it seems to work.*

Screenshot:
Screenshot from 2023-06-07 16-33-15

### DIFF
--- a/backend/oasst_backend/models/message_revision.py
+++ b/backend/oasst_backend/models/message_revision.py
@@ -4,6 +4,7 @@ from uuid import UUID
 
 import sqlalchemy as sa
 import sqlalchemy.dialects.postgresql as pg
+from pydantic import PrivateAttr
 from sqlmodel import Field, SQLModel
 from uuid_extensions import uuid7
 
@@ -23,3 +24,5 @@ class MessageRevision(SQLModel, table=True):
     created_date: Optional[datetime] = Field(
         sa_column=sa.Column(sa.DateTime(timezone=True), nullable=True, server_default=sa.func.current_timestamp())
     )
+
+    _user_is_author: Optional[bool] = PrivateAttr(default=None)
--- a/backend/oasst_backend/prompt_repository.py
+++ b/backend/oasst_backend/prompt_repository.py
@@ -801,6 +801,7 @@ class PromptRepository:
         return query.all()
 
     def fetch_message_revision_history(self, message_id: UUID) -> list[MessageRevision]:
+        # the revisions are sorted by time using the uuid7 id
         revisions: list[MessageRevision] = sorted(
             self.db.query(MessageRevision).filter(MessageRevision.message_id == message_id).all(),
             key=lambda revision: revision.id.int >> 80,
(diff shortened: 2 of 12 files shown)


backend/oasst_backend/models/message_revision.py
```diff
@@ -23,3 +24,5 @@ class MessageRevision(SQLModel, table=True):
     created_date: Optional[datetime] = Field(
         sa_column=sa.Column(sa.DateTime(timezone=True), nullable=True, server_default=sa.func.current_timestamp())
     )
+
+    _user_is_author: Optional[bool] = PrivateAttr(default=None)
```
andreaskoepf, line 28: Does this now work without work-around? e.g. see https://github.com/tiangolo/sqlmodel/issues/149?

In the Message ORM class in message.py I had to insert:
```
    def __new__(cls, *args: Any, **kwargs: Any):
        new_object = super().__new__(cls, *args, **kwargs)
        # temporary fix until https://github.com/tiangolo/sqlmodel/issues/149 gets merged
        if not hasattr(new_object, "_user_emojis"):
            new_object._init_private_attributes()
        return new_object
    ```

File: backend/oasst_backend/prompt_repository.py [hunk 2]
```diff
@@ -800,6 +800,15 @@ def fetch_message_text_labels(self, message_id: UUID, user_id: Optional[UUID] =
             query = query.filter(TextLabels.user_id == user_id)
         return query.all()
 
+    def fetch_message_revision_history(self, message_id: UUID) -> list[MessageRevision]:
+        revisions: list[MessageRevision] = sorted(
+            self.db.query(MessageRevision).filter(MessageRevision.message_id == message_id).all(),
+            key=lambda revision: revision.id.int >> 80,
```
andreaskoepf, line 806: Could you maybe add a comment here that we sort by the date part of the uuid7? Might not be directly obvious to everyone...

backend/oasst_backend/models/message_revision.py
someone13574: Nope, not fixed. It only works because I set `_user_is_author` immediately after fetching the revisions, just like the `_user_is_author` field in the `message` table.

website/src/components/Messages/MessageHistoryTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+              user_id: message.user_id,
+              user_is_author: message.user_is_author,
+            },
+          ] as { text: string; created_date: string; user_id: string; user_is_author: boolean }[])
+        : (revisions.map((revision) => ({
+            text: revision.text,
+            created_date: revision.created_date,
+            user_id: revision.user_id,
+            user_is_author: revision.user_is_author,
+          })) as { text: string; created_date: string; user_id: string; user_is_author: boolean }[])
```
notmd, line 39: You can use `Omit<MessageRevision, ...>` here to avoid lengthy code

notmd (approved):
Thank you! Web change lgtm

andreaskoepf:
I would suggest to invert the order when rendering on the webpage .. more recent changes are probably more relevant and it would be great if they appeared first in the list, i.e. I suggest to render ordered by date descending...

andreaskoepf (approved):
Backend changes LGTM..

PR #1159, opened 2023-02-05
ci: switch flake8 by ruff

Borda:
Ruff is quite a new linter with impresive speed and offers to add multiple checks...

> Ruff aims to be orders of magnitude faster than alternative tools while integrating more functionality behind a single, common interface.
Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing tens or hundreds of times faster than any individual tool.

Also, I would suggest installing pre-commit bot, which helps to fix code within each PR so, lowering developer overhead... see: https://pre-commit.ci/

PR #3323, opened 2023-06-07
ci: switch flake8 by ruff

andreaskoepf:
This PR supersedes #1159 (cloned OA repo was archived by original PR author).
Original text by [Borda](https://github.com/Borda):

> Ruff aims to be orders of magnitude faster than alternative tools while integrating more functionality behind a single, common interface.
Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing tens or hundreds of times faster than any individual tool.

Also, I would suggest installing pre-commit bot, which helps to fix code within each PR so, lowering developer overhead... see: https://pre-commit.ci/


Borda:
> cloned OA repo was archived by original PR author

I would unarchive if you would say that we may continue and eventually land the original PR :)

Borda (approved):

andreaskoepf:
@Borda … your commits are in the new PR, I just merged main to ensure that pre-commit is passing with the latest source revision… which it is. Thanks again.

PR #3324, opened 2023-06-07
Simplify message and token format doc

olliestanley:


model/MESSAGE_AND_TOKEN_FORMAT.md
```diff
@@ -65,82 +56,41 @@ The most up-to-date place to look for code is here:
...
-is empty. **Note:** I've added newlines and comment just for readability here.
-They aren't in the format.
+is empty.
+
+**Note:** I've added newlines and comment just for readability here. They aren't
+in the format.
+
+**Note**: That `<|system|>` refers to a specific token that the model knows
+about. If the user literally typed `<|system|>` then that would be tokenized to
+completely different non-special tokens, reducing possible attacks.
```
andreaskoepf, line 93: hm. Escaping of tokens in actual messages would be a good feature, currently I think this is not the case and  if `<|system|>` would appear in the text it would be translated to the special-system token by the tokenizer.

olliestanley: Ok, this was in the original doc, guess it was optimistic :D Will remove this

andreaskoepf (approved):
OK, this is definitely a step in the right direction! In the future we probably could also move the v2 format description to the top of the document. But step by step & and thanks a lot for reducing the confusion caused by the file!

PR #3327, opened 2023-06-08
Add return value in prepare_tensor

Choiuijin1125:
There is no return value in `prepare_tensor`

https://github.com/LAION-AI/Open-Assistant/blob/0fcf3e08fe62295d4696e590005b0f33383342ea/model/model_training/utils/utils_rl.py#L5-L7

so it will cause None type error while RLHF training

https://github.com/LAION-AI/Open-Assistant/blob/0fcf3e08fe62295d4696e590005b0f33383342ea/model/model_training/utils/ppo_utils.py#L553-L561

https://github.com/LAION-AI/Open-Assistant/blob/0fcf3e08fe62295d4696e590005b0f33383342ea/model/model_training/trainer_rl.py#L93-L99


andreaskoepf (approved):
good catch. thnx.

PR #3330, opened 2023-06-08
Initial inference documentation pass

olliestanley:
In the recent meeting it was raised that the codebase is not documented and hard to understand as a result. This adds some initial documentation. For now it is quite basic but hopefully will allow a bit more understanding of some parts of the code "at a glance", reducing the friction somewhat.


andreaskoepf (approved):
Super nice docstrings, thank you very much!!!

PR #3375, opened 2023-06-10
Fix Dialogs in Chat

### LINKED ISSUE #3331: Both options pertaining to chat management are very nonfunctional
Not only are these functions, opt-out of training data and delete the chat completely, hidden behind other elements (not clickable at all) but they don't actually perform the functionality they depict if you delete the elements in front so that you can get to them - to click them.

image

This is true for both opting out of the chat being used in the training data set and deleting the chat completely (which should also remove it from the data set). Neither do anything at all other than refresh the page.

image

There is another issue [#3267](https://github.com/LAION-AI/Open-Assistant/issues/3267) discussing the fact that the delete button is behind other elements and unclickable, but this seems to be a more problematic issue than just what is described there. As well in issue [#3306](https://github.com/LAION-AI/Open-Assistant/issues/3306) where we see that this is not only relevant to the chat page itself, but also from the chat index page we see the same issues -- that is minus the buttons being unclickable. I would love to see updates on what is in mind to fix this or remove the functionality because it is not feasible, etc.

image

Small sub-note here: The text "Confirm opting of training data" is not clear as to what logical direction "opting" refers to. This could be opting in or opting out and since there is no notification anywhere showing 'this chat is currently opted into the data set' or 'this chat is opted out of the data set' for instance, we have no idea as to what state this option is in as the end user.

AbdBarho:
Fixes #3331

image

### DIFF
--- a/website/public/locales/en/chat.json
+++ b/website/public/locales/en/chat.json
@@ -20,7 +20,8 @@
     "button": "Opt out of training data",
     "success_message": "You have opted out of training data.",
     "dialog": {
-      "title": "Confirm opting of training data"
+      "title": "Are you sure you want to opt this chat out of the data?",
+      "description": "If you confirm, this chat won't be used for improving our model. Please take the time to upvote and downvote responses in other chats to help us make Open Assistant better!"
     }
   },
   "parameter_description": {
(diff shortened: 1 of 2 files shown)


notmd (approved):
Thanks

PR #3382, opened 2023-06-11
Fix for updating search vectors on edit

### LINKED ISSUE #3377: Moderator message search on edited messages shows incorrect results
When a message is edited, message search will still return that message if the original content matches the search phrase, even if the new content doesn't. This also results in the moderator message table showing the original content in the search view and not showing the 'edited' tag.

You can see this by going the the web-dev server and using message search for 'hello'. The top result will show the original message content and is missing the 'edited' tag but when you click on it, it shows the correct history. The admin message table also shows the correct data when not searching.

I'm not entirely sure on how the message search system works, but this appears to be due to something not being updated when it needs to be when a message is edited.

When not searching (correct):
Screenshot from 2023-06-10 17-56-39

When searching for 'hello' (incorrect, outdated info and shouldn't show up):
Screenshot from 2023-06-10 17-53-05

Message history (correct):
Screenshot from 2023-06-10 17-53-17

olliestanley:
Close #3377


backend/oasst_backend/scheduled_tasks.py
```diff
@@ -72,17 +72,50 @@ def update_search_vectors(batch_size: int) -> None:
...
+                        Message.search_vector.is_(None),
+                        MessageRevision.created_date > Message.search_vector_update_date,
+                        and_(
+                            Message.search_vector_update_date.is_(None),
+                            MessageRevision.created_date.isnot(None),
+                        ),
+                    )
+                )
+
+                to_update: list[Message] = query.limit(batch_size).all()
```
melvinebenezer, line 109: i think we need an offset that is incremented by the batch_size , no ?
to_update = query.offset(offset).limit(batch_size).all()
offset += batch_size

melvinebenezer (changes requested):

backend/oasst_backend/scheduled_tasks.py
olliestanley: No, I don't think we should need one as we `session.commit()` after each batch. So when the next batch begins, the messages updated in the previous batch will no longer meet the update criteria so will not be selected by the query.

melvinebenezer (approved):

Issue #3378, opened 2023-06-11
Delete Chat Throwing Runtime Error & Text Running out of container


CertifiedJoon:
Error on delete
<img width="738" alt="image" src="https://github.com/LAION-AI/Open-Assistant/assets/61619422/e9a1892a-da3f-40c6-b296-b28dd6448a58">

Confirm delete popup text would run out of the box
<img width="620" alt="image" src="https://github.com/LAION-AI/Open-Assistant/assets/61619422/da042d48-5d73-414f-9828-855c66d8c205">

Delete process is quite problematic, there is also an issue assigned to me on #3306 

I would like to work on this issue and #3306 all together. Would that be fine?

CertifiedJoon:
Plus : The delete wouldn't trigger a toast popup, meanwhile, optOut does. So I am thinking, for consistency sake, we should add toast for successful delete.

AbdBarho:
I think this is a duplicate of https://github.com/LAION-AI/Open-Assistant/issues/3331?

CertifiedJoon:
@AbdBarho indeed. Should i close the relevant pr?

Issue #3306, opened 2023-06-06
Deleting from chat index page redirects to the chat page.


CertifiedJoon:
Deleting a chat from /chat page will redirect /chat/{id}

AmanSal1:
can i work on this ?

andreaskoepf:
@AmanSal1 Ollie assigned the task to you three days ago. Please confirm that you started working on it.

CertifiedJoon:
I will work on this myself. Would that be okay?

andreaskoepf:
> I will work on this myself. Would that be okay?

Yes sure, thanks a lot.

Issue #3267, opened 2023-05-31
chat deletion confirmation gets covered by chat response


wuhei:
![image](https://github.com/LAION-AI/Open-Assistant/assets/30749950/758948ab-0527-43e5-be62-0d942c52fdca)

JaynouOliver:
Can I work on it? Please assign it to me.

olliestanley:
> Can I work on it? Please assign it to me.

Feel free to take a look, I have assigned you

davidak:
I can reproduce this issue. The dialog appears behind the chat. Also, the chat UI does not get darker.

![Screenshot from 2023-06-06 10-30-58](https://github.com/LAION-AI/Open-Assistant/assets/91113/e04d23e7-3d57-443e-9c5a-8a7242e0d702)

andreaskoepf:
@JaynouOliver Were you able to set up a dev environment and reproduce the problem?

CertifiedJoon:
@andreaskoepf If @JaynouOliver is struggling to fix it, I think I can take over. I have been working on delete feature on #3378 and #3306.

andreaskoepf:
> I think I can take over.

Since we haven't heard back from @JaynouOliver I transfer the issue over to you. Thank you very much for helping to fix this!

PR #3383, opened 2023-06-11
Bug Fix: Delete feature (Fixes three issues)

CertifiedJoon:
Fixes: #3267 
Fixes: #3306 
Fixes: #3378 

Regarding: #3267 
Some of the z-indices were manged by zIndex parameter, while other components where managed by Portals. Somehow the two conflicted and created a weird mix of zIndices. I have changed popover's portal into simple zindex parameter and parents, sibligns, and cousins accordingly.

Regarding: #3306 
The redirection to deleted chat happened because one of its parent elements were a Link component. Previously, upon pressing delete, stopEvent() was called to temporarily pause redirection, and redirection would resume after the delete process. I have changed the Link component into normal button with onClick redirection, maintaining the stopPropogation behavior.

Regarding: #3378 
The default behavior of AlertDialog would wrap the text. But somehow using portal on it would restrict text-wrap behavior. The text-wrap behavior came back after removal of portal while working on #3267 .

deleting from /chat/{id}
image

deleting from /chat
image

toast on delete
image

### DIFF
--- a/website/src/components/Chat/ChatListBase.tsx
+++ b/website/src/components/Chat/ChatListBase.tsx
@@ -14,9 +14,9 @@ import { ChatListViewSelection, useListChatPagination } from "./useListChatPagin
 
 export const ChatListBase = memo(function ChatListBase({
   allowViews,
-  minHeight,
+  noScrollbar,
   ...props
-}: CardProps & { allowViews?: boolean; minHeight?: string }) {
+}: CardProps & { allowViews?: boolean; noScrollbar?: boolean }) {
   const [view, setView] = useState<ChatListViewSelection>("visible");
   const { loadMoreRef, responses, mutateChatResponses } = useListChatPagination(view);
   const chats = responses?.flatMap((response) => response.chats) || [];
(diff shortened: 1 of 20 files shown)


website/src/components/Chat/ChatListBase.tsx
```diff
@@ -79,6 +79,7 @@ export const ChatListBase = memo(function ChatListBase({
         bg: "blackAlpha.400",
       }}
       {...props}
+      zIndex="var(--chakra-zIndices-base)"
```
notmd, line 82: why do we need zIndex on this? Isn't it is the last layer by default?

website/src/components/Chat/ChatListItem.tsx
```diff
@@ -78,7 +78,7 @@ export const ChatListItem = ({
     <Button
       // @ts-expect-error error due to dynamically changing as prop
       ref={rootRef}
-      {...(!isEditing ? { as: Link, href: ROUTES.CHAT(chat.id) } : { as: "div" })}
+      // {...(!isEditing ? { as: Link, href: ROUTES.CHAT(chat.id) } : { as: "div" })}
```
notmd, line 81: Please revert this. Using a tag is much more better than JS

notmd:
Thanks, but I believe you are relying on zIndex too much which will be hard to debug in future. I will take a look and get rid of it.

website/src/components/Chat/ChatListBase.tsx
CertifiedJoon: You are right, I've redacted the said line.

website/src/components/Chat/ChatListItem.tsx
CertifiedJoon: Done.

CertifiedJoon:
> Thanks, but I believe you are relying on zIndex too much which will be hard to debug in future. I will take a look and get rid of it.

Thank you :)

notmd:
I just push all the fix, please check @CertifiedJoon

CertifiedJoon:
@notmd Thank you for your help! Looks great. I will make sure to use portals more often from now on.

Issue #3073, opened 2023-05-07
Ability to change the e-mail like we can change the Nickname


Subarasheese:
Greetings.

It would be useful to have the ability to change our e-mail credentials like most other platforms online do.

The interface would work the same way as we have currently for changing usernames, except it should send a confirmation mail first. If the user opens the token link on the e-mail, then the e-mail for that user will be changed on the database.

PR #3384, opened 2023-06-11
Add "update_email" ability

KosmasKanellakis:
I have worked on the code for the issue #3073.

The email is updated if it is not already in the database, but no token email is sent for confirmation.


AbdBarho (changes requested):
Thank you

> but no token email is sent for confirmation.

This is unfortunately a big security concern and would need to be solved first.

PR #3387, opened 2023-06-11
add Lima dataset

jordiclive:
Add Lima dataset from https://huggingface.co/datasets/64bits/lima_vicuna_format, simply copying vicuna template.


andreaskoepf (approved):
Thanks!

PR #3392, opened 2023-06-12
chat.json

Ain69:
// виправлено деякі помилки, зроблена правильна сортировка за англійською мовою на українську мову // some errors were fixed, the correct sorting from English to Ukrainian was made


Ain69:
Good afternoon. I have improved the Ukrainian language for OA, could you please run a test and accept these changes. I also have some more updated files for "uk-UA"

PR #3416, opened 2023-06-13
leaderboard.json

Ain69:
fixed some bugs, updated through the English version, and sorted correctly from English to Ukrainian


notmd (approved):
Thanks

PR #3425, opened 2023-06-13
Disable data collection features on opt-out

someone13574:
- Prevents draft messages from being generated when a user opts-out from data for a chat
- Hides the thumbs up and down buttons
- Disables encourage message


notmd (approved):
Thanks

PR #3426, opened 2023-06-13
Add missing ai-plugin.json for built-in plugins URLs

andreaskoepf:


olliestanley (approved):

andreaskoepf:
@olliestanley thanks for approval, the builtin plug-ins are currently blocked by the inference workers which need to be updated first.

PR #3420, opened 2023-06-13
account.json

Ain69:
The account.json file has been created. Fixed some bugs, made the correct sorting from English to Ukrainian

PR #3427, opened 2023-06-14
Revert and update some of strings in tasks.json for the Ukrainian translation

nmeln:
The changes in tasks.json in https://github.com/LAION-AI/Open-Assistant/pull/3420 removed the translation of the `reply_as_assistant` block. I also cannot understand some of the new translations :) 

@olliestanley FIY: I didn't remove `moderator_edit_explain` as that seems new. But I may have missed something if there were more strings added in the blocks. Will make another PR later to sort the keys if the goal was to better follow the convention of the `en` translation files


andreaskoepf (approved):
thanks!

Issue #3246, opened 2023-05-28
User accounts deduping


AbdBarho:
## Context

There are some users who are facing problem when logged in through multiple methods, i.e. email first and then google / discord second, or the other way around. 
https://github.com/LAION-AI/Open-Assistant/issues/3136 https://github.com/LAION-AI/Open-Assistant/issues/3232 and probably many more. We don't want their efforts to go to waste.

In the frontend db we have `users` and `accounts`, in theory, it should be possible for a user to have multiple accounts from different providers and still be the same user.

However, the way we identify users in the data backend is a combination of user id and auth method, which makes it harder to reconnect the two. 

For chat, we only use the user's frontend db id, which should make it easier.

## Solution Idea

Ideally, we should be able to identify a user through our entire stack using a single piece of information, which is identical through all auth methods we provide. I see the email address being the only valid option, which is used by `next-auth` for connecting multiple accounts to the same user.

Anything derived from the email address could also work.

This would require at least:
* Update user auth frontend <--> data backend
* Change auth headers sent
* Merge potential duplicate users using their email addresses while preserving their contributions & privileges
* Update tables
* Update user auth frontend <--> inference

This will probably be a gradual process where we would have to support both old and new formats while we do the migration.

## Open questions

* What should this unique identifier be
  * email?
  * id of user in frontend db (since it is the entrypoint for all user interactions)
  * Or maybe a separate "user" service?
* Are there some solutions out there that already solve this problem?

---

@andreaskoepf @yk @notmd @olliestanley would love to hear your opinions on this, since it is a huge topic and basically touches everything.

notmd:
Have you tested what is the next auth behavior if we remove the unique index from the email column?

AbdBarho:
> Have you tested what is the next auth behavior if we remove the unique index from the email column?

I can test it locally, but that would not help much. That being said, I think leaving it unique is the right choice, we don't want multiple users with the same email.

notmd:
When the user login Discord. If they change their discord email and login again, next-auth still matches to old account and doesn't update the email.

AbdBarho:
Discussion from meeting 09.06.2023:

The main point should be to use the same ID across the entire stack.
We should stick to the ID from the data backend if possible.

Merging multiple accounts in the backend should be possible, if we know the corresponding frontend ids.

I will try to analyze the databases and see how much overlap do we have and how easy would it be to migrate.

notmd:
[...]

PR #3475, opened 2023-06-14
Add POST /api/v1/admin/merge_users

andreaskoepf:
First backend support for #3246 : Adds a new `/api/v1/admin/merge_users` endpoint to merge one or more source user accounts into an existing destination user account. The source user accounts are deleted in the process and all objects that belonged to the source users are transferred to the destination user. Rows belonging to the source accounts in the `user_stats` and `troll_stats` tables are also deleted.


melvinebenezer (approved):
LGTM

Issue #2778, opened 2023-04-20
Please add a "Are you Sure you want to Skip" popup to avoid frustration and loss of progress


Logophoman:
Unfortunately I have already written many long answers and while correcting them I have clicked on the `skip` button which lead to my messages being deleted immediately which is extremely frustrating, especially when you curated that particular answer for half an hour or so already... 

If the Assistant or User Message is empty skipping immediately should be no problem, but if there is a message in the response field there should be a popup!

notmd:
That was old behavior. But I think we can add it back and have a "dont show it again in this session" option and store it in the session storage

bterrific2008:
[I think this is the commit that removed the old behavior,](https://github.com/LAION-AI/Open-Assistant/commit/afc06702bc82e19ba81a8647ea6035adfda57d20) but please correct me if I'm wrong.

A possible solution for this would be to bring back the deleted [SkipButton](https://github.com/LAION-AI/Open-Assistant/blob/d3965750bec06b6954db7afaf105eebab37eb573/website/src/components/Buttons/Skip.tsx#L43-L54), remove the [requesting feedback TextArea](https://github.com/LAION-AI/Open-Assistant/blob/d3965750bec06b6954db7afaf105eebab37eb573/website/src/components/Buttons/Skip.tsx#L43-L54) and just replace it with a Text element to ask a user to confirm if they want to skip or not. I'm less certain on the correct way to interact with the session storage. I see in the existing website code there's reference to [`useSession`](https://next-auth.js.org/getting-started/client#usesession) from next-auth, but I'm not sure if its best practice to use that to update the session storage.

I'm unlikely to work on this issue, but hope this comment helps anyone else interested in working on this!

MehrJonas:
I would like to work on this. I had a look at the code, I think it should be enough to just look at the replyValidity, no need to do anything fancy with the session storage

andreaskoepf:
@MehrJonas great, I assigned this issue to you.

AbdBarho:
#3479

PR #3479, opened 2023-06-15
Added confirmation dialog when skipping after entering text

MehrJonas:
Adds a dialog asking the user to confirm whether they want to skip a question after they have already entered text to prevent accidentally losing an answer by clicking the wrong button as requested in https://github.com/LAION-AI/Open-Assistant/issues/2778


website/public/locales/en/common.json
```diff
@@ -48,6 +48,7 @@
   "sign_in": "Sign In",
   "sign_out": "Sign Out",
   "skip": "Skip",
+  "skip_confirmation": "Are you sure you want to skip?",
```
notmd, line 51: Can you move this to task.json file?

website/src/components/Buttons/Skip.tsx
```diff
@@ -0,0 +1,62 @@
...
+  confirmSkip: boolean;
+}
+
+export const SkipButton = ({ ...props }: SkipButtonProps) => {
+  const { t } = useTranslation();
+  const { isOpen, onOpen: showModal, onClose: closeModal } = useDisclosure();
+
+  const onClick = () => {
+    if (props.confirmSkip) {
+      showModal();
```
notmd, line 26: Also can you only show the modal when the input is not empty?

MehrJonas: I'm already doing that... that's what the confirmSkip flag is for. I'm setting that to true if taskstatus.replyValidity === "VALID" (see TaskControls.tsx), which corresponds to a non-empty reply field.

notmd: ah thanks

notmd (approved):

PR #3480, opened 2023-06-15
Fix out of bound index in chat paging

### LINKED ISSUE #3478: Chat error: "Cannot read properties of undefined (reading 'children')"
When I use chat (web version), sometimes I get "Application error: a client-side exception has occurred (see the browser console for more information).". This sometimes happens when assistant generate response or when I press left/right buttons (the ones under my request, which has been edited) to "switch to different branch" (I don't know how to properly call it).

Browser console output:
```
TypeError: Cannot read properties of undefined (reading 'children')
    at tb ([id]-52c36ab214538aed.js:1:25941)
    at ab (framework-ce84985cd166733a.js:9:60917)
    at ud (framework-ce84985cd166733a.js:9:72803)
    at i (framework-ce84985cd166733a.js:9:120372)
    at oO (framework-ce84985cd166733a.js:9:99114)
    at framework-ce84985cd166733a.js:9:98981
    at oF (framework-ce84985cd166733a.js:9:98988)
    at ox (framework-ce84985cd166733a.js:9:95740)
    at oC (framework-ce84985cd166733a.js:9:96129)
    at r4 (framework-ce84985cd166733a.js:9:44778)
J
```

someone13574:
closes #3478 

`useIsomorphicLayoutEffect` wasn't being called until the first render meaning if the index could be out of bounds if the parent message was changed to a tree with less siblings than the previous. A simple min against the number of siblings prevents the out of bound index and then `useIsomorphicLayoutEffect ` is called the next frame, setting the correct thread.

### DIFF
--- a/website/src/components/Chat/ChatConversationTree.tsx
+++ b/website/src/components/Chat/ChatConversationTree.tsx
@@ -56,7 +56,7 @@ const TreeChildren = ({
 } & StrictOmit<ChatMessageEntryProps, "message" | "canRetry">) => {
   const sortedTrees = useMemo(() => trees.sort((a, b) => Date.parse(a.created_at) - Date.parse(b.created_at)), [trees]);
   const [index, setIndex] = useState(0);
-  const currentTree = sortedTrees[index];
+  const currentTree = sortedTrees[Math.min(index, trees.length - 1)];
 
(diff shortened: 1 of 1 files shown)


notmd (approved):
Thanks

PR #3488, opened 2023-06-17
Add frontend code to merge users

AbdBarho:


website/src/lib/users.ts
```diff
@@ -111,3 +113,75 @@ export const getBatchFrontendUserIdFromBackendUser = async (users: { username: s
...
+  let backendId: string;
+
+  if (backendIds.length === 1) {
+    backendId = backendIds[0];
+    logger.info({ message: `Wanted to merge user accounts, but found only one, skipping.`, frontendId, backendId });
+  } else {
+    logger.info({ message: "Merging user accounts", frontendId, accounts, backendIds });
+    let remainingIds: string[];
+
+    [backendId, ...remainingIds] = backendIds;
```
AbdBarho, line 166: here I am just taking the first, maybe there is a better idea.

notmd:
Thanks but I am really against this decision. The backend should depend on the frontend, just like the inference is doing now.

andreaskoepf:
> The backend should depend on the frontend, just like the inference is doing now.

Sorry, but I don’t understand what you mean by „the backend should depend on the frontend“. Could you please explain how you propose to solve the duplicate accounts problem?
BTW you seem to imply thatcher is good that the inference Server now depends on the frontend but IMO this is a big architectural mistake which should be fixed, e.g. to allow discord bots etc to also use the inference api..

AbdBarho:
I do see the value of using the frontend id since it is for the time being the central service where all accounts are stored (without duplicates)

Even if we want to add discord bot support to the inference, we would still have the duplication problem, but now in the inference not in the data backend.

As long as any given user could have multiple accounts, we need a way to identify them across all services in our stack, which means we need a central service for auth.

In our current state, this service is the frontend, but could also be the data backend, which does not make any difference from an architectural point of view.

I could see us separating the auth part from the frontend to its own service, which is used by any component we want.


I am very interested in your ideas.

notmd:
I agree with @AbdBarho to make the frontend as an auth service since it all the entry point of our app and we also can reuse the next auth logic. 

> Sorry, but I don’t understand what you mean by „the backend should depend on the frontend“. Could you please explain how you propose to solve the duplicate accounts problem?

Here is my idea for the migration
- Add a `frontend_id` column (unique and nullable) to the backend user table. The frontend will send a frontend_id header and backend will use that to identify user. The backend still can use the `user.id` for it internal usage . We still can use `auth_method:id` as a fallback while we migrate data. This approach will prevent creating duplicate account. 
- Write a script to dedup accounts and run it.
- After migrate all data, make the `frontend_id` column is required. Enfore the frontend always send the `frontend_id`, remove the `auth_method:id` header, and remove the `auth_method`, `username`, and `display_name` from `user` table.

These steps can be simplified if we use the `frontend_id` as the `id` column in user table.

> BTW you seem to imply that it is good that the inference server now depends on the frontend but IMO this is a big architectural mistake which should be fixed, e.g. to allow discord bots etc to also use the inference api..

The discord bot needs to communicate with the frontend for authentication. There is no way you can avoid this behavior in microservice.

PR #3489, opened 2023-06-17
Add linked accounts to account page

AbdBarho:
Does not look the best, but we should add buttons for users to unlink their accounts (or link more)


notmd (approved):
Look perfect lol

AbdBarho:
> Look perfect lol

Well I guess we can just merge it and add the other stuff later

PR #3491, opened 2023-06-18
Update profile image on login

### LINKED ISSUE #2196: Debug user discord image not always available.
Users logged in with discord don't always have valid images.

My initial hypothesis:

We implemented a refresh of the user's profile images https://github.com/LAION-AI/Open-Assistant/pull/2043, which worked, but fails sometimes.

It seems that this is related to how nextauth handles credentials, or more precisely, how it does not correctly handles refresh tokens.

https://github.com/nextauthjs/next-auth/issues/3599

https://github.com/nextauthjs/next-auth/issues/3599#issuecomment-1091608216

TLDR: we might need to handle the refresh of the user token ourself.

AbdBarho:
Fixes #2196 (not only for discord, but all providers)

Instead of doing the big mess of having to refresh images in the background, we update the images when the user signs in, from the profile we get from the provider.

This has the side effect that if you have 2 different images on 2 different login providers, your image will always change depending on the last provider you logged in with, but I think this is ok.

Also, consolidate all of the updates that are applied to the user to one database call.


website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -127,60 +126,44 @@ const authOptions: AuthOptions = {
       session.user.role = token.role;
       session.user.isNew = token.isNew;
       session.user.name = token.name;
+      session.user.image = token.picture;
```
Comment (notmd) [inline website/src/pages/api/auth/[...nextauth].ts:129]: You can receive image from user property in the callback here

Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts:129]: you mean as a parameter to the session callback?

the `user` object is only passed if we are using the database strategy, but we are using jwt.

Or did I  misunderstand what you say?

Comment (notmd) [reply website/src/pages/api/auth/[...nextauth].ts:129]: > the `user` object is only passed if we are using the database strategy, but we are using jwt.

Hmm I don't see where the docs mention this, the TS signature doesn't mark it as nullable as well.

notmd (approved):

website/src/pages/api/auth/[...nextauth].ts
Comment (AbdBarho) [reply website/src/pages/api/auth/[...nextauth].ts:129]: image

if I console log it is undefined, would be a lot easier if we did not have to pass everything though the token.

PR #3549, opened 2023-07-05
inlang: update Dutch account and chat translations

rikmwouters:
Added Dutch translations for account and chat related messages.


notmd (approved):
Thanks

PR #3564, opened 2023-07-11
Add RoPE Interpolation

shahules786:
Added support for RopE interpolation via the SuperHOT method and its variants proposed in
[reddit](https://www.reddit.com/r/LocalLLaMA/comments/14lz7j5/ntkaware_scaled_rope_allows_llama_models_to_have/)
[scaled-rope
](https://github.com/jquesnelle/scaled-rope/tree/master/scaled_rope)

Supported methods
- Linear scaling
- NTK aware scaling
- Dynamic NTK

Supported Models
- LLAMA
- Falcon 

This can be easily extended and experimented with by configuring two parameters
`superhot` and `superhot_config`


model/model_training/configs/config.yaml
```diff
@@ -779,3 +782,32 @@ debug:
   verbose: true
   num_train_epochs: 0.2
   dtype: fp32
+
+patching-test:
```
andreaskoepf, line 786: Could we change the name to something like "rope_scaling_test"?

model/model_training/models/rope.py
```diff
@@ -0,0 +1,175 @@
+import torch
+
+
+# rotary pos emb helpers (torch.jit.script does not seem to support staticmethod...)
+def rotate_half(x):
+    x1, x2 = x[..., : x.shape[-1] // 2], x[..., x.shape[-1] // 2 :]
+    return torch.cat((-x2, x1), dim=x1.ndim - 1)  # dim=-1 triggers a bug in torch < 1.8.0
+
+
+class RWNTKScaledRope(torch.nn.Module):
```
andreaskoepf, line 10: Could you add some comments about the origin of these rope scaling variants? Were they copied from the [scaled-rope](https://github.com/jquesnelle/scaled-rope) repository? If they were imported from a different place we should add a proper attribution.

model/model_training/trainer_sft.py
```diff
@@ -331,6 +331,7 @@ def main():
     if not training_conf.deepspeed or training_conf.local_rank == 0:
         tokenizer_sanity_check(tokenizer)
 
+    print("POINT 1")
```
andreaskoepf, line 334: remove?

model/model_training/trainer_sft.py
```diff
@@ -416,9 +416,14 @@ def main():
         sampler = None
 
     metrics, preprocess_fns = get_metrics(training_conf, tokenizer)
-
     model = get_model(training_conf, tokenizer)
 
+    from model_training.models.patching import RopePatch
```
andreaskoepf, line 421: Is there a benefit of late import, otherwise I would recommend to move this to the other imports at the beginning of the file (to make it easier to see all main dependencies).

model/model_training/configs/config.yaml
```diff
@@ -97,6 +97,9 @@ webgpt_dataset_only:
   datasets:
     - webgpt
 
+instruction_datasets:
+  datasets:
+    - dolly15k
 per_digit_tokens:
```
andreaskoepf, line 103: A newline should be added here to separate the configuration sections.

andreaskoepf (changes requested):
Thanks a lot for adding the scaled-rope implementation! I left some minor comments.

andreaskoepf (approved):

Issue #3517, opened 2023-06-25
Implement Celery task for Discord notifications of reports


olliestanley:
The previous implementation caused a bug and so is reverted by #3516

nablabits:
Hi @olliestanley , I'm happy to give this a try. I've managed to get the dev server running locally and spotted where that button lives and what are the methods called. I have a couple of questions:
- Did you revert the previous implementation as a quick fix or rather because it was a dead end? I say so because if it was a quick fix I can reinstate it and debug it.
- If it was a dead end, do you have in the code base some similar task I can take a look at as a guide?
- Finally, do you have a discord channel sandbox that I can use to test the messages arriving? Alternatively I can set up something on my own

Chances are this issue might have a steep learning curve, so if anyone more knowledgeable is keen to pick this up, feel free, I'm happy with the learning experience anyways.

olliestanley:
> Hi @olliestanley , I'm happy to give this a try. I've managed to get the dev server running locally and spotted where that button lives and what are the methods called. I have a couple of questions:
> 
> * Did you revert the previous implementation as a quick fix or rather because it was a dead end? I say so because if it was a quick fix I can reinstate it and debug it.
> * If it was a dead end, do you have in the code base some similar task I can take a look at as a guide?
> * Finally, do you have a discord channel sandbox that I can use to test the messages arriving? Alternatively I can set up something on my own
> 
> Chances are this issue might have a steep learning curve, so if anyone more knowledgeable is keen to pick this up, feel free, I'm happy with the learning experience anyways.

I think the previous approach was fine in principle, but there was a bug causing the report function to be broken entirely, and we wanted to fix prod ASAP so just reverted the change.

We don't have a sandbox, you might have to set something up for that unfortunately.

Hopefully this should be a fairly small change so would be great if you could take a look!

nablabits:
Cool, thanks for the answers, I will work through this these days and will let you know if I get stuck :crossed_fingers:

PR #3566, opened 2023-07-13
3517 - Send reports to discord via celery task

nablabits:
It should fix: https://github.com/LAION-AI/Open-Assistant/issues/3517

In debeb353cc0104a0a330637d078b9b0bc41afe84 we reverted the original implementation of the task as it broke the production environment. The problem was that celery tasks need objects that could be JSON serializable and we were passing a `Message` instance, so just passing the details fixed the issue.

Screenshot from 2023-07-13 08-04-49

A few things that I noticed:
- There's no `pre-commit` version pinned in the requirements.txt so I just used the latest one (3.3.3). In the past I had experienced issues with this so it might be a good idea to pin it.
- There are a few automated tests, I was about to write them but found no installation of `pytest` or similar, if you consider so, I'm happy to write them and install the necessary libraries as eventually that will make the codebase more robust.

### DIFF
--- a/backend/oasst_backend/prompt_repository.py
+++ b/backend/oasst_backend/prompt_repository.py
@@ -30,8 +30,8 @@ from oasst_backend.models import (
 from oasst_backend.models.payload_column_type import PayloadContainer
 from oasst_backend.task_repository import TaskRepository, validate_frontend_message_id
 from oasst_backend.user_repository import UserRepository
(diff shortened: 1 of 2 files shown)


melvinebenezer (approved):
Thanks, LGTM

melvinebenezer:
@nablabits : we can discuss on the automated tests in discord

PR #3571, opened 2023-07-15
Configure PEFT from config

shahules786:
## What
Added support to configure PEFT from config and save `WTE` embeddings with adapter files to enable easy loading of OA Lora weights.

## Why
Earlier PEFT modules were hardcoded for llama model only. This was an issue when  training other models using peft like RWModel, GPTNeoX, etc

## How
Introduces extra parameter `peft_config` to config.yml


andreaskoepf (approved):
very nice, thanks!

PR #3575, opened 2023-07-16
Add debug setup for inference server & worker

0xfacade:
I couldn't find an existing way to attach a debugger to the inference server or worker. In this PR, I enable both to be debugged using [debugpy](https://github.com/microsoft/debugpy), the Python debugger that integrates easily with VS Code.

Open issues:
- [ ] add documentation

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -106,6 +106,38 @@
         "CUDA_VISIBLE_DEVICES": "1,2,3,4,5",
         "OMP_NUM_THREADS": "1"
       }
-    }
+    },
+    {
+      "name": "Debug: Inference Server",
+      "type": "python",
+      "request": "attach",
+      "connect": {
+        "host": "localhost",
+        "port": 5678
+      },
+      "pathMappings": [
+        {
+          "localRoot": "${workspaceFolder}/inference/server",
+          "remoteRoot": "/opt/inference/server"
+        }
+      ],
+      "justMyCode": false
+    },
+    {
+      "name": "Debug: Worker",
+      "type": "python",
+      "request": "attach",
+      "connect": {
+        "host": "localhost",
+        "port": 5679
+      },
+      "pathMappings": [
+        {
+          "localRoot": "${workspaceFolder}/inference/worker",
+          "remoteRoot": "/opt/inference/worker"
+        }
+      ],
+      "justMyCode": false
+    },
   ]
 }
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -237,6 +237,7 @@ services:
     restart: unless-stopped
     ports:
       - "8000:8000"
+      - "5678:5678"  # Port to attach debugger
     depends_on:
       inference-redis:
         condition: service_healthy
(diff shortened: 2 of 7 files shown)


docker-compose.yaml
```diff
@@ -231,12 +231,14 @@ services:
       TRUSTED_CLIENT_KEYS: "6969"
       ALLOW_DEBUG_AUTH: "True"
       API_ROOT: "http://localhost:8000"
+      DEBUG: "True"
```
0xfacade, line 234: If I understand correctly, the compose file is only meant for local development, so setting this here shouldn't be a problem?

.vscode/launch.json
```diff
@@ -106,6 +106,38 @@
...
+      ],
+      "justMyCode": false
+    },
+    {
+      "name": "Debug: Worker",
+      "type": "python",
+      "request": "attach",
+      "connect": {
+        "host": "localhost",
+        "port": 5679
```
0xfacade, line 132: Note the different ports for server and worker

inference/server/main.py
```diff
@@ -148,3 +148,19 @@ async def maybe_add_debug_api_keys():
...
+    port = int(os.getenv('PORT', "8000"))
+    is_debug = bool(os.getenv("DEBUG", "False"))
+
+    if is_debug:
+        import debugpy
+        debugpy.listen(("0.0.0.0", "5679"))
+        # Uncomment to wait here until a debugger is attached
+        # debugpy.wait_for_client()
+    
+    uvicorn.run("main:app", host="0.0.0.0", port=port, reload=is_debug)
```
0xfacade, line 166: This method of starting the server is only used for development - the docker image for production still invokes the `uvicorn` command

olliestanley (approved):

.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@
 exclude: build|stubs|^bot/templates/$|openassistant/templates|docs/docs/api/openapi.json|scripts/postprocessing/regex_pii_detector.py
 
 default_language_version:
-  python: python3
+  python: python3.10
```
0xfacade, line 32: @olliestanley This is what was causing pre-commit to fail on my machine. `python3` is interpreted as python3.7, which is not new enough for some of the syntax, or isort. When I am more specific, like here, it works. Should this be examined some more?

0xfacade:
@olliestanley From my side, this is ready to be merged, now that I've figured out why pre-commit was failing (see other message).

.pre-commit-config.yaml
olliestanley: I suspect this wouldn't be an issue if you ran the commands from inside a Python 3.10 virtual environment (I guess that's why others haven't had any similar issues) but I don't see any reason we can't make this config change if it helps make things easier

olliestanley:
> @olliestanley From my side, this is ready to be merged, now that I've figured out why pre-commit was failing (see other message).

I'm happy for it to be merged but it touches repo root files etc so will need approval from @andreaskoepf or @yk

.pre-commit-config.yaml
0xfacade: I'm on a rolling release Linux system with Python 3.10.10. It worked correctly inside a Ubuntu Docker container with the same Python version.. so it's probably some other dependency that broke this

PR #3580, opened 2023-07-17
Add flash-attention patch for falcon-7b

andreaskoepf:
Enable the `use_flash_attention` configuration flag for Falcon models. When `use_flash_attention` is set to `true` the [FalconAttention.forwad()](https://github.com/huggingface/transformers/blob/c965d302791cf935d6ea7776428749be678cf509/src/transformers/models/falcon/modeling_falcon.py#L281) method is replaced with a variant that uses Tri Dao's flash_attention instead of pytorch's `scaled_dot_product_attention` function.

At the moment the patch works only for falcon-7b but technically it will also work for falcon-40b with the right configuration. The falcon model situation is currently a bit messy: The Falcon model was recently added to Huggingface transformers (see [PR transformers#24523](https://github.com/huggingface/transformers/pull/24523)) but the falcon models on the hugginface hub use still the code which is shipped together with the weights (a PR to change this [was reverted](https://huggingface.co/tiiuae/falcon-7b/discussions/66)). Falcon-7b and 40b use both slightly different code (which was unified in the HF transformers impl and can there be controlled via a configuration member called `new_decoder_architecture` see [configuration_falcon.py#L65-L67](https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon/configuration_falcon.py#L65-L67)). The HF Falcon impl uses different names in the configuration class, e.g. compare new [configuration_falcon.py](https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon/configuration_falcon.py) and old [configuration_RW.py](https://huggingface.co/tiiuae/falcon-7b/blob/main/configuration_RW.py)

HF Falcon implementation compatible model configurations can be found here:
7B: [config.json](https://huggingface.co/tiiuae/falcon-7b/blob/4e2d06f0a7c6370ebabbc30c6f59377ae8f73d76/config.json)
40B: [config.json](https://huggingface.co/tiiuae/falcon-40b/blob/f1ba7d328c06aa6fbb4a8afd3c756f46d7e6b232/config.json)


model/model_training/models/patching_utils.py
```diff
@@ -7,11 +7,14 @@ def compute_flash_attention(flash_attn, q, k, v, attention_mask=None, head_mask=
     # attention_mask (float): [bs, seq_len]
     batch_size, max_len = q.size(0), q.size(1)
 
-    qkv = torch.stack([q, k, v], dim=2).to(torch.float16)  # need to truncate in case input is fp32
+    qkv = torch.stack([q, k, v], dim=2)
+    dtype_in = qkv.dtype
+    if dtype_in == torch.float32:
+        qkv = qkv.to(torch.float16)  # need to truncate in case input is fp32
```
jordiclive, line 13: Falcon is torch.bfloat16? Will the casting to float16 pose a problem.

model/pyproject.toml
```diff
@@ -10,13 +10,13 @@ authors = [
...
-    "datasets>=2.8.0",
-    "deepspeed>=0.8.3",
+    "accelerate>=0.21.0",
+    "bitsandbytes>=0.40.2",
+    "datasets>=2.13.1",
+    "deepspeed>=0.9.5",
     "evaluate>=0.4.0",
     "fastlangid>=1.0.11",
-    "flash_attn>=0.2.8",
+    "flash_attn>=2.0.0",
```
jordiclive, line 19: 🚀 tested with llama 2 just now vs flash 1 and is faster.

model/model_training/models/patching_falcon.py
```diff
@@ -0,0 +1,78 @@
...
+    q = query_layer_.permute(0, 2, 1, 3)
+    k = key_layer_.permute(0, 2, 1, 3).expand(q.shape)
+    v = value_layer_.permute(0, 2, 1, 3).expand(q.shape)
+
+    if attention_mask is not None:
+        attention_mask = attention_mask[:, 0, -1]
+
+    flash_attn.train(self.training)
+    attn_output = compute_flash_attention(flash_attn, q, k, v, attention_mask=attention_mask)
+    attn_output = attn_output.reshape(batch_size, query_length, self.num_heads * self.head_dim)
```
jordiclive, line 74: ok so we dont need attn_output = attn_output.permute(0, 2, 1, 3) here ? I notice we have that for llama and in the falcon modelling code

jordiclive (commented):
LGTM, minor clarification

model/model_training/models/patching_utils.py
andreaskoepf: This is a minor modification of the old patching_code for llama which potentially can be trained in float32 (at least the small ones). The Old code always casted to float16 .. but flash_attention also supports bfloat16 .. I tested differences between always casting to float32 or as it is now keeping bfloat16 and the output logit-differences were minimal.

jordiclive: Ah ok nice. I know float32 is used for the flash-attn softmax but the dtype_in should never be float32 unless you trained it with float32.

model/model_training/models/patching_falcon.py
andreaskoepf: Yes, I think that is not required here. If channels were mixed I would expect the code to break or to see >0.25 differences in the output logits... I tested with `test_patched_falcon.py`.

model/model_training/models/__init__.py
```diff
@@ -25,17 +21,27 @@ def freeze_top_n_layers(model, target_layers):
...
+        if "falcon-7b" in model_name:
+            # temporary hack until tiiuae/falcon-7b uses the transformer's Falcon impl by default
+            # in-library PR was reverted https://huggingface.co/tiiuae/falcon-7b/commit/378337427557d1df3e742264a2901a49f25d4eb1
+            model = transformers.models.falcon.modeling_falcon.FalconForCausalLM.from_pretrained(
+                model_name, cache_dir=cache_dir, **kwargs
+            )
+        else:
+            if "falcon" in model_name:
+                kwargs["trust_remote_code"] = True
+            model = transformers.AutoModelForCausalLM.from_pretrained(model_name, cache_dir=cache_dir, **kwargs)
```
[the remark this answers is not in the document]
andreaskoepf: It would work with the transformers code if the revision [f1ba7d328c06aa6fbb4a8afd3c756f46d7e6b232](https://huggingface.co/tiiuae/falcon-40b/commit/f1ba7d328c06aa6fbb4a8afd3c756f46d7e6b232) would be explicitly specified. I am not sure why this patch was reverted .. potentially because there is not yet an officially released transformers version which includes the new falcon code ... I left it in there for backward compatibility until the official Falcon HF repos are changed.

model/model_training/models/patching_utils.py
andreaskoepf: correct, this conversion code is mainly for the case that models are trained with `dtype: float32` .. the patching_utils.py is used for all supported models. In the previous version the input was always converted with `.to(torch.float16)` .. and a potential reverse conversion happend in outer code (see here for [llama](https://github.com/LAION-AI/Open-Assistant/blob/64ece883f479ff185297cb8e9b33d2d8f78f2a2c/model/model_training/models/patching_llama.py#L62)) but I think it is better when the attention function returns in the same dtype that it got as input .. therefore I added the reverse dtype tracking and reverse conversion here.

jordiclive (approved):
LGTM

PR #3583, opened 2023-07-19
Orca chat Dataloader

shahules786:
## What
Added support for loading https://huggingface.co/datasets/shahules786/orca-chat

## Why
This dataset contains samples of larger context length (~8k+) that will be used to fine-tune OA models using the scaled Rope method.

## How
Added Dataloader module

### DIFF
--- a/model/model_training/custom_datasets/formatting.py
+++ b/model/model_training/custom_datasets/formatting.py
@@ -91,6 +91,9 @@ class Utterance(BaseModel):
             else:
                 fragments.append(f"{k}: {v}")
 
+        if len(fragments) == 0:
+            return ""
+
         content = "\n".join(fragments)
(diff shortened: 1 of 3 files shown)


model/model_training/custom_datasets/formatting.py
```diff
@@ -109,13 +112,15 @@ class DatasetEntrySft(DatasetEntry):
...
     conversation: list[Utterance]
+    system_message: Optional[str]
 
     def get_formatted(
         self,
         eos_token: str,
         use_system_tag: bool = False,
         system_property_dropout: float = 0.5,
         system_add_length: bool = False,
+        system_message: str = "",
```
andreaskoepf, line 123: this line can be removed

andreaskoepf (approved):
super cool, thanks!

PR #3592, opened 2023-07-21
Use `rng_seed` param when creating custom dataset sampler

andreaskoepf:
Use the `rng_seed` configuration parameter in class `PerDatasetSampler.build_sampler_from_config()` static factory class method. Until now always the fixed default value of 0 was used as seed for the dataset sampling (which I think was as a bug).


jordiclive (approved):
lgtm

shahules786 (approved):

PR #3595, opened 2023-07-23
Update Flash Attention forward for Llama 2: 

jordiclive:
GQA for 34B and 70B and tp have been [added ](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L281)

The current flash attn patch forward is for the old transformers' attention forward.


andreaskoepf (approved):
Very cool thanks a lot!

PR #3597, opened 2023-07-23
Custom instructions feature

draganjovanovich:
Added basic functionality of "custom-instructions"

It utilises two additional fields in work_parameters:
- user_profile (here user would describe what he wants to share with
  llm, for chat session)
- user_response_instructions (here user would describe how they want llm to
  respond to questions and queries, like format, style, length, etc...)
  
Here are some of the current UI changes introduced with this PR.
<details>
  <summary>Custom instructions</summary>
Screenshot 2023-07-23 at 21 46 36
Screenshot 2023-07-23 at 21 46 42
Screenshot 2023-07-23 at 21 47 16
Screenshot 2023-07-23 at 21 48 21
Screenshot 2023-07-23 at 21 48 29
</details>

### DIFF
--- a/inference/worker/chat_chain.py
+++ b/inference/worker/chat_chain.py
@@ -6,6 +6,7 @@ import utils
 import websocket
 from chat_chain_prompts import (
     ASSISTANT_PREFIX,
+    CUSTOM_INSTRUCTIONS_PREFIX,
     HUMAN_PREFIX,
     JSON_FORMAT_NO_PAYLOAD,
     JSON_FORMAT_PAYLOAD,
@@ -56,6 +57,7 @@ class PromptedLLM:
         tool_names: list[str],
         language: str,
         action_input_format: str,
+        custom_instructions: str = "",
     ):
         self.tokenizer = tokenizer
         self.worker_config = worker_config
@@ -66,6 +68,7 @@ class PromptedLLM:
         self.language = language
         self.action_input_format = action_input_format
         self.current_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        self.custom_instructions = custom_instructions
 
     def call(self, prompt: str) -> tuple[str, str]:
         """Prepares and truncates prompt, calls LLM, returns used prompt and response."""
@@ -79,6 +82,7 @@ class PromptedLLM:
             self.tokenizer,
             self.worker_config,
             self.action_input_format,
+            self.custom_instructions,
         )
 
         # We do not strip() outputs as it seems to degrade instruction-following abilities of the model
@@ -111,6 +115,7 @@ def handle_plugin_usage(
     plugin_max_depth: int,
     ws: websocket.WebSocket,
     work_request_id: str,
+    custom_instructions: str = "",
 ) -> tuple[str, inference.PluginUsed]:
     execution_details = inference.PluginExecutionDetails(
         inner_monologue=[],
@@ -142,7 +147,15 @@ def handle_plugin_usage(
     tool_names = [tool.name for tool in tools]
 
     chain = PromptedLLM(
-        tokenizer, worker_config, parameters, prompt_template, memory, tool_names, language, action_input_format
+        tokenizer,
+        worker_config,
+        parameters,
+        prompt_template,
+        memory,
+        tool_names,
+        language,
+        action_input_format,
+        custom_instructions,
     )
 
     # send "thinking..." intermediate step to UI (This will discard queue position 0) immediately
@@ -245,6 +258,7 @@ def handle_plugin_usage(
                 tokenizer,
                 worker_config,
                 action_input_format,
+                custom_instructions,
             )
 
             inner_prompt = f"{inner_prompt}\n{THOUGHT_SEQ} I now know the final answer\n{ASSISTANT_PREFIX}:  "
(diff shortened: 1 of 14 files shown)


inference/worker/work.py
```diff
@@ -40,9 +41,17 @@ def _prepare_message(message: inference.MessageRead) -> str:
...
-    if work_request.parameters.system_prompt:
-        pre_prompt = V2_SYSTEM_PREFIX + work_request.parameters.system_prompt + eos_token
+    # Prepend system prompt and custom_instructions if it was specified in work parameters
+    if (
+        work_request.parameters.system_prompt
+        or work_request.parameters.user_profile
+        or work_request.parameters.user_response_instructions
+    ):
+        pre_prompt = f"""{V2_SYSTEM_PREFIX}{work_request.parameters.system_prompt if work_request.parameters.system_prompt else ""}\n{
+        CUSTOM_INSTRUCTIONS_PREFIX.format(
```
olliestanley, line 51: Can we change this so that the whole custom instruction prefix is not prepended at all if no instructions are given? Otherwise we are just using up tokens for no reason, as the worker retains the system prompt in context always at the cost of other tokens

draganjovanovich: It should be like that right now, At least in my env, while testing.

When there is not system prompt or instructions, nothing is prepended to
the prompt.

draganjovanovich: Or you meant, that we need to have, ```work_request.parameters.system_prompt or (work_request.parameters.user_profile
        and work_request.parameters.user_response_instructions)```
        ?

olliestanley: > Or you meant, that we need to have, `work_request.parameters.system_prompt or (work_request.parameters.user_profile and work_request.parameters.user_response_instructions)` ?

I mean, in the case where `system_prompt` is provided, but neither `user_profile` or `user_response_instructions` are provided, we will currently get a system prompt with the `USER_INSTRUCTIONS_PREFIX` included, right?

So my suggestion would be something like this (not tested, I just wrote this in GitHub editor):
```
    work_params = work_request.parameters
    if work_params.system_prompt or work_params.user_profile or work_params.user_response_instructions:
        pre_prompt = V2_SYSTEM_PREFIX + (work_params.system_prompt or "")

        if work_params.user_profile or work_params.user_response_instructions:
            pre_prompt = f"""{pre_prompt}\n{CUSTOM_INSTRUCTIONS_PREFIX.format(user_profile=work_params.user_profile or "", user_response_instructions=work_params.user_response_instructions or "")}"""

        pre_prompt = pre_prompt + eos_token
```
Unless I'm missing something here and the prefix already would be omitted somehow in that case?

inference/worker/chat_chain.py
```diff
@@ -355,11 +379,17 @@ def handle_conversation(
             "language",
             "current_time",
             "action_input_format",
+            "custom_instructions",
         ] + (["tools_names"] if plugin_enabled else [])
 
         # TODO: Consider passing language from the UI here
         prompt_template = PromptTemplate(input_variables=input_variables, template=TEMPLATE)
 
+        custom_instructions = f"""\n{CUSTOM_INSTRUCTIONS_PREFIX.format(
```
olliestanley, line 388: Also here - surely if `user_profile` and `user_response_instructions` are empty, we want this value `custom_instructions` to also be an entirely empty string, without the contents of `CUSTOM_INSTRUCTIONS_PREFIX`

inference/worker/work.py
draganjovanovich: Thanks,
I will checkout this and commit changes.

inference/worker/work.py
```diff
@@ -40,9 +41,20 @@ def _prepare_message(message: inference.MessageRead) -> str:
...
-        pre_prompt = V2_SYSTEM_PREFIX + work_request.parameters.system_prompt + eos_token
+    # Prepend system prompt and custom_instructions if it was specified in work parameters
+    if (
+        work_request.parameters.system_prompt
+        or work_request.parameters.user_profile
+        or work_request.parameters.user_response_instructions
+    ):
+        work_params = work_request.parameters
+        if work_params.system_prompt or work_params.user_profile or work_params.user_response_instructions:
+            pre_prompt = V2_SYSTEM_PREFIX + (work_params.system_prompt or "")
```
olliestanley, line 52: This can be simplified a bit (currently the same `if` condition is repeated)
```suggestion
    work_params = work_request.parameters
    if work_params.system_prompt or work_params.user_profile or work_params.user_response_instructions:
        pre_prompt = V2_SYSTEM_PREFIX + (work_params.system_prompt or "")
```

draganjovanovich: Yes, I just badly copy/pasted. I committed your change, Thanks!

olliestanley (approved):
Inference backend/worker changes lgtm (have not reviewed frontend changes)

website/src/components/Chat/ChatConfigForm.tsx
```diff
@@ -104,6 +111,10 @@ export const ChatConfigForm = memo(function ChatConfigForm() {
   const selectedModel = getValues("model_config_name"); // have to use getValues to here to access latest value
   const selectedPlugins = getValues("plugins");
   const presets = modelInfos.find((model) => model.name === selectedModel)!.parameter_configs;
+  const [customInstructions, setCustomInstructions] = useState<CustomInstructionsType>({
+    user_profile: "",
+    user_response_instructions: "",
+  });
```
notmd, line 117: This state can be handled by react-hook-form. You just need to wrap the `CustomInstructions` with the `Controller` component.

notmd (approved):
I haven't run the code yet but overall it looks good. Just one small thing to clean up the code and avoid re-render. Feel free to merge if you can't make it work with react-hook-form.

website/src/components/Chat/ChatConfigForm.tsx
draganjovanovich: Thanks, could you please provide a commit/change for that here if you have time and it's small?

PR #3606, opened 2023-07-25
Dolphin gpt 3.5 data mix

shahules786:
- Added dolphin random data mix to form conversations from gpt 3.5 file.
- Instructions of the same type are only considered while picking at random to form conversation
- Also ensured that same samples are not considered more than once 

Configure
```
- dolphin-mix
        num_samples: 100000
        max_char_len: 32000
        seed: 44
```

### DIFF
--- a/model/model_training/custom_datasets/prompt_dialogue.py
+++ b/model/model_training/custom_datasets/prompt_dialogue.py
@@ -4,6 +4,7 @@ import re
 from pathlib import Path
 from typing import List, Optional, Union
 
+import numpy as np
 import requests
 from datasets import load_dataset
 from model_training.custom_datasets.formatting import DatasetEntrySft, Role, Utterance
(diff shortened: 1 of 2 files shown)


model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -193,3 +194,61 @@ def __getitem__(self, idx):
...
+
+
+class DolphinMix(Dataset):
+    name = "dophin-mix"
+
+    def __init__(self, cache_dir, num_samples=100000, max_char_len=8000, seed=42):
+        self.dataset = load_dataset(
+            "ehartford/dolphin", data_files="flan5m-alpaca-uncensored.jsonl", cache_dir=cache_dir
+        )
+        self.dataset = self.dataset["train"].shuffle(seed).select(range(num_samples))
```
andreaskoepf, line 206: I should have used shuffle(seed) too in my code ;-) .. will update later.

model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -193,3 +194,61 @@ def __getitem__(self, idx):
...
+                conversation_len = len(inst)
+                if idx not in removed_indices and conversation_len < self.max_char_len:
+                    conversation = {"conversation": []}
+                    conversation["instruction"] = inst
+                    input, output = [data_sample[idx][key] for key in ("input", "output")]
+                    conversation["conversation"].append({"input": input, "output": output})
+                    conversation_len += len(input) + len(output)
+                    removed_indices.append(idx)
+                    while conversation_len < self.max_char_len:
+                        indices_to_pick = np.setdiff1d(available_indices, removed_indices)
```
andreaskoepf, line 225: Generating the indices to pick set every time seems a bit costly .. but since it is only done during startup it might be ok.

andreaskoepf (approved):
Thanks a lot .. very pythonic. ;-)

PR #3609, opened 2023-07-26
Dolphin-mix code refactor

shahules786:
## What
refactor dolphin-mix loader for optimized and reproducible loading

## Why
The earlier version has optimization and randomization issues. 

## How
Simplified the code and seeded every random


model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -199,40 +198,30 @@ def __getitem__(self, idx):
...
         self.dataset = load_dataset(
             "ehartford/dolphin", data_files="flan5m-alpaca-uncensored.jsonl", cache_dir=cache_dir
         )
         self.dataset = self.dataset["train"].shuffle(seed).select(range(num_samples))
         self.max_char_len = max_char_len
         instructions = set([item["instruction"] for item in self.dataset])
-
         self.conversations = []
         for inst in instructions:
             data_sample = self.dataset.filter(lambda example: example["instruction"] == inst)
```
andreaskoepf, line 212: I printed the length of this data_sample .. without ordering the instructions the different deep-speed processed showed different numbers here. In general I am now unsure if the dataset operations filter/select etc. are stable and always return elements in the same order. We probably need to investigate this more.

model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -199,40 +198,30 @@ def __getitem__(self, idx):
...
+            for idx in range(0, len(data_sample)):
+                input_, output = [data_sample[int(idx)][key] for key in ("input", "output")]
+                input_len = len(input_) + len(output)
+                if conversation_len + input_len < self.max_char_len:
+                    conversation["conversation"].append({"input": input_, "output": output})
+                    conversation_len += len(input_) + len(output)
+                else:
                     self.conversations.append(conversation)
+                    conversation_len = len(inst)
+                    conversation = {"conversation": [], "instruction": inst}
```
andreaskoepf, line 224: the last conversation (that might now have reached `max_char_len` is lost

model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -199,40 +198,30 @@ def __getitem__(self, idx):
 class DolphinMix(Dataset):
     name = "dophin-mix"
 
-    def __init__(self, cache_dir, num_samples=100000, max_char_len=8000, seed=42):
+    def __init__(
+        self, cache_dir: Optional[str] = None, num_samples: int = 100000, max_char_len: int = 8000, seed: int = 42
```
andreaskoepf, line 202: `max_char_len` is actually more `min_char_len` ...

PR #3612, opened 2023-07-27
Changes for orcacode experiment

andreaskoepf:


shahules786:
Besides this I think there is a bug in this [line](https://github.com/LAION-AI/Open-Assistant/blob/9c4eafca993f8effcd08f59d08864552dcc6999b/model/model_training/models/patching.py#L250). Config cannot be updated. I removed it from my version. Can you check that too with this PR? @andreaskoepf

model/model_training/custom_datasets/prompt_dialogue.py
```diff
@@ -199,45 +198,37 @@ def __getitem__(self, idx):
 class DolphinMix(Dataset):
     name = "dophin-mix"
 
-    def __init__(self, cache_dir, num_samples=100000, max_char_len=8000, seed=42):
+    def __init__(self, cache_dir, num_samples: Optional[int] = None, max_char_len: int = 8000, seed: int = 42):
```
shahules786, line 201: Isn't it better to add `data_files` as an argument with the default value "flan-5m..."?  In this way, we can change to gpt-4 version using config if required. 
`

model/model_training/models/patching.py
shahules786, line 1: Can you remove the `update_config` function definition too since it's not used?

shahules786 (changes requested):

shahules786:
Nice, thanks.

shahules786 (approved):

s0mt0chukwu (approved):

PR #3614, opened 2023-07-28
[Fix] Consume much more gpt memory running eval_rm

### LINKED ISSUE #3611: Consuming much more gpu memory than expected using model_training and model_eval
Previously, I have trained a pythia-6.9b using code here: [dolly](https://github.com/databrickslabs/dolly)
I can train with the below setting on 4xA100(80G) without GPU OOM:
```yaml
per-device-train-batch-size: 8
per-device-eval-batch-size: 8
gradient-accumulation-steps: 2
max len: 2048
gradient checkpointing: false
use_cache: true
bf16: true
```
with deepspeed config [here](https://github.com/databrickslabs/dolly/blob/master/config/a100_config.json)
I can also evaluate the output model with [lm-evaluation-harness](https://github.com/EleutherAI/lm-evaluation-harness) on single gpu with a non-one batch size.
However, now I am using `model_training` to train a reward model.
I can only run with the below setting on 8xA100(80G):
```yaml
per_device_train_batch_size: 4 # can be bigger using gradient checkpointing
per_device_eval_batch_size: 4
gradient_accumulation_steps: 4
max len: 2048
gradient checkpointing: true # otherwise got GPU OOM even with per_device_train_batch_size 1
use_cache: false # have to turn off since conflict with grandient checkpointing
bf16: true
```
with deepspeed config `zero3_config_sft.config`.(As you can see, it is very alike with the one above)
In addition, I can not eval the output model using `eval_rm.py` on single gpu(even with batch size 1) because of GPU OOM.
I didnt find any code that reduce GPU memory in `dolly` or `lm-evaluation-harness`. And the model `GPTNeoXforCasualLm` should consume more memory than `GPTNeoXRewardModel` as I see the code of the output layer.

SingL3:
Fix #3611.
Still debugging or model_training.

### DIFF
--- a/model/model_eval/eval_rm.py
+++ b/model/model_eval/eval_rm.py
@@ -10,6 +10,7 @@ from torch.utils.data import DataLoader
 from transformers import AutoModelForSequenceClassification, AutoTokenizer
 from transformers.trainer_utils import EvalPrediction
 from utils import write_to_json
+from tqdm import tqdm
 
 DATASETS = ["SHP", "Hellaswag", "HFdataset"]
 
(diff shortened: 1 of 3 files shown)


shahules786 (approved):
LGTM

PR #3637, opened 2023-08-06
Add paper acknowledgement form

yk:
I know all of this is suboptimal, but I'd like to get people to sign up to be mentioned asap. Feel free to improve (or to tell me how to improve) now or in subsequent PRs. I anticipate this form will be going away in the future, so I'm not sure a lot of work will be really worth it.


AbdBarho:
Hey sorry I have guests over I will try to give it a look in the evening, what is the context for this? is it required as part of the paper submission?

yk:
> Hey sorry I have guests over I will try to give it a look in the evening, what is the context for this? is it required as part of the paper submission?

It's not required, no. But it would be nice to acknowledge contributors. So it's not a super stress, but we should get it going in the next days.

AbdBarho:
Would it make sense to allow the users to change their user name, and then make the leaderboard public? with all users with more than 0 points?

yk:
> Would it make sense to allow the users to change their user name, and then make the leaderboard public? with all users with more than 0 points?

we could do that as well, but for a lot of people I think it would be cool to see their name listed in a published paper.

AbdBarho:
Oh, of course, that was my intention as well, my proposal was to read from the leaderboards, but I guess we would need consent and a "real" name anyway, so no problem.

I took the liberty of cleaning up some stuff so it is not a "hack" anymore, I hope you are fine with that

Although the database changes are not the best, it is not a problem since they should be only temporary.

AbdBarho (approved):

andreaskoepf (approved):
Ok, nice.

PR #3638, opened 2023-08-06
Added plugin repositories to add-plugin menu

yk:
For now I've added getit.ai as a repository, but we can easily add others, they're just a link for now

### DIFF
--- a/website/src/components/Chat/PluginsChooser.tsx
+++ b/website/src/components/Chat/PluginsChooser.tsx
@@ -2,6 +2,7 @@ import {
   Avatar,
   Box,
   Button,
+  Flex,
   IconButton,
   Input,
   Menu,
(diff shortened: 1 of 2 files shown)


website/src/components/Chat/PluginsChooser.tsx
```diff
@@ -234,6 +236,16 @@ export const PluginsChooser = ({ plugins, setPlugins }: PluginsChooserProps) =>
               placeholder={t("plugin_url_placeholder")}
             />
             {selectedForEditPluginIndex !== null && <JsonCard>{plugins[selectedForEditPluginIndex]}</JsonCard>}
+            <Box>
+              <Text fontSize="md" fontWeight="bold" my={2}>
+                {t("plugin_repositories")}
+              </Text>
+              <Flex gap={2}>
+                <Button as={Link} href="https://www.getit.ai/gpt-plugins" target="_blank" rel="noreferrer">
```
digi604, line 244: if you have you would have callback url for plugin installs you could use getit's callback mechanism: https://www.getit.ai/gpt-plugins/chat-integration

andreaskoepf: We don't have this yet but sounds like a good idea .. could you maybe create a new issue for this?

PR #3646, opened 2023-08-08
Long context datasets

shahules786:
## What
Added support for RAG based dataloaders 
Currently supports 
- shahules786/Multi-chapter-summaries


s0mt0chukwu (commented):
Interesting UwU

andreaskoepf (approved):

PR #3649, opened 2023-08-11
Dev zhanglu

zhanglu0704:
Check whether the size of the embedding layer has been resized, rather than checking if the length of the tokenizer is equal to the size of the embedding layer. 
This is because the embedding layer of the model may be resized and padded to a multiple of 16, but the tokenizer may not be padded, which will cause the freezing layer to fail. However, at this time, the embedding layer will not be resized and freezing can still be performed.

### DIFF
--- a/model/pyproject.toml
+++ b/model/pyproject.toml
@@ -42,7 +42,7 @@ tests = [
 ]
 
 [tool.setuptools]
-py-modules = ["model_training"]
+packages = ["model_training"]
 
 [tool.black]
 line-length = 120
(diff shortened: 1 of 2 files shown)


model/pyproject.toml
```diff
@@ -42,7 +42,7 @@ tests = [
 ]
 
 [tool.setuptools]
-py-modules = ["model_training"]
+packages = ["model_training"]
```
andreaskoepf, line 45: I am unsure why this change is shown, this was already merged in https://github.com/LAION-AI/Open-Assistant/pull/3643.

andreaskoepf (approved):
ok, thanks for the fix. Maybe it could later be changed to completely use the embedding sizes and not use the size of the vocab as the basis. I will merge it for now as you proposed.

PR #3651, opened 2023-08-12
Make sure eval subset is sampled without replacing

andreaskoepf:
Explicitly specify `replace=False` for [numpy.random.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html) (it was missing and default is `replace=True` which could lead to duplicate examples in the evaluation set).


shahules786 (approved):
Nice catch

PR #3654, opened 2023-08-15
Add pretokenizer utility

andreaskoepf:
The pretokenizer utility (`pretokenizer/pretokenize.py`) allows to tokenize datamixes in advance for use with the [epfLLM/Megatron-LLM/](https://github.com/epfLLM/Megatron-LLM) trainer.

The datamix configuration can be defined in a yaml file similarly to the classic training configurations of trainer_sft.py. For loading the datasets the functions from `model_training` are used (therefore the model_training module needs to be installed).


shahules786 (approved):
LGTM

PR #3664, opened 2023-08-20
Added "Please wait..." loading on "continue with email" Button

taqui-786:
Adding "Please wait..." on "continue with google" Button in login page
when a user visit login page "continue with email" button it is disabled for a second beause of initailizing captcha in backend 
**There for i want to add "please wait..." in the place of "continue with email"  to tell user that button is getting ready**


taqui-786:
pull request failed

PR #3674, opened 2023-08-29
Fix german leaderboard translation

andreaskoepf:
- fix order to match leaderboard.json of other locales
- fix template: `{{Bedarf, Zahl, Ganzzahl}}` -> `{{level,number,integer}}`


AbdBarho (approved):
Danke!

PR #3679, opened 2023-09-05
Added and updated plugin repositories

targed:
Added More plugin repositories to the chat plugin interface and added their corresponding references in the docs.


targed:
I can't figure out what is going on with Prettier that is making it go all crazy. 
If I figure out how to fix it, I will.

PR #3754, opened 2024-08-17
Use the `dtype` parameter of `pd.read_csv`.

monster29000:


RahulVadisetty91 (commented):
re4544

The function num_tokens_from_string is using the hardcoded model "gpt-3.5-turbo" in tiktoken. encoding_for_model("gpt-3.5-turbo"). If you are working with BART models this might not be the appropriate tokenizer. Instead, use the correct tokenizer based on the facebook/bart-large-cnn model, which you are using for summarization.

